PR #57, opened 2021-02-01
Add version field to default serialized format

JLockerman:
This commit adds a version byte, currently always set to 1, to the default serialization format. This will allow us to change the serialization format in a forwards compatible manner, by dispatching on this field.

### DIFF
--- a/extension/src/type_builder.rs
+++ b/extension/src/type_builder.rs
@@ -144,19 +144,35 @@ macro_rules! flatten {
     }
 }
 
+#[repr(u8)]
+pub enum SerializationType {
+    Default = 1,
+}
+
 #[macro_export]
 macro_rules! do_serialize {
     ($state: ident) => {
         {
+            $crate::do_serialize!($state, version: 1)
+        }
+    };
+    ($state: ident, version: $version: expr) => {
+        {
+            use $crate::type_builder::SerializationType;
+
             let state = &*$state;
             let size = bincode::serialized_size(state)
-            .unwrap_or_else(|e| pgx::error!("serialization error {}", e));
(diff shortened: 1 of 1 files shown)


extension/src/type_builder.rs
```diff
@@ -150,13 +150,15 @@ macro_rules! do_serialize {
...
-            .unwrap_or_else(|e| pgx::error!("serialization error {}", e));
+                .unwrap_or_else(|e| pgx::error!("serialization error {}", e)) + 1;
             let mut bytes = Vec::with_capacity(size as usize + 4);
             let mut varsize = [0; 4];
             unsafe {
                 pgx::set_varsize(&mut varsize as *mut _ as *mut _, size as _);
             }
             bytes.extend_from_slice(&varsize);
+            // serialization version; 1 for bincode is currently the only option
+            bytes.push(1);
```
davidkohn88, line 161: Why not make this a parameter of the function and have everything using it call with it?

davidkohn88 (commented):
I'm not sure why we're setting this to 1 always now rather than making it settable by the calling function. Is the plan that we increment this version for everything all at once? Or that each type of thing will have its own versioning?

JLockerman:
I expect the most likely scenario is that we never change the version number, and that the default should make this unchanging use case as easy as possible. I think the second most likely scenario is that we change our default serialization format for everything. In the event we do want change a specific one, we'll have to add an optional argument.

extension/src/type_builder.rs
WireBaron, line 161: It would be better if we used an enum here instead of a magic constant.

JLockerman:
bors retry

PR #66, opened 2021-02-05
Initial implementation of time_weighted_average

davidkohn88:


extension/src/time_weighted_average.rs
```diff
@@ -0,0 +1,435 @@
...
+// WITH t as (SELECT time_bucket('5 min', ts) as bucket, id,  time_weight(ts, value, method=>'locf') as tw
+// FROM foo
+// WHERE ts > '2020-10-01' and ts <= '2020-10-02'
+// [AND id IN ('foo', 'bar', 'baz')]
+// GROUP BY 1, 2 )
+// SELECT bucket, id, average(with_bounds(tw, bounds => time_bucket_range(bucket, '5 min'), prev=>(SELECT tspoint(ts, value) FROM foo f WHERE f.id = t.id AND f.ts < '2020-10-01' ORDER BY ts DESC LIMIT 1)) OVER (PARTITION BY id ORDER BY bucket ASC ))
+// FROM t;
+// with_bounds(time_weight(ts, val, 'linear'), bounds=> tstzrange, prev => tspoint(), next => tspoint())
+// SELECT average(time_weight_ordered(time, value, 'linear'))
+// SELECT time_weight_ordered(time, value, 'linear') |> with_bounds(bounds)
```
JLockerman, line 44: this should be deleted I think

JLockerman (changes requested):

temp.sql
```diff
@@ -0,0 +1,25 @@
...
+                SELECT tspoint(ts, value)
+                FROM foo f
+                WHERE f.id = t.id
+                    AND f.ts < '2020-10-01'
+                ORDER BY ts DESC
+                LIMIT 1
+            )
+        ) OVER (PARTITION BY id ORDER BY bucket ASC)
+    )    
+FROM t;
```
JLockerman, line 25: same with this file

extension/src/time_weighted_average.rs
davidkohn88: yea and probably temp.sql too

extension/src/time_weighted_average.rs
```diff
@@ -0,0 +1,422 @@
...
+            Some(m)=> {if m != method {panic!("Mismatched methods")}}
+        }
+        self.buffer.push(value);
+    }
+
+    fn sort_and_calc(&mut self) {
+        if self.buffer.is_empty() {
+            return
+        }
+        self.buffer.sort_unstable_by_key(|p| p.ts);
```
WireBaron, line 233: It seems like we're trying to tolerate data being entered out of order here.  However it looks like this will be in an invalid state if anything in the buffer has a timestamp earlier than the last timestamp present when we last called this function.  Is that a problem?  If we do find ourselves in that state we won't actually blow up until the combine_sorted_iter call in `combine_sort_and_calc` later on, which isn't great.

extension/src/time_weighted_average.rs
```diff
@@ -0,0 +1,422 @@
...
+        };
+        self.buffer.clear();
+    }
+    fn combine_push(&mut self, other: &TimeWeightTransState) {
+        let cb = other.combine_buffer.clone();
+        for val in cb.into_iter(){
+            self.combine_buffer.push(val);
+        };
+    }
+    fn combine_sort_and_calc(&mut self) {
```
WireBaron, line 246: Alright, these function names are painful, and it might just come from how your variables are named.  Maybe `computed_summaries` or `summary_buffer` instead of `combine_buffer` and then `push_summary` and `combine_summaries` for the functions affecting it.

extension/src/time_weighted_average.rs
```diff
@@ -0,0 +1,422 @@
...
+        if self.combine_buffer.len() <= 1 {
+            return
+        }
+        self.combine_buffer.sort_unstable_by_key(|s| s.first.ts);
+        self.combine_buffer = vec![TimeWeightSummary::combine_sorted_iter(&self.combine_buffer).unwrap()];
```
WireBaron, line 251: I don't have a great suggestion for improving this aside from maybe some comments, but this line was extremely confusing for me.

extension/src/time_weighted_average.rs
```diff
@@ -0,0 +1,422 @@
...
+            match (state1, state2) {
+                (None, None) => None,
+                (None, Some(state2)) => {let mut s = state2.clone(); s.sort_and_calc(); Some(s.into())},
+                (Some(state1), None) => {let mut s = state1.clone(); s.sort_and_calc(); Some(s.into())}, //should I make these return themselves?
+                (Some(state1), Some(state2)) => {
+                    let mut s1 = state1.clone(); // is there a way to avoid if it doesn't need it?
+                    s1.sort_and_calc(); 
+                    let mut s2 = state2.clone();
+                    s2.sort_and_calc();
+                    s2.combine_push(&s1);
```
WireBaron, line 341: Should we ensure these have the same TimeWeightMethod here, or inside this function?  I think right now it won't complain until someone calls `combine_sort_and_calc` at some point in the future.

extension/src/time_weighted_average.rs
```diff
@@ -0,0 +1,422 @@
...
+    sfunc = time_weight_trans,
+    stype = internal,
+    finalfunc = time_weight_final,
+    combinefunc = time_weight_combine,
+    serialfunc = time_weight_trans_serialize,
+    deserialfunc = time_weight_trans_deserialize,
+    parallel = restricted
+);
+
+CREATE AGGREGATE time_weight(tws time_weight_summary)
```
WireBaron, line 393: I'm not sure I understand how this will be used.  Is this to handle aggregation of the already aggregated data?

WireBaron (commented):
I'm having a hard time understanding how this will be used without any postgres test code to look at.  I'd recommend just providing the time_weight aggregate since it should work fine with ordered data, and saving the time_weight_ordered specialization unless we find we need it.

davidkohn88:
> I'm having a hard time understanding how this will be used without any postgres test code to look at. I'd recommend just providing the time_weight aggregate since it should work fine with ordered data, and saving the time_weight_ordered specialization unless we find we need it.

There are two reasons it was around: 
1) because it was the first implementation, which is not a great reason to keep it around
2) because it will work with basically any size data and you don't risk OOMing with it. I'm not sure if it's a good enough reason to keep it around for now, but I'd like @JLockerman to weigh in on that.

extension/src/time_weighted_average.rs
davidkohn88: Yeah, we will error in that case, but I'm not sure it's worth it to check, as there's any number of ways we could be in an invalid state here, no two summaries can have overlapping state, so checking that would require scanning the whole thing. It's also relatively unlikely that we will end up with two calls to that function anyway.

davidkohn88: (But I'm going to add comments to clarify some of this, because it is quite confusing)

extension/src/time_weighted_average.rs
davidkohn88: modified the names.

extension/src/time_weighted_average.rs
davidkohn88: Yes. I will leave some comments about this, but, if we don't already have this for tdigest and the other things, we shoudl absolutely add, it's a huge improvement in the UI for re-aggregation from Continuous aggregates cause I don't think the Timescale side solution for that is coming for a little while and having that is a huge improvement.

extension/src/time_weighted_average.rs
```diff
@@ -0,0 +1,422 @@
...
+
+// hack to allow us to qualify names with "timescale_analytics_experimental"
+// so that pgx generates the correct SQL
+mod timescale_analytics_experimental {
+    pub(crate) use super::*;
+    extension_sql!(r#"
+        CREATE SCHEMA IF NOT EXISTS timescale_analytics_experimental;
+    "#);
+}
+extension_sql!(r#"set search_path to 'timescale_analytics_experimental', 'public';"#);
```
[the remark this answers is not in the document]
davidkohn88: yea. I did that, I'm not horribly happy about it, because it's a real PITA and makes things much worse and more likely that we forget to do it correctly somewhere, but yeah, you could lose search_path modifications, this is an edge case definitely, because if you have search path modifications you made them when you connected and probably can make them again, but does seem nicer to not do that. I think this will eventually be addressed with an aggregate builder macro in pgx that will allow for schema qualification, which will be a much more elegant want to deal with this.

extension/src/time_weighted_average.rs
```diff
@@ -0,0 +1,422 @@
...
+                (Some(ts), Some(val)) => TSPoint{ts, val},
+            };
+            let method = match method.to_lowercase().as_str() {
+                "linear"=> TimeWeightMethod::Linear,
+                "locf" => TimeWeightMethod::LOCF,
+                &_ => panic!()
+            };
+            let state = match state {
+                None => TimeWeightSummary::new(p, method).into(),
+                Some(state) => {state.clone().accum(p).unwrap(); state},
```
JLockerman, line 56: I don't think this `clone()` should be needed 🤔

extension/src/time_weighted_average.rs
```diff
@@ -0,0 +1,422 @@
...
+            let p = match (ts, val) {
+                (_, None) => return state,
+                (None, _) => return state, 
+                (Some(ts), Some(val)) => TSPoint{ts, val},
+            };
+            let method = match method.to_lowercase().as_str() {
+                "linear"=> TimeWeightMethod::Linear,
+                "locf" => TimeWeightMethod::LOCF,
+                &_ => panic!()
+            };
```
JLockerman, line 53: I think this should be moved into the `None` branch of the `match state`; no reason to reparse after the first time

extension/src/time_weighted_average.rs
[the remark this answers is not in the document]
JLockerman: I'd perfer to just keep the `set search_path; ... reset search_path` to a single `extension_sql!` block; I don't want to rely on pgx not doing something weird between them.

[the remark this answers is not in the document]
JLockerman: (forgetting to add the schema seems to me like a non-issue right now, we test for that, and my experience adding deliberately-stable features is that the tests work annoyingly well)

crates/time-weighted-average/src/lib.rs
```diff
@@ -0,0 +1,334 @@
...
+impl TimeWeightSummary {
+    pub fn new(pt: TSPoint, method: TimeWeightMethod) -> Self {
+        TimeWeightSummary { method: method, first: pt, last: pt, w_sum: 0.0 }
+    }
+
+    pub fn accum(&mut self, pt: TSPoint) -> Result<(), TimeWeightError> {
+        if pt.ts < self.last.ts {
+            return Err(TimeWeightError::OrderError);
+        }
+        if pt.ts == self.last.ts {
```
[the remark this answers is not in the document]
davidkohn88: This is discussed here: https://github.com/timescale/timescale-analytics/discussions/65

extension/src/time_weighted_average.rs
davidkohn88: Hmm...is there a way to mark a function input that way in an aggregate? I could imagine weird cases where you'd have this change, but they're all very odd and would lead to weird errors, ie a case statement in the select that coudl give odd results if you're not careful...

extension/src/time_weighted_average.rs
davidkohn88: Yeah. checking this is a bit weird right now. I think I want to modify when that happens in general, I'm going to leave that as outstanding for now.

davidkohn88:
This still needs significantly more testing and definitely isn't ready for stable release, but I think we can begin to get feedback from users in the experimental part of it and continue to improve testing and documentation in further commits.

JLockerman (commented):
I think it's ready to merge when you and @WireBaron think it is. (could use a bit of cleaning, but that can come after)

JLockerman (approved):

WireBaron (approved):

PR #69, opened 2021-02-08
Fix t-digest round-tripping issue

JLockerman:
Even when `count=max_buckets` there can be cases where the number of full buckets is smaller than `max_buckets`. Update the t-digest serialization to handle this case.


WireBaron (approved):
Good catch!

PR #70, opened 2021-02-09
Json in/out

JLockerman:
This allows us to to have TEXTual input for our types as well as output, allowing us to work correctly with pg_dump/restore, for not much more work. I'm not _entirely_ satisfied with either the formatting or effort-required, but both can be fixed later.

### DIFF
--- a/extension/src/serialization/collations.rs
+++ b/extension/src/serialization/collations.rs
@@ -16,6 +16,12 @@ use pgx::*;
 #[repr(transparent)]
 pub struct PgCollationId(pub Oid);
 
+impl PgCollationId {
+    pub fn is_invalid(&self) -> bool {
+        self.0 == pg_sys::InvalidOid
+    }
+}
+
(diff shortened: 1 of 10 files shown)


extension/src/serialization/collations.rs
```diff
@@ -68,6 +74,11 @@ impl Serialize for PgCollationId {
         S: serde::Serializer,
     {
         unsafe {
+            let mut layout: Option<(&str, &str)> = None;
+            if self.0 == pg_sys::InvalidOid {
```
WireBaron, line 78: Didn't you just add a function with this very logic?
```suggestion
            if self.is_invalid() {
```

WireBaron (approved):

extension/src/serialization/collations.rs
JLockerman: Yeah. That code predated the function 😅

PR #74, opened 2021-02-11
Hyperloglog union and bug fixes

JLockerman:
This operation calculates the count of distinct values across the combined Hyperloglog.

Also fixes a bug in our varlen handling: varlen types have a short version where there's only 1 byte of header and everything is stored in-line. This breaks one of the assumptions our value-wrapper is doing, so perform a copy in the case where we detect our value is in that form to ensure that we have the full 4-byte header. I believe this is the correct trade off, as the additional copy in the rare case that the value is small enough to use this optimization _should_ be less expensive than the additional disk space would be.


JLockerman:
bor r+

PR #80, opened 2021-02-22
Schema-qualify all the things

JLockerman:
Our release process dictates that all new features get added to an experimental schema, and only get added to the public schema once they are fully baked and ready for release. This commit adds a test that all new functionality is in said schema, and schema-qualifies all currently in-development features.

PR #81, opened 2021-02-22
Schema-qualify experimental functions and remove unneeded prefixes

JLockerman:
Our release process dictates that all new features get added to an experimental schema, and only get added to the public schema once they are fully baked and ready for release. This commit adds a test that all new functionality is in said schema, and schema-qualifies all currently in-development features.

Also, now that can have separate SQL and rust names, remove all the unneeded prefixes to the SQL functions.

supersedes #80


WireBaron (approved):
May make sense to try to give some of the hyperloglog commands prefix free names now as well.
Speaking of, we still need to update the documentation to use the new command names for section titles and content links, but that doesn't have to happen in this change.

JLockerman:
@WireBaron yeah I figured the links weren't as urgent since they weren't immediately visible (besides I wanted to have another look at how the links looked in practice, I could see us wanting to keep the prefix there if it makes them nicer 😋)

PR #85, opened 2021-02-25
Add trigger disabling deps on experimental features

JLockerman:
"experimental" features are just that, and there usage can be fraught: we are neither capable, not desirous, of providing an upgrade path for someone using experimental features to a newer version of this extension, and we would prefer users to not become reliant on experimental features. As such, our upgrade scripts for the extension are intended to contain a
`DROP SCHEMA timescale_analytics_experimental` at the start of the script. To ensure users are adequately warned of this risk, this commit adds a ddl_end trigger that prevents users from creating dependencies on experimental features unless they set a guc indicating they know the
implications of what they're doing.

### DIFF
--- a/extension/src/lib.rs
+++ b/extension/src/lib.rs
@@ -1,3 +1,6 @@
+
+pub mod triggers;
+
 pub mod tdigest;
 pub mod hyperloglog;
 pub mod uddsketch;
@@ -8,7 +11,24 @@ mod type_builder;
 mod serialization;
 mod schema_test;
 
+use pgx::*;
+
 pgx::pg_module_magic!();
+
+static EXPERIMENTAL_ENABLED: GucSetting<bool> = GucSetting::new(false);
+
+#[pg_guard]
+pub extern "C" fn _PG_init() {
+    GucRegistry::define_bool_guc(
+        "timescale_analytics_acknowledge_auto_drop",
+        "enable creation of auto-dropping objects using experimental timescale_analytics_features",
+        "experimental features are very unstable, and objects \
+            depending on them will be automatically deleted on extension update",
+        &EXPERIMENTAL_ENABLED,
+        //TODO should this be superuser?
+        GucContext::Userset,
+    );
+}
 
 #[cfg(test)]
 pub mod pg_test {
--- a/extension/src/triggers.rs
+++ b/extension/src/triggers.rs
@@ -1,0 +1,105 @@
+use pgx::*;
+
+/// nop function that forces the extension binary to be loaded, this ensures
+/// that if a user sees the error message the guc will in fact be active
+#[pg_extern]
+fn timescale_analytics_probe() {
+
+}
+
+extension_sql!(r#"
+CREATE OR REPLACE FUNCTION disallow_experimental_dependencies()
+  RETURNS event_trigger
+ LANGUAGE plpgsql
+  AS $$
+DECLARE
+  guc_set TEXT;
+  experimental_schema_id oid;
+BEGIN
+
+  guc_set := current_setting('timescale_analytics_acknowledge_auto_drop', true);
+  IF guc_set IS NOT NULL AND guc_set = 'on' THEN
+    RETURN;
+  END IF;
+
+  SELECT oid schema_oid
+  INTO experimental_schema_id
+  FROM pg_catalog.pg_namespace
+  WHERE nspname='timescale_analytics_experimental'
+  LIMIT 1;
+
+  IF EXISTS (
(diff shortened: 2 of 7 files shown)


extension/src/triggers.rs
```diff
@@ -0,0 +1,56 @@
+use pgx::*;
+
+#[pg_extern]
+fn timescale_analytics_probe() {
+
+}
+
+extension_sql!(r#"
+CREATE OR REPLACE FUNCTION disallow_experimental_dependencies()
```
davidkohn88, line 9: Can/should we add a few tests around this? To make sure it's catching dependencies?

extension/src/triggers.rs
```diff
@@ -0,0 +1,56 @@
...
+  RETURNS event_trigger
+ LANGUAGE plpgsql
+  AS $$
+DECLARE
+  guc_set TEXT;
+  cmd record;
+  experimental_schema_id oid;
+BEGIN
+
+  guc_set := current_setting('timescale_analytics_install_experimental', true);
```
davidkohn88, line 19: This GUC doesn't match the one in the detail on the exception

extension/src/triggers.rs
```diff
@@ -0,0 +1,56 @@
+use pgx::*;
+
+#[pg_extern]
+fn timescale_analytics_probe() {
```
davidkohn88, line 4: Add a comment about why this is necessary?

extension/src/triggers.rs
```diff
@@ -0,0 +1,56 @@
...
+    SELECT top_dep.objid, top_dep.deptype, top_dep.refobjid
+    FROM pg_catalog.pg_depend top_dep
+    WHERE top_dep.refobjid=experimental_schema_id
+    AND top_dep.objid IN (
+      SELECT depend.refobjid dep_id
+      FROM pg_catalog.pg_depend depend
+      INNER JOIN (
+        SELECT obj.objid id
+        FROM pg_event_trigger_ddl_commands() as obj
+        WHERE NOT obj.in_extension
```
davidkohn88, line 40: Doesn't seem like there's a good thing for this, but I kinda wish we could exclude temporary objects from this as well.

extension/src/triggers.rs
```diff
@@ -0,0 +1,56 @@
...
+    RETURN;
+  END IF;
+
+  SELECT oid schema_oid
+  INTO experimental_schema_id
+  FROM pg_catalog.pg_namespace
+  WHERE nspname='timescale_analytics_experimental'
+  LIMIT 1;
+
+  FOR cmd IN
```
davidkohn88, line 30: Seems slightly weird to me to do this in a loop rather than an IF block with an exists? But it works I guess.

davidkohn88 (commented):
Looks okay, a few comments, some that need addressing and some that are just style things. @feikesteenbergen might also have some thoughts on this...

extension/src/triggers.rs
JLockerman: It does, it broke a bunch of our tests 😋

extension/src/triggers.rs
JLockerman: hmm, I don't care about that for this release, but maybe for a future version

extension/src/triggers.rs
davidkohn88: ha. well, I guess that works, but probably won't work in the future if we ever modify this?

JLockerman: also added a must-fail test

extension/src/triggers.rs
JLockerman: fixed

extension/src/triggers.rs
JLockerman: fixed

extension/src/lib.rs
```diff
@@ -8,8 +11,25 @@ mod type_builder;
...
+
+#[pg_guard]
+pub extern "C" fn _PG_init() {
+    GucRegistry::define_bool_guc(
+        "timescale_analytics_acknowledge_auto_drop",
+        "enable creation of auto-dropping objects using experimental timescale_analytics_features",
+        "experimental features are very unstable, and objects \
+            depending on them will be automatically deleted on extension update",
+        &EXPERIMENTAL_ENABLED,
+        //TODO should this be superuser?
```
davidkohn88, line 28: No. Won't work on Forge I think.

JLockerman: also now that it's not a different extension, I think it's fine

PR #89, opened 2021-03-02
Adding compound aggregate for uddsketch

WireBaron:
This change also fixes some errors in the udd sketch combining code.

### DIFF
--- a/crates/udd-sketch/src/lib.rs
+++ b/crates/udd-sketch/src/lib.rs
@@ -41,14 +41,13 @@ impl std::cmp::PartialOrd for SketchHashKey{
 
 impl SketchHashKey {
     /// This is the key corresponding to the current key after the SketchHashMap it refers to has gone through one compaction.
-    /// compact bucket index towards the smaller bucket size,
-    /// aka 0 on the number line, -INF on the bucket indicies
+    /// Note that odd buckets get combined with the bucket after them (i.e. old buckets -3 and -2 become new bucket -1, {-1, 0} -> 0, {1, 2} -> 1)
     fn compact_key(&self) -> SketchHashKey {
         use SketchHashKey::*;
 
         match *self {
-            Negative(x) => Negative(if x < 0 {x-1} else {x} /2),
-            Positive(x) => Positive(if x < 0 {x-1} else {x} /2),
+            Negative(x) => Negative(if x > 0 {x+1} else {x} /2),
+            Positive(x) => Positive(if x > 0 {x+1} else {x} /2),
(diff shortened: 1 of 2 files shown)


crates/udd-sketch/src/lib.rs
```diff
@@ -47,8 +47,8 @@ impl SketchHashKey {
         use SketchHashKey::*;
 
         match *self {
-            Negative(x) => Negative(if x < 0 {x-1} else {x} /2),
-            Positive(x) => Positive(if x < 0 {x-1} else {x} /2),
+            Negative(x) => Negative(if x > 0 {x+1} else {x} /2),
+            Positive(x) => Positive(if x > 0 {x+1} else {x} /2),
```
JLockerman, line 51: could you explain the change?

crates/udd-sketch/src/lib.rs
```diff
@@ -292,7 +292,7 @@ impl UDDSketch {
             self.compact_buckets();
         }
 
-        for entry in other.buckets.iter() {
+        for entry in target.buckets.iter() {
```
JLockerman, line 295: maybe we should just do `let other = other.clone()` above, so the original `other` is no longer reachable, should prevent bugs like this in the future 🤔

JLockerman (approved):

crates/udd-sketch/src/lib.rs
WireBaron: Certainly, the problem here is that I was combining bins the wrong direction...so I was pairing bins like this: ..., (-2, -1), (0, 1), (2, 3), ... when it should have been ..., (-3, -2), (-1, 0), (1, 2), ...

I've updated the comment on this function indicating the order of combination.

crates/udd-sketch/src/lib.rs
WireBaron: Good point, made this change

extension/src/uddsketch.rs
```diff
@@ -335,4 +370,57 @@ mod tests {
...
+                
+            let (test_value, test_error) = client
+                .select("SELECT \
+                    timescale_analytics_experimental.quantile(uddsketch, 0.9), \
+                    timescale_analytics_experimental.error(uddsketch) \
+                    FROM composite", None, None)
+                .first()
+                .get_two::<f64, f64>();
+
+            apx_eql(test_value.unwrap(), value.unwrap(), 0.0001);
```
davidkohn88, line 421: I think for uddsketch, the merges shouldn't matter and the result is deterministic whether or not we have the intermediate result, so the apx_eql shouldn't be necessary?

PR #92, opened 2021-03-04
Add compound aggregate for tdigest

WireBaron:
This change also modifies the tdigest internal implementation to
store weights as integers instead of floats.

### DIFF
--- a/extension/src/tdigest.rs
+++ b/extension/src/tdigest.rs
@@ -1,9 +1,5 @@
 
-use std::{
-    convert::TryInto,
-    mem::replace,
-    slice,
-};
+use std::{convert::TryInto, mem::replace, ops::Deref, slice};
 
 use serde::{Serialize, Deserialize};
 
@@ -107,6 +103,7 @@ pub fn tdigest_combine(
                 (None, Some(state2)) => Some(state2.clone().into()),
                 (Some(state1), None) => Some(state1.clone().into()),
                 (Some(state1), Some(state2)) => {
+                    assert_eq!(state1.digested.max_size(), state2.digested.max_size());
                     let digvec = vec![state1.digested.clone(), state2.digested.clone()];
                     if !state1.buffer.is_empty() {
                         digvec[0].merge_unsorted(state1.buffer.clone());  // merge_unsorted should take a reference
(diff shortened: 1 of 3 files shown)


extension/src/tdigest.rs
```diff
@@ -170,12 +170,34 @@ impl<'input> TDigest<'input> {
...
             *self.sum,
-            *self.count as f64,
+            *self.count,
             *self.max,
             *self.0.min,
             *self.max_buckets as usize
         )
     }
+
+    fn from_tdigest(digest: &InternalTDigest) -> TDigest<'input> {
```
davidkohn88, line 180: maybe from_internal_tdigest?

extension/src/tdigest.rs
```diff
@@ -245,6 +251,48 @@ CREATE AGGREGATE timescale_analytics_experimental.tdigest(size int, value DOUBLE
 );
 "#);
 
+#[pg_extern(schema = "timescale_analytics_experimental")]
+pub fn tdigest_compound_trans(
+    state: Option<timescale_analytics_experimental::TDigest<'static>>,
```
davidkohn88, line 256: I'd probably have the internal state still be an InternalTDigest, and let the value be the PG TDigest type.

davidkohn88: It seems like converting between the different forms could be bad, and also, it would make it easier to modify in place?

davidkohn88 (commented):
Mostly looks good a couple small things that you may want to address, but I also am not horribly great at reviewing Rust yet...

extension/src/tdigest.rs
WireBaron: Good call, changed.

extension/src/tdigest.rs
WireBaron: Agreed, this behaves much more reasonably as a InternalTDigest.

extension/src/tdigest.rs
```diff
@@ -245,6 +247,89 @@ CREATE AGGREGATE timescale_analytics_experimental.tdigest(size int, value DOUBLE
...
+extension_sql!(r#"
+CREATE AGGREGATE timescale_analytics_experimental.tdigest(
+    timescale_analytics_experimental.tdigest
+) (
+    sfunc = timescale_analytics_experimental.tdigest_compound_trans,
+    stype = internal,
+    finalfunc = timescale_analytics_experimental.tdigest_compound_final,
+    combinefunc = timescale_analytics_experimental.tdigest_compound_combine,
+    serialfunc = timescale_analytics_experimental.tdigest_compound_serialize,
+    deserialfunc = timescale_analytics_experimental.tdigest_compound_deserialize
```
davidkohn88, line 329: Need to add parallel = safe?

extension/src/tdigest.rs
```diff
@@ -245,6 +247,89 @@ CREATE AGGREGATE timescale_analytics_experimental.tdigest(size int, value DOUBLE
...
+) -> Option<Internal<InternalTDigest>> {
+    unsafe {
+        in_aggregate_context(fcinfo, || {
+            match (state1, state2) {
+                (None, None) => None,
+                (None, Some(state2)) => Some(state2.clone().into()),
+                (Some(state1), None) => Some(state1.clone().into()),
+                (Some(state1), Some(state2)) => {
+                    Some(InternalTDigest::merge_digests(
+                            vec![state1.deref().clone(), state2.deref().clone()]  // TODO: TDigest merge with self
```
davidkohn88, line 285: Maybe I missed this, but what happens if you try to merge digests that have different sizes specified? Will there be an error somewhere?

davidkohn88 (commented):
One small thing on the agg definition. And then I realized, we don't check for the same size at least in this code? does that happen elsewhere? Are there tests for it that I'm missing?

extension/src/tdigest.rs
WireBaron: Ah, thanks for the reminder.  Added to both aggregates here and also in uddsketch.

extension/src/tdigest.rs
WireBaron: I had thought this was handled in the merge_digests code, but looking closer, such merges are actually explicitly allowed.  This actually raises the question of whether we should actually support mixed size merge?  I've added asserts everywhere to fail for now, but we can discuss this further.

PR #93, opened 2021-03-09
 Twa improvements

davidkohn88:
Added docs and made some testing improvements, going to finish the testing up soon, but feeling a bit better about it, would love someone to read through the docs and give comments there.

### DIFF
--- a/extension/docs/time_weighted_average.md
+++ b/extension/docs/time_weighted_average.md
@@ -1,10 +1,11 @@
 # Time Weighted Average [<sup><mark>experimental</mark></sup>](/extension/docs/README.md#tag-notes)
 
 > [Description](#time-weighted-average-description)<br>
+> [Example Usage](time-weighted-average-examples)<br>
 > [API](#time-weighted-average-api) <br>
-> [Sample Usage](time-weighted-average-examples)<br>
-> [Interpolation Methods Details](#time-weight-methods)<br>
 > [Notes on Parallelism and Ordering](#time-weight-ordering)<br>
+> [Interpolation Methods Details](#time-weight-methods)<br>
+
 
 ## Description [](time-weighted-average-description)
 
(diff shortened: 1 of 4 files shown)


extension/docs/time_weighted_average.md
```diff
@@ -0,0 +1,254 @@
...
+
+## Description [](time-weighted-average-description)
+
+Time weighted averages are commonly used in cases where a time series is not evenly sampled, so a traditional average will give misleading results. Consider a voltage sensor that sends readings once every 5 minutes or whenever the value changes by more than 1 V from the previous reading. If the results are generally stable, but with some quick moving transients, a simple average over all of the points will tend to over-weight the transients instead of the stable readings. A time weighted average weights each value by the duration over which it occured based on the points around it and produces correct results for unevenly spaced series. 
+
+Timescale Analytics' time weighted average is implemented as an aggregate which weights each value either using a last observation carried forward (LOCF) approach or a linear interpolation approach ([see interpolation methods](#time-weight-methods)). While the aggregate is not parallelizable, it is supported with [continuous aggregation](https://docs.timescale.com/latest/using-timescaledb/continuous-aggregates). 
+
+Additionally, see the notes on
```
WireBaron, line 15: Assuming there's more coming here?

extension/docs/time_weighted_average.md
```diff
@@ -0,0 +1,254 @@
...
+<br>
+
+² Note that `ts` and `value` can be `null`, however the aggregate is not evaluated on `null` values and will return `null`, but it will not error on `null` inputs.
+
+### Returns
+
+|Column|Type|Description|
+|---|---|---|
+| `time_weight` | `TimeWeightSummary` | A TimeWeightSummary object that can be passed to other functions within the time weighting API. |
+<br>
```
WireBaron, line 51: I do think that having a short example for each command in-line in the documentation (in addition to the following section) is useful.

extension/docs/time_weighted_average.md
```diff
@@ -0,0 +1,254 @@
...
+    ts              TIMESTAMPTZ ,
+    val             DOUBLE PRECISION,
+    PRIMARY KEY (measure_id, ts)
+);
+
+```
+In this case we'll also make a hypertable partitioned on the ts column, with a relatively large chunk_time_interval because the data isn't too high rate, and it'll allow us to create [continuous aggregates](https://docs.timescale.com/latest/using-timescaledb/continuous-aggregates) later.
+
+```SQL
+SELECT create_hypertable('foo', 'ts', chunk_time_interval=> '15 days'::interval);
```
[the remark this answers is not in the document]
davidkohn88: I can move this to before the continuous aggregate example as that's why it's there, so I'll do the other examples and then convert to a hypertable later.

extension/docs/time_weighted_average.md
davidkohn88: yep, I noticed that right after I pushed :)

extension/docs/time_weighted_average.md
```diff
@@ -0,0 +1,254 @@
...
+Additionally, see the notes on 
+
+
+## Command List (A-Z) [](time-weighted-average-api)
+> - [time_weight() (point form)](#time_weight_point)
+> - [time_weight() (summary form)](#time-weight-summary) 
+> - [average()](#time-weight-average)
+
+---
+## **time_weight() (point form)** [](time_weight_point)
```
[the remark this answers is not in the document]
davidkohn88: okay, so you think move the command list to the end and have the examples earlier and maybe link to the commands in each of the examples?

extension/docs/time_weighted_average.md
davidkohn88: hmmm...yeah. I think that's probably true, it's just that they're kinda interlinked here, but I'll give it a shot.

extension/docs/time_weighted_average.md
```diff
@@ -0,0 +1,305 @@
...
+
+### Sample Usage
+```SQL
+WITH t as (
+    SELECT 
+        date_trunc('day', ts) as dt,   
+        timescale_analytics_experimental.time_weight('Linear', ts, val) AS tw -- get a time weight summary
+    FROM foo
+    WHERE id = 'bar'
+    GROUP BY date_trunc('day')
```
JLockerman, line 183: We use `date_trunc` here and `time_bucket` above

extension/docs/time_weighted_average.md
```diff
@@ -0,0 +1,305 @@
...
+), q as (
+    SELECT timescale_analytics_experimental.time_weight(tw) AS full_tw -- do a second level of aggregation to get the full time weighted average
+    FROM t
+)
+SELECT 
+    dt, 
+    timescale_analytics_experimental.average(tw),  -- extract the average from the time weight summary
+    timescale_analytics_experimental.average(tw) / (SELECT full_tw FROM q LIMIT 1)  as normalized -- get the normalized average
+FROM t;
+```
```
JLockerman, line 193: In general, I think this example would be clearer with the input out put pair; something like (note I have _not_ checked my math)
```SQL
WITH t as (
    SELECT 
        time_bucket('1 day'::interval, ts) as dt,   
        timescale_analytics_experimental.time_weight('Linear', ts, val) AS tw -- get a time weight summary
    FROM (VALUES
        ('2020-01-01 00:00',  0),
        ('2020-01-01 12:00', 20),
        ('2020-01-02 00:00', 5),
        ('2020-01-02 00:00', 10)
    ) points(ts, val)
    WHERE id = 'bar'
    GROUP BY time_bucket('1 day'::interval, ts)
) 
SELECT 
    dt, 
    timescale_analytics_experimental.average(tw) -- extract the average from the time weight summary
FROM t;
```
```output
 average
----------
      15
```

JLockerman (commented):
Putting down some non-blocking thoughts for posterity

PR #97, opened 2021-04-01
Counter agg

davidkohn88:


crates/counter-agg/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "counter-agg"
+version = "0.1.0"
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+serde = { package = "serde", version = "1.0", optional = false, default-features = false }
```
WireBaron, line 9: Can this just be `serde = "1.0"`, like in the extension cargo.toml

crates/counter-agg/src/lib.rs
```diff
@@ -0,0 +1,258 @@
...
+// you can always subtract a common near value from all your times, then add it back in, the regression analysis will be unchanged.
+// TODO Should this be a method?
+fn ts_to_xy(pt: TSPoint) -> XYPair{
+    XYPair{
+        x: pt.ts as f64,
+        y: pt.val,
+    }
+}
+
+pub fn to_seconds(t: f64)-> f64{
```
WireBaron, line 39: Not sure if this needs to be public, but if so, perhaps `us_to_seconds` would be a more descriptive name.

crates/counter-agg/src/lib.rs
```diff
@@ -0,0 +1,258 @@
...
+    t / 1_000_000 as f64// by default postgres timestamps have microsecond precision
```
WireBaron, line 40:
```suggestion
    t / 1_000_000.0 // by default postgres timestamps have microsecond precision
```

crates/counter-agg/src/lib.rs
```diff
@@ -0,0 +1,258 @@
...
+pub mod range;
+mod tests;
+
+#[derive(Debug, PartialEq)]
+pub enum CounterError{
+    OrderError,
+    BoundsInvalid,
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
```
WireBaron, line 15: Perhaps a comment here or at the top of the file explaining a bit better what these are and how they work (i.e. track monotonically increasing sensors which need to handle resets).

crates/counter-agg/src/lib.rs
```diff
@@ -0,0 +1,258 @@
...
+        self.regress.accum(incoming_xy).unwrap();
+        Ok(())
+    }
+
+    fn single_value(&self) -> bool {
+        self.last == self.first
+    }
+
+    // combining can only happen for disjoint time ranges 
+    pub fn combine(&mut self, incoming: &CounterSummary) -> Result<(), CounterError> {
```
WireBaron, line 90: I think `append` better captures the behavior and requirements of this function.

crates/counter-agg/src/lib.rs
```diff
@@ -0,0 +1,258 @@
...
+    pub fn time_delta(&self) -> f64{
+        to_seconds((self.last.ts - self.first.ts) as f64)
+    }
+
+    pub fn delta(&self) -> f64 {
+        self.last.val + self.reset_sum - self.first.val 
+    }
+
+    pub fn rate(&self) -> Option<f64> {
+        if self.last.ts == self.first.ts {
```
WireBaron, line 134:
```suggestion
        if self.single_value() {
```

crates/counter-agg/src/range.rs
```diff
@@ -0,0 +1,194 @@
...
+impl I64Range {
+    pub fn has_infinite(&self)-> bool{
+        self.left.is_none() || self.right.is_none()
+    }
+    
+    pub fn is_valid(&self) -> bool{
+        match (self.left, self.right) {
+            (None, _) => true,
+            (_, None) => true,
+            (Some(a), Some(b)) => a <= b,
```
WireBaron, line 24:
```suggestion
            (Some(a), Some(b)) => a <= b, 
            _ => true,
```
Suggestion applies to the following functions as well.

Edit: suggestion is to drop the first two cases of the match in favor of the new catchall, doesn't look like github is properly capturing that.

crates/counter-agg/src/range.rs
```diff
@@ -0,0 +1,194 @@
...
+        };
+        self.right = match (self.right, other.right) {
+            (None, _) => None,
+            (_, None) => None,
+            (Some(a), Some(b)) => Some(max(a, b))
+        };
+    }
+
+    pub fn contains(&self, pt: i64) -> bool {
+        (match self.left {
```
WireBaron, line 51: This would look cleaner as a single match like the preceding functions.

crates/counter-agg/src/lib.rs
davidkohn88: Hmmm...I feel like append seems like you're just pushing another into a list of them or something, combine seems like it works better for me, I suppose maybe `ordered_combine` or something. I do see where you're coming from though, I just think append is worse, but I think there's probably a better name than either of them out there...I just can't think of it :)

crates/counter-agg/src/range.rs
davidkohn88: Did this for this one and the next, but left it alone for a couple others where I thought it was more clear what the logic was in the other way.

PR #101, opened 2021-04-07
Add LTTB implementation

JLockerman:
Largest Triangle Three Buckets is a downsampling method that attempts to preserve visual similarity with the original data. This PR adds a simple implementation that ingests and sorts the entire data range. We may eventually want to add a version that takes sorted input to better handle large data sets.

This PR also fixes some bugs in our docs.


extension/src/lttb.rs
```diff
@@ -0,0 +1,219 @@
...
+}
+
+extension_sql!(r#"
+CREATE TYPE timescale_analytics_experimental.SortedTimeseries;
+"#);
+
+pg_type! {
+    #[derive(Debug)]
+    struct SortedTimeseries {
+        num_points: u64,  // required to be aligned
```
WireBaron, line 29: Would you be losing anything by making this a general Timeseries object with a 32 bit size and a bit indicating that it's sorted?

WireBaron (approved):
This seems good, though I didn't really look at the lttb code in terms of correctness.  One change I would like to see is to refactor this with the time series bits in asap.rs, but that's certainly not something that has to happen before the 0.2 release.

extension/src/lttb.rs
JLockerman: 🤔 we'd have to do something like
```rust
SortedTimeseries {
    num_points: u32,
    sorted: bool,
    padding: [u8; 3],
    points: [TsPoint; self.num_points],
}
```
but it _should_ work.

That said, I'm not sure if it's all that useful; for one, I'm not sure yet if we ever want to return an unsorted timeseries back to the user, for another, I think knowing that a timeseries is sorted at the type level is useful, it means you don't have to do a defensive re-sort if you want sorted output after `unnest`. Since there's no real way to do a data-dependent sort in SQL, I think storing it internally would force the user to always sort the output of LTTB. (Another option is to add a `sort_timeseries()` function, but that might force us to re-allocate the timeseries even if it's already sorted, I'm not confident enough about the expression context to say whether it is or isn't required)

JLockerman:
bors retry

JLockerman:
bors retry

PR #106, opened 2021-04-12
Switch to varlena_type!() to construct out custom types

JLockerman:
Here's something for you guys to looks at. This commit switches over our type construction to a new `varlena_type!()` pseudo-macro in order to reduce the boilerplate needed to construct our custom types. This pseudo-macro causes our fork of `cargo-pgx` to generate the same SQL we're writing by hand now. Currently it needs to be put in the same hacky `timescale_analytics_experimental` pseudo-`mod` we use to make sure the types are namespaced correctly, not sure yet if I can fix that.

So what do you think, is this better or worse than what we had before? (If it's better, I'll probably move the `timescale_analytics_experimental` pseudo mod down to the type-definition in each file; I think it'll read better there)

### DIFF
--- a/extension/src/asap.rs
+++ b/extension/src/asap.rs
@@ -12,14 +12,6 @@ use flat_serialize::*;
 
 use time_weighted_average::tspoint::TSPoint;
 
-// hack to allow us to qualify names with "timescale_analytics_experimental"
-// so that pgx generates the correct SQL
-mod timescale_analytics_experimental {
-    pub(crate) use super::*;
-
-    varlena_type!(NormalizedTimeSeries);
-}
-
 // This is included for debug purposes and probably should not leave experimental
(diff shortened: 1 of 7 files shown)


WireBaron (commented):
These changes look nice as far as they go, I assume the rest of the changes are in your pgx repo?  It's not clear to me how this is used to create postgres code in the non-experimental namespace?

JLockerman:
@WireBaron yeah, the rest are in this commit: https://github.com/JLockerman/pgx/commit/fd865c34cd1d055e0e0c24405b4403e80f13509b

JLockerman:
> It's not clear to me how this is used to create postgres code in the non-experimental namespace?

If we don't put in a
```rust
mod timescale_analytics_experimental {
    // ...
}
```
it won'y be in the experimental schema

PR #114, opened 2021-04-21
Create bug report template

WireBaron:


davidkohn88 (commented):
LGTM, might be slightly involved for us, but is fine.

PR #118, opened 2021-04-23
Adding fuzzing test for tdigest

WireBaron:
This also includes fixes for numerous found issues.

### DIFF
--- a/extension/src/tdigest.rs
+++ b/extension/src/tdigest.rs
@@ -64,7 +64,8 @@ pub fn tdigest_trans(
         in_aggregate_context(fcinfo, || {
             let value = match value {
                 None => return state,
-                Some(value) => value,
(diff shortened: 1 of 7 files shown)


extension/src/tdigest.rs
```diff
@@ -64,7 +64,8 @@ pub fn tdigest_trans(
         in_aggregate_context(fcinfo, || {
             let value = match value {
                 None => return state,
-                Some(value) => value,
+                // NaNs are nonsensical in the context of a percentile, so exclude them
```
davidkohn88, line 67: Should probably add docs explaining this behavior

davidkohn88 (commented):
THis looked good to me, sorry it took so long, I see you just merged, but yeah, it's fine, on comment on adding docs about the behavior with NaNs

PR #119, opened 2021-04-27
Switch Uddsketch to delta-encoded varints

JLockerman:
Preliminary benchmarks suggest that this may give a 10x improvement on
uddsketch storage space.

STILL TO COME:
  - switch serialized form of the transition state to be the same
  - switch to zero-copy quantile algorithms


WireBaron (approved):

crates/udd-sketch/src/lib.rs
```diff
@@ -27,7 +26,7 @@ pub enum SketchHashKey {
 }
 
 // we're going to write SketchHashKey's to disk, so ensure they have no padding/are the correct size
```
WireBaron, line 28: This comment doesn't seem useful anymore

WireBaron (approved):
Looks like you need to provide a UddSketch implementation of `estimate_quantile_at_value` that maps to the top level implementation, or fix the test uses that try to call this.  Aside from that, this looks good.

JLockerman:
must've lost that

bors r+

JLockerman:
order by is a bit too weak

PR #123, opened 2021-05-04
Adding continuous agg example to UddSketch docs

WireBaron:
Also adding a simple utility to create a sample table.

### DIFF
--- a/extension/docs/uddsketch.md
+++ b/extension/docs/uddsketch.md
@@ -3,6 +3,7 @@
 > [Description](#uddsketch-description)<br>
 > [Details](#uddsketch-details)<br>
 > [Example](#uddsketch-example)<br>
+> [Example in a Continuous Aggregates](#uddsketch-cagg-example)<br>
 > [API](#uddsketch-api)
 
 ## Description <a id="uddsketch-description"></a>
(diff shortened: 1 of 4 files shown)


extension/docs/uddsketch.md
```diff
@@ -111,7 +112,59 @@ FROM daily_rain;
...
+CREATE TABLE test(time TIMESTAMPTZ, value DOUBLE PRECISION);
+SELECT create_hypertable('test', 'time');
+```
+
+Now we'll create a continuous aggregate which will group all the points for each week into a UddSketch:
+```SQL ,non-transactional,ignore-output
+CREATE MATERIALIZED VIEW weekly_sketch
+WITH (timescaledb.continuous)
+AS SELECT
+    time_bucket('1 week'::interval, time) as week,
```
davidkohn88, line 129: I think time_bucket does weird stuff with weeks, could we do a couple days or a single day or something like that instead? (Just so we don't end up with people doing things that get odd results and complain)

extension/docs/uddsketch.md
```diff
@@ -111,7 +112,59 @@ FROM daily_rain;
...
+```SQL ,non-transactional
+INSERT INTO test
+    SELECT time, value
+    FROM timescale_analytics_experimental.generate_periodic_normal_series('2020-01-01 UTC'::timestamptz, NULL, NULL, NULL, NULL, NULL, NULL, rng_seed => 12345678); 
+```
+```
+INSERT 0 4032
+```
+
+Finally, we can query the aggregate to see various approximate percentiles from different weeks.
```
davidkohn88, line 145: This is nice. I'd also add an example of re-aggregation over the continuous aggregate with the uddsketch(uddsketch)) sort of thing. (If you switch to days rather than weeks, this becomes easier to do, I would be okay with days -> weeks, but we can't go weeks -> months cause time_bucket doesn't support that.

davidkohn88 (commented):
LGTM, but I would make a couple small changes to the example.

extension/src/utilities.rs
```diff
@@ -0,0 +1,63 @@
+use pgx::*;
+use pg_sys::{TimestampTz};
+
+#[pg_extern(schema = "timescale_analytics_experimental")]
+pub fn generate_periodic_normal_series(
```
davidkohn88, line 5: we will probably want some tests for this at some point, but for now, I think it's fine to be a bit vague as it's only a utility for internal use for now...

WireBaron:
I'm going to continue to work on this documentation and utility, but am merging the current state to unblock Josh.

PR #127, opened 2021-05-05
Fixing estimate_quantile_at_value

WireBaron:
Estimate_quantile_at_value will fail with values larger than
the sketch maximum.  This change fixes it to return 1.0 in
those cases.


davidkohn88 (approved):
LGTM

PR #133, opened 2021-05-11
Removing experimental tag from TDigest

WireBaron:
Also refactored some APIs to match UddSketch, added a continous aggregate example to the docs, improved the UddSketch docs a bit and added some usabllity to the generate_periodic_normal_series utility


extension/docs/tdigest.md
```diff
@@ -161,25 +280,25 @@ Get the minimum value from a t-digest.
...
- tdigest_min
--------------
-           1
+ get_min
+-----------
+         1
 ```
 ---
-## **tdigest_max** <a id="tdigest_max"></a>
+## **get_max** <a id="tdigest_max"></a>
```
davidkohn88, line 298: I think we may want to change this to max_val for consistency with num_vals?

extension/docs/tdigest.md
```diff
@@ -130,23 +206,66 @@ This will construct and return a TDigest with the specified number of buckets ov
...
 
-## **tdigest_min** <a id="tdigest_min"></a>
+## **tdigest (summary form)** <a id="tdigest-summary"></a>
+```SQL ,ignore
+tdigest(
+    digest TDigest
+) RETURNS TDigest
+```
+
+This will combine multiple already constructed TDigests, if they were created with the same size. This is very useful for re-aggregating digests already constructed using the [point form](#tdigest).
```
davidkohn88, line 229: I think we should probably add a note about how this potentially introduces more error and the non-deterministic nature of this sketch here?

davidkohn88: ie: Note that the tdigest approximation will not necessarily produce the same results with different ordered inputs, nor will it necessarily produce the same results when used with re-aggregation steps as if it were computed using the point form. Usually the differences will be minor, especially if there are relatively large numbers of values in each of the digests before re-aggregating.

extension/docs/uddsketch.md
```diff
@@ -242,7 +256,7 @@ This will combine multiple already constructed UddSketches, they must have the s
 | `uddsketch` | `UddSketch` | A UddSketch object which may be passed to other UddSketch APIs. |
 <br>
 
-### Sample Usages <a id="uddsketch-examples"></a>
+### Sample Usages <a id="uddsketch-summary-examples"></a>
 For this examples assume we have a table 'samples' with a column 'data' holding `DOUBLE PRECISION` values, and an 'id' column that holds the what series the data belongs to, we can create a view to get the UddSketches for each `id` using the [point form](#uddsketch-point) like so:
```
davidkohn88, line 260: seems like "this examples" is all over the place, may want to do a find/replace.

extension/docs/tdigest.md
```diff
@@ -90,21 +91,96 @@ FROM high_temp;
 (4 rows)
 ```
 
+## Example Using TimeScale Continuous Aggregates (tdigest-cagg-example)
+Timescale [continuous aggregates](https://docs.timescale.com/latest/using-timescaledb/continuous-aggregates)
+provide an easy way to keep a tdigest up to date as more data is added to a table.  The following example
```
davidkohn88, line 96: do we want to add make the types code ie `tdigest` rather than tdigest (seems like we do with a few other types like `DOUBLE PRECISION` elsewhere?

davidkohn88 (commented):
LGTM, a few nits, but nothing major.

PR #141, opened 2021-05-25
Stats agg

davidkohn88:


docs/rolling_average_api_working.md
```diff
@@ -0,0 +1,191 @@
+
+# Info dump on rolling average APIs #
+
+Rolling averages are currntly nasty to do with with timescaledb (user complaint https://news.ycombinator.com/item?id=27051005).  In our timeseries API we will eventually provide a function like
+```SQL
+moving_average(window => '30 minutes', slide => '5 minutes', data)
+```
+However, because set-returning aggregates cannot not exist in Postgres, this will not work outside of the timeseries API. Currently, doing rolling average properly requires windowed aggregates. In base SQL it is a real PITA because you have to do sum and count separately and then divide them yourself.
```
davidkohn88, line 8: We should probably also think about the continuous aggregate implications ie https://github.com/timescale/timescale-analytics/issues/138 in this document.

crates/counter-agg/src/lib.rs
```diff
@@ -21,14 +22,14 @@ pub struct CounterSummary {
...
-    pub regress: RegressionSummary,
+    pub stats: StatsSummary2D,
     pub bounds: Option<range::I64Range>,
 }
 
 // Note that this can lose fidelity with the timestamp, but it would only lose it in the microseconds, 
-// this is likely okay in most applications. However, if you need better regression analysis at the subsecond level, 
-// you can always subtract a common near value from all your times, then add it back in, the regression analysis will be unchanged.
-// Note that convert the timestamp into seconds rather than microseconds here so that the slope and any other regression analysis, is done on a per-second basis.
+// this is likely okay in most applications. However, if you need better statsion analysis at the subsecond level,
```
WireBaron, line 30: Comment got caught in search/replace

crates/stats-agg/src/stats1d.rs
```diff
@@ -0,0 +1,249 @@
...
+    }
+
+    fn has_infinite(&self) -> bool {
+        self.sx.is_infinite()
+            || self.sxx.is_infinite()
+    }
+
+    fn check_overflow(&self, old: &StatsSummary1D, p: f64) -> bool {
+        //Only report overflow if we have finite inputs that lead to infinite results.
+        (self.sx.is_infinite() || self.sxx.is_infinite()) && old.sx.is_finite() && p.is_finite()
```
WireBaron, line 72:
```suggestion
        self.has_infinite() && old.sx.is_finite() && p.is_finite()
```

crates/stats-agg/src/stats1d.rs
```diff
@@ -0,0 +1,249 @@
...
+            sx: self.sx - p,
+            sxx: 0.0, // initialize this for now.
+        }; 
+        let tmpx = p * self.n64() - self.sx;
+        let scale = 1.0 / (self.n64() * new.n64());
+        new.sxx = self.sxx - tmpx * tmpx * scale;
+        Some(new) 
+    }
+
+    // convenience function for testing.
```
WireBaron, line 123: I think you can use the attribute `#[cfg(test)]` to capture this.

crates/stats-agg/src/stats1d.rs
```diff
@@ -0,0 +1,249 @@
...
+#[derive(Debug, PartialEq, Copy, Clone, Serialize, Deserialize)]
+#[repr(C)]
+pub struct StatsSummary1D {
+    pub n: u64,
+    pub sx: f64,
+    pub sxx: f64,
+}
+
+impl StatsSummary1D{
+    pub fn n64(&self) -> f64 {
```
WireBaron, line 14: This name seems a bit weird...does it need to be `pub`?

extension/src/stats_agg.rs
```diff
@@ -0,0 +1,820 @@
...
+                },
+            }
+        })
+    }
+}
+
+
+#[pg_extern(schema = "timescale_analytics_experimental")]
+pub fn stats1d_summary_trans<'s, 'v>(
+    state: Option<Internal<StatsSummary1D<'s>>>,
```
WireBaron, line 262: It seems like the logic of all of these pg aggregate functions would be streamlined if the internal representation was a `StatsSummary1DInternal`.  This would require a separate trans function for the rollup, but still seems like it might be cleaner.

PR #146, opened 2021-05-27
Use rollup() for combining functions

JLockerman:
This commit adds a general `rollup()` function that can be used to combine
instances of already aggregated data into a larger aggregate, eg.
combining multiple t-digests into a single t-digest. We find that
otherwise reaggregating reads like a stutter, compare for instance

-   `tdigest(tdigest)` vs `rollup(tdigest`) or
-   `precentile_agg(percentile_agg)` vs `rollup(percentile_agg)`


JLockerman:
bors retry

JLockerman:
bors retry

PR #155, opened 2021-06-08
Update Readme.md

JLockerman:
Note that `rustfmt` is required to build the extension.

bors r+


JLockerman:
(bors was down today, so it took a couple of tries)

PR #159, opened 2021-06-11
Adding timeseries aggregate

WireBaron:
This change adds a postgres aggregate to construct a timeseries object directly
from a time and value column.  It also adds implementations for LTTB and ASAP
to operate directly on such timeseries aggregates.


JLockerman (approved):
bors r+

PR #167, opened 2021-06-24
Add time-weighted average UI tests

JLockerman:
This PR adds UI tests time-weighted average, ensuring that our textual representation of them will remain consistent.

This PR also adds textual serialization and deserialization to `TSPoint` when it's being in/output to a human-readable format. Outputting the integer representation of Postgres timestamps isn't particularly usable to human beings, nor consumable by non-Postgres software, and this work better for both vases.

The translation is done by an in-tree copy of `timestamptz_in`/`out`. Since we only link to Postgres in the top-level extension crate, we currently use undefined symbol hackery in crates/time-series to make this work. Long-term, we should likely move it, and its transitive dependencies into the extension crate.

### DIFF
--- a/crates/time-series/src/lib.rs
+++ b/crates/time-series/src/lib.rs
@@ -1,9 +1,9 @@
 
-use serde::{Deserialize, Serialize};
+use serde::{Deserialize, Serialize, ser::SerializeStruct};
 
-use std::borrow::Cow;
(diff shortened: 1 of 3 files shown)


JLockerman:
whoops, forgot to stage one of the files 😅

bors retry

PR #175, opened 2021-07-02
Use HyperLogLog++ implementation

JLockerman:
This commit changes our hyperloglog() function to use a HyperLogLog++ implementation. HyperLogLog++ provides more accurate counts than vanilla HyperLogLog, especially on the low end of data sizes.

Supersedes PR https://github.com/timescale/timescaledb-toolkit/pull/170


JLockerman:
bors retry

k-bx:
Apologies if this is already covered somewhere or I've missed in the PR, but what would be the upgrade path to those of us already using HLL, exactly? Will it "just work" or are there steps needed? Thanks

JLockerman:
@k-bx it depends on what your usage is:

- if you're just calling the HLL function, it will just work, no need to change your code right now
- continuous aggregates and similar things storing HLL objects need to be rebuilt from the original (this is what experimental features are for, we label them experimental so we can get them to you really quickly without having to worry about maintaining backwards compatibility)

k-bx:
@JLockerman asking for continuous aggregates. Yeah, I don't mind rebuilding them. Was just making sure having old (HLL, not HLL++) aggregates would not break timescale, or whether I'd need to drop those materialized views before upgrading the toolkit

JLockerman:
@k-bx Everything that doesn't use the experimental features should continue working just fine. The continuous aggregate _might_ end up in a strange state where it drops the columns that depend on experimental features but the other columns remain; it did so at one point in the past, and I'm not sure if that was considered expected behavior or if the entire aggregate was supposed to be dropped. Either way recreating the aggregate should fix the issue.

----

Anyway: we'd love to hear your thoughts on our hyperloglog implementation! This change is the last once we have planned, so if it's working for users we'll try to stabilize it for 1.2.0. Please let us know what you think!

k-bx:
@JLockerman thanks for pointing this out, I'm happy to test it out this week on both my local machine and our prod data

PR #186, opened 2021-07-27
Br/pipeline downsampling

WireBaron:
This change adds a new type of timeseries, GappyNormalTimeseries, and two new pipeline elements: Downsample and Gapfill.  There still needs to be some discussion on the final API for these elements, in particular Gapfill can't currently handle an ExplicitTimeseries, but this PR should serve as a starting point.

### DIFF
--- a/extension/src/time_series/pipeline.rs
+++ b/extension/src/time_series/pipeline.rs
@@ -23,6 +23,11 @@ pg_type! {
                 interval: i64,
                 downsample_method: i64,
             },
+            // TODO: this currently only works with GappyNormalSeries (and as a noop on NormalSeries),
+            // we need to decide on what the semantics of this are with regard to an ExplicictSeries
+            Gapfill: 3 {
+                gapfill_method: i64,
+            },
         },
     }
 }
@@ -106,6 +111,12 @@ pub fn execute_pipeline_element<'s, 'e>(
         }
         (Element::Downsample{..}, Owned(timeseries)) => {
             return Some(downsample(timeseries, &element));
+        }
+        (Element::Gapfill{..}, Borrowed(timeseries)) => {
+            return Some(gapfill(timeseries, &element));
+        }
+        (Element::Gapfill{..}, Owned(timeseries)) => {
+            return Some(gapfill(timeseries, &element));
         }
     }
 }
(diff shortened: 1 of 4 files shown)


extension/src/time_series/pipeline.rs
```diff
@@ -182,6 +204,302 @@ pub fn lttb_pipeline_element<'p, 'e>(
...
+        match self {
+            DownsampleMethod::Average => 1,
+            DownsampleMethod::WeightedAverage => 2,
+            DownsampleMethod::Nearest => 3,
+        }
+    }
+}
+
+impl DownsampleMethod {
+    pub fn process(&self, vals: &std::vec::Vec<TSPoint>, target: i64, interval: i64) -> TSPoint {
```
JLockerman, line 237:
```suggestion
    pub fn process(&self, vals: &[TSPoint], target: i64, interval: i64) -> TSPoint {
```

extension/src/time_series/pipeline.rs
```diff
@@ -182,6 +204,302 @@ pub fn lttb_pipeline_element<'p, 'e>(
...
+                    let weight = (ts - target).abs() as f64 / interval as f64 / 2.0;
+                    sum += val * weight;
+                    wsum += weight;
+                }
+                TSPoint{ts: target, val: sum / wsum as f64}
+            }
+            DownsampleMethod::Nearest => {
+                let mut closest = i64::MAX;
+                let mut result = 0.0;
+                for TSPoint{ts, val} in vals.iter() {
```
JLockerman, line 259: can we use a binary search for this?

extension/src/time_series/pipeline.rs
```diff
@@ -182,6 +204,302 @@ pub fn lttb_pipeline_element<'p, 'e>(
...
+            match series.series {
+                SeriesType::GappyNormalSeries{start_ts, step_interval, count, present, values, ..} => {
+                    match self {
+                        GapfillMethod::LOCF => {
+                            let mut results = Vec::new();
+                            let mut last_val = 0.0;
+                            let mut vidx = 0;
+
+                            for pidx in 0..count {
+                                if present[pidx as usize / 64] & 1 << (pidx % 64) != 0 {
```
JLockerman, line 400: this should be a method

extension/src/time_series/pipeline.rs
```diff
@@ -182,6 +204,302 @@ pub fn lttb_pipeline_element<'p, 'e>(
...
+            if current != None {
+                let new_pt = method.process(&points, current.unwrap(), interval);
+                match result {
+                    None => result = Some(InternalTimeSeries::new_gappy_normal_series(new_pt, interval)),
+                    Some(ref mut series) => series.add_point(new_pt),
+                }
+            }
+            
+            current = Some(target);
+            points = Vec::new();
```
JLockerman, line 349:
```suggestion
            vec.clear();
```

extension/src/time_series/pipeline.rs
WireBaron: Fixed

extension/src/time_series/pipeline.rs
WireBaron: fixed

extension/src/time_series/pipeline.rs
WireBaron: True, but that method belongs on the bitmap type that we need to create for the present array.  I'll save this change until we implement that, as otherwise I'm not really sure where this method should live.

extension/src/time_series/pipeline.rs
WireBaron: Doesn't look like any of the existing binary_search implementations will match this functionality (return the matching point if present, else return the closest value, in case of a tie return the average).  We could rewrite this to be more binary search-ish, but I'm not sure it's worth the added complexity unless we expect people to be using really populated buckets here.  Leaving this as a future optimization.

WireBaron (commented):
The issues raised are going to be addressed in the next PR.

PR #201, opened 2021-08-05
Adding initial draft for timeseries pipeline documentation

WireBaron:


docs/timeseries_pipeline_elements.md
```diff
@@ -0,0 +1,367 @@
...
+SELECT timeseries |> elementA |> elementB;
+```
+
+However, it is possible to explicitly group elements using parentheses:
+
+```SQL ,ignore
+SELECT timeseries |> (elementA |> elementB);
+```
+
+This will result in a pipeline object being created from elements A and B, which will then be applied to the timeseries.  While we don't presently take maximum advantage of this internally, these multiple element pipelines should enable optimizations moving forward.  Therefore, this second form should be preferred where possible.
```
ryanbooz, line 25: "Therefore, this second form should be preferred where possible"

This feels like guidance that's just way too general. Are there drawbacks? Is there a time you wouldn't want to do that? IDK, just not sure where to go with this as a user because I'm sure at some point it won't always work as I want??

Anyway, just some feedback to consider.

docs/timeseries_pipeline_elements.md
```diff
@@ -0,0 +1,367 @@
...
+SELECT setseed(0.456);
+INSERT INTO test_data 
+    SELECT
+        '2020-01-01 00:00:00+00'::timestamptz + ((random() * 2592000)::int * '1 second'::interval),
+        floor(random() * 10 + 1),
+        50 + random() * 20
+    FROM generate_series(1,10000);
+```
+
+Now suppose we want to know how much the temperature fluctuates on a daily basis for each device.  It is actually simple to do this using timeseries and pipelines:
```
ryanbooz, line 48:
```suggestion
Now suppose we want to know how much the temperature fluctuates on a daily basis for each device.  Using timeseries and pipelines can simplify the process of finding the answer:
```
(not fully happy with the suggestion, but we generally try to avoid saying things are "easy" and "simple". It's not to everyone. ;-)

docs/timeseries_pipeline_elements.md
```diff
@@ -0,0 +1,367 @@
...
+
+## **delta** <a id="timeseries_pipeline_delta"></a>
+```SQL ,ignore
+delta(
+) RETURNS TimeseriesPipelineElement
+```
+
+This element will return a new timeseries where each point is the difference between the current and preceeding value in the input timeseries.  The new series will be one point shorter as it will not have a preceding value to return a delta for the first point.
+
+### Required Arguments <a id="timeseries_pipeline_delta-arguments"></a>
```
ryanbooz, line 133: Just FYI, when you move these over to our docs, we use the form of

`[](#timeseries_pipeline_delta-arguments)` for anchor links because we have a few different ways of creating the in-page links so this allows consistency.

docs/timeseries_pipeline_elements.md
```diff
@@ -0,0 +1,367 @@
...
+|---|---|---|
+| `timeseries` | `Timeseries` | This creates a complete normal timeseries (no missing values) from the input series.. |
+<br>
+
+### Sample Usage <a id="timeseries_pipeline_fill_holes-examples"></a>
+```SQL
+SELECT time, value
+FROM toolkit_experimental.unnest_series(
+    (SELECT toolkit_experimental.timeseries('2020-01-01'::timestamptz + step * step * '1 hour'::interval, step * step) 
+        |> (toolkit_experimental.resample_to_rate('nearest', '1 hour', true)
```
ryanbooz, line 198: can this be addressed before adding it into an example?

docs/timeseries_pipeline_elements.md
```diff
@@ -0,0 +1,367 @@
...
+---
+
+## **lttb** <a id="timeseries_pipeline_lttb"></a>
+```SQL ,ignore
+lttb(
+    resolution int,
+) RETURNS TimeseriesPipelineElement
+```
+
+This element will return a [largest triangle three buckets](lttb.md#description) approximation of a given timeseries.  It's behavior is the same as the lttb function documented [here](lttb.md#lttb), save that it expects the series to be sorted.
```
ryanbooz, line 226:
```suggestion
This element will return a [largest triangle three buckets](lttb.md#description) approximation of a given timeseries.  Its behavior is the same as the lttb function documented [here](lttb.md#lttb), save that it expects the series to be sorted.
```

docs/timeseries_pipeline_elements.md
```diff
@@ -0,0 +1,367 @@
...
+
+---
+
+## **sort** <a id="timeseries_pipeline_sort"></a>
+```SQL ,ignore
+sort(
+) RETURNS TimeseriesPipelineElement
+```
+
+This element takes in a timeseries and returns a timeseries consisting of the same points, but in order of increasing time values.
```
ryanbooz, line 334: Always increasing? Ever a need for decreasing time? (mostly just curious)

WireBaron:
Not entirely happy with all the wording in the current version (the resample rate table under `resample_to_rate` in particular is a bit messy), but we can clean it up as we get feedback.  Failing that, we'll make a much more detailed pass before lifting this to the docs repo.

bors r+

Issue #202, opened 2021-08-12
Aggregate HyperLogLogs into another HyperLogLog


k-bx:
Hi. There is a common scenario when you have a per-hour materialized view with your HyperLogLog distinct stats data, and you want to further "merge" it into per-day or per-month results.

For example, let's consider "events" table which has a "device_id" field, and we want a materialized view which shows how many unique device IDs you've had every hour. Furthermore, you'd like to show stats for per-day and per-month unique devices on your portal.

The data can look something like this:

```
CREATE TABLE IF NOT EXISTS test_hll_events (
    time TIMESTAMP WITHOUT TIME ZONE NOT NULL,
    device_id text,
    game_id int8
);
SELECT create_hypertable('test_hll_events','time');
insert into test_hll_events values 
  ('2021-08-01 01:00:00', 'dev01', 1)
, ('2021-08-01 02:00:00', 'dev01', 2)
, ('2021-08-02 01:00:00', 'dev01', 2)
, ('2021-08-02 02:00:00', 'dev03', 3)
;
CREATE MATERIALIZED view test_hll_mv_devices_hourly  WITH (timescaledb.continuous) AS
  select time_bucket('1 hour', time) as bucket, toolkit_experimental.hyperloglog(32, device_id) as val from test_hll_events group by 1 
  with no data;
```

Now, what we want to see to get a daily view of unique devices is something like this:

```
select time_bucket('1 day', bucket), toolkit_experimental.hyperloglog_count(toolkit_experimental.rollup(val)) from test_hll_mv_devices_hourly group by 1 order by time_bucket asc;
```

This is currently not possible.

I will also make a PR implementing a solution that works but I think is not good enough (but is totally usable), but that should be discussed separately in that PR.

PR #203, opened 2021-08-12
Aggregate HyperLogLogs into another HyperLogLog

k-bx:
Hi. I'm making a PR implementing https://github.com/timescale/timescaledb-toolkit/issues/202. I'm not sure how useful it would be in terms of documenting and merging into TimescaleDB. It allows you to solve the problem described in the issue, but it does so:

- With a code that's probably shit. I didn't really get into details of how things work and just copy-pasted a bunch of things extending a 2-parameter function into a vec-function essentially. This is minor as this can be improved in this very PR
- With an approach that doesn't directly work as described in the issue, and requires an additional layer of an `array_agg` call, which consumes the group-by-streamed data into an array. This is not great it you have a stream of billions of HyperLogLogs, although I would assume that's not a very common scenario, and hey, maybe this will still work ok?

Usage of this solution would look like this (see https://github.com/timescale/timescaledb-toolkit/issues/202 for initial data and table structure):

```sql
select
  time_bucket('1 day', bucket), 
  toolkit_experimental.hyperloglog_count(
    toolkit_experimental.hyperloglog_union_many(array_agg(val))) 
from test_hll_mv_devices_hourly group by 1 order by time_bucket asc;
```

If it's ok to keep this function merged and present in the toolbox, as it's intended as an "experimental", or if there is no clear vision/understanding of how exactly to make a proper implementation of the feature in the near term, I'm happy to document and propose to merge this, as it would be able to help users solve specific problems they will have. Otherwise -- let's discuss how we want to reimplement this properly. Would it be a new aggregate with some name?

Thanks!

### DIFF
--- a/extension/src/hyperloglog.rs
+++ b/extension/src/hyperloglog.rs
@@ -214,8 +214,46 @@ pub fn hyperloglog_union<'input>(
     flatten_log(&mut a)
 }
 
-fn flatten_log(hyperloglog: &mut HLL<Datum, DatumHashBuilder>)
--> toolkit_experimental::HyperLogLog<'static> {
+#[pg_extern(schema = "toolkit_experimental")]
+pub fn hyperloglog_union_many<'input>(
+    // bs: pgx::Array<toolkit_experimental::HyperLogLog<'input>>,
+    bs: Vec<Option<toolkit_experimental::HyperLogLog<'input>>>,
+) -> Option<toolkit_experimental::HyperLogLog<'static>> {
+    let mut rv: Option<toolkit_experimental::HyperLogLog<'static>> = None;
+    for b in bs {
+        match rv {
+            None => {
+                if let Some(b) = b {
+                    rv = Some(flatten_log(&mut unflatten_log(b.clone())));
+                }
+            }
+            Some(a) => {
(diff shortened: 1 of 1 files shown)


JLockerman:
@k-bx  Thank you so much for the PR, we're excited to have you contributing 🎉 

For arrays versus aggregates: do you have a use case in mind for union'ing over arrays of hyperloglogs or is it just an ease-of-implementation thing? If it's the latter I'll try to sit down tomorrow and see what turning this into an aggregate would take. (`hyperloglog_union`/`rollup` was _supposed_ to be that aggregate but... well, actually I'm not sure what happened there 🤔 )

k-bx:
@JLockerman yeah, just the fact that it’s my first pgx/Postgres thing, and I might be ready to upgrade this PR to make this an aggregate, but I don’t think I’ll have time in the near days, so if TimescaleDB team can get to that sooner — that’d be great

JLockerman:
@k-bx would you take a look at https://github.com/timescale/timescaledb-toolkit/pull/210 see if it satisfies your needs? (sorry, took me a bit longer to get to this than I'd hoped)

PR #209, opened 2021-08-18
Update version to 1.2

JLockerman:
Preparing for the next release cycle.

bors r+


JLockerman:
cc @WireBaron @davidkohn88 you should see how this is done

PR #211, opened 2021-08-20
Implemention of SpaceSaving algorithm for TopN

WireBaron:
This adds a TopN implementation to Postgres using the SpaceSaving algorithm.

Currently this is only implemented for integer data, and should be generalized to any type before it's ready for being stabilized.

### DIFF
--- a/extension/Cargo.toml
+++ b/extension/Cargo.toml
@@ -28,6 +28,7 @@ stats_agg = {path="../crates/stats-agg"}
 time_weighted_average = {path="../crates/time-weighted-average"}
 time_series = {path="../crates/time-series"}
 asap = {path="../crates/asap"}
+spacesaving = {path="../crates/spacesaving"}
 
 approx = {version = "0.4.0", optional = true}
 bincode = "1.3.1"
@@ -37,6 +38,7 @@ paste = "1.0"
 rand = { version = "0.8.3", features = ["getrandom", "small_rng"] }
 rand_distr = "0.4.0"
 rand_chacha = "0.3.0"
+itertools = "0.10.1"
 
 [dev-dependencies]
 pgx-tests = {git="https://github.com/JLockerman/pgx.git", branch="timescale2"}
--- a/extension/src/topn.rs
+++ b/extension/src/topn.rs
@@ -1,0 +1,337 @@
+
+use std::slice;
+
+use pgx::*;
+
+use flat_serialize::*;
+
+use itertools::multizip;
+
+use crate::{
+    aggregate_utils::in_aggregate_context,
+    json_inout_funcs,
+    flatten,
+    palloc::Internal, pg_type
+};
+
+use spacesaving::SpaceSaving;
+type InternalTopN = SpaceSaving<i64>;
+
+#[allow(non_camel_case_types)]
+type int = u32;
+
+#[allow(non_camel_case_types)]
+type bytea = pg_sys::Datum;
+
+#[pg_extern(immutable, parallel_safe, schema = "toolkit_experimental")]
+pub fn topn_serialize(
+    state: Internal<InternalTopN>,
+) -> bytea {
+    crate::do_serialize!(state)
+}
+
+#[pg_extern(immutable, parallel_safe, strict,schema = "toolkit_experimental")]
+pub fn topn_deserialize(
+    bytes: bytea,
+    _internal: Option<Internal<InternalTopN>>,
+) -> Internal<InternalTopN> {
+    crate::do_deserialize!(bytes, InternalTopN)
+}
+
+// PG object for topn
+pg_type! {
+    #[derive(Debug)]
+    struct TopN<'input> {
+        num_values: u32,
+        max_values: u32,
+        total_inputs: u64,
+        values: [i64; self.num_values],
+        counts: [u64; self.num_values],
+        overcounts: [u64; self.num_values],
+    }
+}
+
+json_inout_funcs!(TopN);
+
+// hack to allow us to qualify names with "toolkit_experimental"
+// so that pgx generates the correct SQL
+pub mod toolkit_experimental {
+    pub(crate) use super::*;
+    varlena_type!(TopN);
+}
+
+impl<'input> TopN<'input> {
+    fn to_internal_topn(&self) -> InternalTopN {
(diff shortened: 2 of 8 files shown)


extension/src/topn.rs
```diff
@@ -0,0 +1,325 @@
...
+
+#[pg_extern(immutable, parallel_safe, name="topn", schema = "toolkit_experimental")]
+pub fn topn_iter (
+    n: i32,
+    agg: toolkit_experimental::TopN<'static>,
+) -> impl std::iter::Iterator<Item = (name!(value,i64),name!(min_freq,f64),name!(max_freq,f64))> + '_ {
+    assert!(n <= agg.num_values as _);
+    let mut count = 0;
+    let total = agg.total_inputs as f64;
+    Box::new(multizip((agg.values.iter(), agg.counts.iter(), agg.overcounts.iter())).filter(move |_| {count = count + 1; count <= n}).map(move |x| {let (val, count, over) = x; (*val, (count-over) as f64 / total, *count as f64 / total)}))
```
JLockerman, line 214:
```suggestion
    multizip((agg.values.iter(), agg.counts.iter(), agg.overcounts.iter())).take(n).map(move |x| {let (val, count, over) = x; (*val, (count-over) as f64 / total, *count as f64 / total)})
```

extension/src/topn.rs
```diff
@@ -0,0 +1,325 @@
...
+pub fn num_vals(
+    agg: toolkit_experimental::TopN<'static>,
+) -> int {
+    agg.total_inputs as _
+}
+
+#[pg_extern(immutable, parallel_safe, name="topn", schema = "toolkit_experimental")]
+pub fn topn_iter (
+    n: i32,
+    agg: toolkit_experimental::TopN<'static>,
```
JLockerman, line 209:
```suggestion
    agg: toolkit_experimental::TopN<'input>,
```

This should _never_ be `'static` or the same as the output; that claims that the input argument will live forever when it only lives for this one call.

JLockerman (changes requested):

extension/src/topn.rs
```diff
@@ -0,0 +1,325 @@
...
+    assert!(n <= agg.num_values as _);
+    let mut count = 0;
+    let total = agg.total_inputs as f64;
+    Box::new(multizip((agg.values.iter(), agg.counts.iter(), agg.overcounts.iter())).filter(move |_| {count = count + 1; count <= n}).map(move |x| {let (val, count, over) = x; (*val, (count-over) as f64 / total, *count as f64 / total)}))
+}
+
+#[pg_extern(immutable, parallel_safe, schema = "toolkit_experimental")]
+pub fn guaranteed_topn(
+    n: i32,
+    agg: toolkit_experimental::TopN<'static>,
```
JLockerman, line 220:
```suggestion
pub fn guaranteed_topn<'input>(
    n: i32,
    agg: toolkit_experimental::TopN<'input>,
```

extension/src/topn.rs
```diff
@@ -0,0 +1,325 @@
...
+            return false;
+        }
+    }
+
+    true
+}
+
+#[pg_extern(immutable, parallel_safe, schema = "toolkit_experimental")]
+pub fn max_ordered_n(
+    agg: toolkit_experimental::TopN<'static>,
```
JLockerman, line 238:
```suggestion
pub fn max_ordered_n<'input>(
    agg: toolkit_experimental::TopN<'input>,
```

extension/src/topn.rs
```diff
@@ -0,0 +1,325 @@
...
+    serialfunc = toolkit_experimental.topn_serialize,
+    deserialfunc = toolkit_experimental.topn_deserialize,
+    parallel = safe
+);
+"#);
+
+//---- Available PG operations on the topn structure
+#[pg_extern(immutable, parallel_safe, schema = "toolkit_experimental")]
+pub fn num_vals(
+    agg: toolkit_experimental::TopN<'static>,
```
JLockerman, line 201:
```suggestion
pub fn num_vals<'input>(
    agg: toolkit_experimental::TopN<'input>,
```

extension/src/topn.rs
```diff
@@ -0,0 +1,325 @@
...
+    serialfunc = toolkit_experimental.topn_serialize,
+    deserialfunc = toolkit_experimental.topn_deserialize,
+    parallel = safe
+);
+"#);
+
+#[pg_extern(immutable, parallel_safe, schema = "toolkit_experimental")]
+pub fn topn_compound_trans(
+    state: Option<Internal<InternalTopN>>,
+    value: Option<toolkit_experimental::TopN<'static>>,
```
JLockerman, line 169:
```suggestion
pub fn topn_compound_trans<'b>(
    state: Option<Internal<InternalTopN>>,
    value: Option<toolkit_experimental::TopN<'b>>,
```

extension/Cargo.toml
```diff
@@ -37,6 +38,7 @@ paste = "1.0"
 rand = { version = "0.8.3", features = ["getrandom", "small_rng"] }
 rand_distr = "0.4.0"
 rand_chacha = "0.3.0"
+itertools = "0.7.8"
```
JLockerman, line 41:
```suggestion
itertools = "0.10.1"
```

extension/src/topn.rs
```diff
@@ -0,0 +1,325 @@
...
+    agg: toolkit_experimental::TopN<'static>,
+) -> int {
+    for i in 1..agg.num_values as usize {
+        if agg.counts[i] > agg.counts[i-1] - agg.overcounts[i-1] {
+            return (i - 1) as _;
+        }
+    }
+
+    (agg.num_values - 1) as _
+}
```
JLockerman, line 247: Note for the future: we probably want a `top_guaranteed()` which outputs elements until this loop condition is false.

extension/src/topn.rs
WireBaron: This actually couldn't be made non-static without breaking the compilation.  After discussing this we've come up with an explanation of why this is okay, which has been added as a comment here.

extension/src/topn.rs
WireBaron: Agreed.

PR #215, opened 2021-08-24
Add user-defined mapping elements

JLockerman:
This PR adds a number of utilities for user-defined pipeline elements:

1. `map_series()` allows users to use `fn(timeseries) returns timeseries` to create their own pipeline elements
2.  A wildly experimental `|>>` operator allows using custom pipeline elements almost as ergonomically as ones we define.
3. `map_data()` is a convenience function that allows users to apply `fn(double precision) returns double precision` to apply data-local transformations.

-----

**`map_series()`**

This function allows users to write their own pipeline elements by writing a SQL function `fn(timeseries) returns timeseries`. For example

```SQL
CREATE FUNCTION jan_3_x3(timeseries) RETURNS timeseries AS $$
    SELECT timeseries(time, value * 3)
    FROM (SELECT (unnest_series($1)).*) a
    WHERE time='2020-01-03 00:00:00+00';
$$ LANGUAGE SQL;

SELECT series |> map_series('jan_3_x3');
```

This greatly expands the power users have, and should enable them to fill in gaps in the timeseries API if needed.

-----

**`|>>`**

This operator, `|>>` can be used anywhere our other operator can except that right side is a regproc instead of a pipeline element eg. it looks like
```SQL
CREATE FUNCTION yesterday_squared(timeseries) RETURNS timeseries AS $$
    SELECT timeseries(time, value * value) \
    FROM (SELECT (unnest_series($1)).*) a \
    WHERE time>now() - '1 day'::interval;\
$$ LANGUAGE SQL;

SELECT series |>> 'yesterday_squared';
```

It is wildly experimental, even compared to the rest of the pipeline API. This PR creates a new operator instead of reusing the `|>` because if we use `|>` for both this and and the regular timeseries elements trying to do `series |> 'custom_element'` gets an ambiguous operator error `timeseries |> unknown` is not unique. In the future we could consider changing the element input function to fallback to checking if the input is a regproc if it doesn't recognize it; the formats should be different enough that there's no risk of collision.

-----

**`map_data()`**

`map_data()` is a convenience element for users that just want to apply mappings over the data stored in a timesries; it takes a `fn(double precision) returns double precision` and applies it to every datapoint eg.
```SQL
CREATE FUNCTION x2(double precision) RETURNS DOUBLE PRECISION
AS 'SELECT $1 * 2;' LANGUAGE SQL;

SELECT series |> map_data('x2');
```

-----

These functions should add a lot more power to the pipeline API since with the addition of `map_series()` users can create their own timeseries pipeline elements if the find the preexisting set lacking. Though `map_series()` should be "complete" in that any element one wants could be defined in terms of `map_series()` we may want to consider adding other connivence methods for common interactions. In particular, a `map_points(timestamptz, double precision)` seems like a promising candidate.

-----

The PR contains ≈1 commit per function, though note that the first commit is `map_data()` as that was simplest to implement. The PR also contains some refactoring, notably switching our pipeline executor to take in timeseries by-value; this was needed to make user-defined elements work.

### DIFF
--- a/extension/src/time_series/pipeline.rs
+++ b/extension/src/time_series/pipeline.rs
@@ -28,6 +28,12 @@ use resample_to_rate::{
 use sort::sort_timeseries;
 use delta::timeseries_delta;
 
+use map::{
+    map_series_pipeline_element,
+    check_user_function_type,
+    apply_to_series,
+};
+
 use crate::serialization::PgProcId;
 
 // TODO once we start stabilizing elements, create a type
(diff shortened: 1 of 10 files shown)


extension/src/time_series/pipeline.rs
```diff
@@ -228,6 +234,136 @@ CREATE OPERATOR |> (
...
+    PROCEDURE=toolkit_experimental."build_unstable_pipepine_a",
+    LEFTARG=toolkit_experimental.UnstableTimeseriesPipelineElement,
+    RIGHTARG=regproc
+);
+
+CREATE OPERATOR |>> (
+    PROCEDURE=toolkit_experimental."build_unstable_pipepine_b",
+    LEFTARG=regproc,
+    RIGHTARG=toolkit_experimental.UnstableTimeseriesPipelineElement
+);
```
WireBaron, line 352: Not a fan of using the new operator for a non-regproc rhs.  I'd prefer to see this as `|>`.

WireBaron (commented):
For the `|>>` operator, can we key off the regproc signature to treat the element as a `map_series` or a `map_data`.

PR #218, opened 2021-08-26
Fix duration calculations in Prom extrapolation functions

cevian:
Prom calculates duration usinng inclusive values on both ends,
we store bounds exclusively on the right end. Thus we have to
adjust the right bound to it's inclusive value (in milliseconds)
before doing the duration calculation.


davidkohn88:
fixed by #223

PR #225, opened 2021-09-13
Add pipeline elements for PG functions

JLockerman:
This commit adds pipeline elements for most of [Table 9.5](https://www.postgresql.org/docs/12/functions-math.html#FUNCTIONS-MATH-FUNC-TABLE). These pipeline elements apply the function to each value in the timeseries. For unary functions the element takes no arguments, eg. `series |> abs()` for binary functions the element takes one argument, the second argument to the function eg. `series |> div(2)` divides every element in the timeseries in half.

### DIFF
--- a/extension/src/time_series/pipeline/arithmetic.rs
+++ b/extension/src/time_series/pipeline/arithmetic.rs
@@ -1,0 +1,284 @@
+
+use pgx::*;
+
+use super::*;
+
+use super::Element::Arithmetic;
+use Function::*;
+
+#[derive(Debug, Copy, Clone, flat_serialize_macro::FlatSerializable, serde::Serialize, serde::Deserialize)]
+#[repr(u64)]
+//XXX note that the order here _is_ significant
+pub enum Function {
+    // binary functions
+    Add = 1,
+    Sub = 2,
+    Mul = 3,
+    Div = 4,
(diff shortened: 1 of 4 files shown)


extension/src/time_series/pipeline/arithmetic.rs
```diff
@@ -0,0 +1,638 @@
...
+            ]");
+        });
+    }
+
+    #[pg_test]
+    fn test_simple_arith_unaryops() {
+        Spi::execute(|client| {
+            client.select("SET timezone TO 'UTC'", None, None);
+            // using the search path trick for this test b/c the operator is
+            // difficult to spot otherwise.
```
davidkohn88, line 472: I thought we'd decided that the operator was going to go in the non-schema qualified part?

davidkohn88 (commented):
Really cool and surprisingly simple. 
One question: how hard would it be to add a fuzzing test of some sort that compares the unnested timeseries to the pg output for each of the functions? Maybe one that doesn't run all the time if it might take a while? I dunno, seems like that could be generated programmatically and would just require mapping function names to function names in pg? We did a similar thing for stats_agg and it gave me much more confidence in them...If it's a lot of work might not be worth it, but seems like it could limit work for future additions?

PR #251, opened 2021-10-04
add to_epoch() utility function, fix #96

### LINKED ISSUE #96: Simplify Epoch Parsing
Postgres has a `to_timestamp` function that takes a double precision unix epoch, but uses `EXTRACT EPOCH FROM <timestamp>` to go the other way. I think making a `to_epoch` function to make that more parallel would be a better UI and would be a small lift.

```SQL
CREATE OR REPLACE FUNCTION to_epoch(timestamptz) RETURNS DOUBLE PRECISION LANGUAGE SQL IMMUTABLE PARALLEL SAFE AS $$
SELECT EXTRACT(EPOCH FROM $1)
$$;
```

burmecia:
This PR is to add an utility function `to_epoch()` as per #96 to simplify epoch parsing, test case is also added.

### DIFF
--- a/extension/src/utilities.rs
+++ b/extension/src/utilities.rs
@@ -87,3 +87,28 @@ pub fn generate_periodic_normal_series(
         (time, base + error)
     })
 }
+
+// Convert a timestamp to a double precision unix epoch
+extension_sql!(r#"
+CREATE OR REPLACE FUNCTION toolkit_experimental.to_epoch(timestamptz) RETURNS DOUBLE PRECISION LANGUAGE SQL IMMUTABLE PARALLEL SAFE AS $$
+SELECT EXTRACT(EPOCH FROM $1);
+$$;
+"#);
+
+#[cfg(any(test, feature = "pg_test"))]
+mod tests {
+    use std::time::{SystemTime, UNIX_EPOCH};
+    use pgx::*;
+
+    #[pg_test]
+    fn test_to_epoch() {
+        Spi::execute(|client| {
+            let test_val = client
+                .select("SELECT toolkit_experimental.to_epoch(now())", None, None)
+                .first()
+                .get_one::<f64>().unwrap();
+            let now =  SystemTime::now().duration_since(UNIX_EPOCH).unwrap();
+            assert!((now.as_secs_f64() - test_val).abs() < 0.05);
+        });
+    }
+}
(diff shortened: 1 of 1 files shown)


JLockerman:
@burmecia Thank your for the PR! I'm going to ask @davidkohn88 to review it since he's the one who opened the initial feature request.

extension/src/utilities.rs
```diff
@@ -87,3 +87,28 @@ pub fn generate_periodic_normal_series(
...
+
+    #[pg_test]
+    fn test_to_epoch() {
+        Spi::execute(|client| {
+            let test_val = client
+                .select("SELECT toolkit_experimental.to_epoch(now())", None, None)
+                .first()
+                .get_one::<f64>().unwrap();
+            let now =  SystemTime::now().duration_since(UNIX_EPOCH).unwrap();
+            assert!((now.as_secs_f64() - test_val).abs() < 0.05);
```
davidkohn88, line 111: Hmmm... this seems to be a test of how long this takes and whether the system clocks of our test computer and the one the server is running are synced nicely, I'm not sure that's exactly what we want to test. Would some tests of constant values maybe work better?

davidkohn88 (commented):
I think this test might be testing the wrong thing implicitly. Otherwise, looks good to me!

extension/src/utilities.rs
burmecia: It's not to test if time is in sync, using `now()` is just a simple way to get a timestamp to test if this function is working. As this unit test normally happens on a local server during development, I think the tiny time difference shouldn't be a problem. Anyway, I am happy to remove it if it is not appropriate and I've added another constant value check and some edge test cases.

davidkohn88: Yeah. I think sometimes using now() can cause problems and you *end up* testing the thing you don't want to. I think the constant timestamps should work nicely enough and not cause odd errors in edge cases where the server is far away...

burmecia: It's not to test if time is in sync, using now() is just a simple way to get a timestamp to test if this function is working. As this unit test normally happens on a local server during development, I think the tiny time difference shouldn't be a problem. Anyway, I am happy to remove it if it is not appropriate and I've added another constant value check and some edge test cases.

davidkohn88: I think removing it is probably best, these others should be fine! Just remove and then it should be good to merge! Thanks for the PR!

davidkohn88 (approved):

davidkohn88:
Looks good, going to merge. 

bors r+

PR #253, opened 2021-10-05
EXPERIMENT lambda pipeline element

JLockerman:
This PR adds an experimental lambda pipeline elements for `map` and `filter`. This element used like
```SQL
series -> map($$ ($time + '1 day'i, $value * 2) $$) -> filter($$ $time != '2020-01-01' and $value != 0 $$)
```
can be used to map arbitrary expression yielding either a `f64` or a `(TimestampTZ, f64)` pair over a timeseries, and filter timeseries based on an expression that returns TRUE if an element should be retained by the timeseries.

These elements are _very_ experimental, and if merged, would likely bake in experimental for a good while before they're ready for merge.

---

This PR is now in a reviewable state. It _could_ use a revalidation that the lifetime changes are correct.

<details><summary>old checklist</summary>
This PR is still incomplete as of yet, and there are a number of features which would ideally be added before merge:

- [x] `filter()` element.
- [x] order comparison operators: `<`, `>`, `<=`, `>=`.
- [x] mathematical functions such as `abs()` see [the postgres tables](https://www.postgresql.org/docs/12/functions-math.html#FUNCTIONS-MATH-FUNC-TABLEl).
- [x] prefix operators: `-`, `NOT`.
</details>

### DIFF
--- a/extension/src/time_series/pipeline/lambda.rs
+++ b/extension/src/time_series/pipeline/lambda.rs
@@ -1,0 +1,691 @@
+
+use pgx::*;
+
+use super::*;
+
+pub use executor::ExpressionExecutor;
+
+mod parser;
+mod executor;
+
+mod toolkit_experimental {
+    pub(crate) use super::*;
+
+    varlena_type!(Lambda);
+}
+
+
+//
+// lambda type
+//
+
+pg_type! {
+    #[derive(Debug)]
+    struct Lambda<'input> {
+        len: u32,
+        string: [u8; self.len],
+    }
+}
+
+
+impl<'input> InOutFuncs for Lambda<'input> {
+    fn output(&self, buffer: &mut StringInfo) {
+        use crate::serialization::{EncodedStr::*, str_to_db_encoding};
+
+        let stringified = std::str::from_utf8(self.string.as_slice()).unwrap();
+        match str_to_db_encoding(&stringified) {
+            Utf8(s) => buffer.push_str(s),
+            Other(s) => buffer.push_bytes(s.to_bytes()),
+        }
+    }
+
+    fn input(input: &std::ffi::CStr) -> Self
+    where
+        Self: Sized,
+    {
+        use crate::serialization::str_from_db_encoding;
+
+        let s = str_from_db_encoding(input);
+        // validate the string
+        let _ = parser::parse_expression(s);
+        unsafe {
+            flatten! {
+                Lambda {
+                    len: s.as_bytes().len() as _,
+                    string: s.as_bytes().into(),
+                }
+            }
+        }
+    }
+}
+
+impl<'a> LambdaData<'a> {
+    pub fn parse(&self) -> Expression {
+        parser::parse_expression(std::str::from_utf8(self.string.as_slice()).unwrap())
+    }
+}
+
+
+//
+// Direct lambda execution functions for testing
+//
+
+#[pg_extern(
+    stable,
+    parallel_safe,
+    schema="toolkit_experimental"
+)]
+pub fn bool_lambda<'s>(
+    lambda: toolkit_experimental::Lambda<'s>,
+    time: pg_sys::TimestampTz,
+    value: f64
+) -> bool {
+    let expression = lambda.parse();
+    if expression.expr.ty() != &Type::Bool {
+        panic!("invalid return type, must return a BOOLEAN for {:?}", expression)
+    }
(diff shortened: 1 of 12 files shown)


extension/src/time_series/pipeline/lambda/executor.rs
```diff
@@ -0,0 +1,334 @@
...
+        }
+    }
+
+    fn force_var(&mut self, i: usize, value: f64, time: i64) -> Value {
+        if let Some(value) = &self.var_vals[i] {
+            return value.clone()
+        }
+
+        let value = self.exec_expression(&self.exprs.variables[i], value, time);
+        self.var_vals[i] = Some(value.clone());
```
WireBaron, line 60: So variables are set to the result of evaluating their expressions on the first (time, value) pair?

extension/src/time_series/pipeline/lambda/executor.rs
```diff
@@ -0,0 +1,334 @@
...
+        use Function::*;
+        macro_rules! unary_function {
+            ($func:ident ( )) => {
+                {
+                    let then = self.exec_expression(&args[0], value, time).float();
+                    then.$func().into()
+                }
+            };
+        }
+        macro_rules! binary_function {
```
WireBaron, line 80: Unless you're planning on adding some binary functions other than `Log`, remove this macro.  (if you keep it, use it for `log`)

extension/src/time_series/pipeline/lambda/lambda_expr.pest
```diff
@@ -0,0 +1,47 @@
+calculation = _{ SOI ~ let_expr ~ EOI }
+let_expr = { ("let" ~ var ~ "=" ~ tuple ~ ";")* ~ tuple }
+tuple = { binops ~ ("," ~ binops)* }
+binops = { unary ~ (operation ~ unary)* }
+unary = _{ neg | not | term }
+neg = { "-" ~ unary }
```
WireBaron, line 6:
```suggestion
neg = { "-" ~ term }
```
I would think we don't want to support things like `-----$val`.  Same comment for `not` on the next line.  Unless we're planning on expanding our selection of unary ops in the future, in which case ignore this comment.

extension/src/time_series/pipeline/lambda.rs
```diff
@@ -0,0 +1,670 @@
...
+            interval_lambda_eq!(client, "'1 day'i + '1 day'i", "2 days");
+            interval_lambda_eq!(client, "'1 day'i + '1 week'i", "8 days");
+            interval_lambda_eq!(client, "'1 week'i - '1 day'i", "6 days");
+
+            interval_lambda_eq!(client, "'1 day'i * 3", "3 days");
+            interval_lambda_eq!(client, "4 * '1 day'i", "4 days");
+            interval_lambda_eq!(client, "'4 day'i / 4", "1 day");
+        });
+    }
+}
```
WireBaron, line 670: I'd also like to see a test of variables (i.e. `"let $foo = 5; $foo * $value"`).

extension/src/time_series/pipeline/lambda.rs
```diff
@@ -0,0 +1,670 @@
...
+
+
+//
+// Common types across the parser and executor
+//
+
+// expressions
+#[derive(Debug)]
+pub struct Expressions {
```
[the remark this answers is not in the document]
JLockerman: @WireBaron what do you think of renaming the current `Expression` to `ExpressionSegment` or `ExpressionWithoutLocals` and renaming this type to `Expression`? I think that captures the intended usage better 🤔

extension/src/time_series/pipeline/lambda/lambda_expr.pest
JLockerman: Unfortunately, I think `not not $foo` is a valid, if silly, expression in SQL so I think it’d be more surprising if we didn’t support it. (In general, most programming language seem to accept things like `!!foo` regardless of whether there’s a good reason to so.) (Long-term, if we ever extend this language to support postgres’s custom operators we’ll need to be able to support arbitrary combinations of prefix operators, and I think it’s better to define the grammar such that this is supported rather than risk backing ourselves into a corner)

extension/src/time_series/pipeline/lambda.rs
JLockerman: Huh, I thought I had one of those. Will add.

Issue #180, opened 2021-07-14
Gen_random_timeseries


davidkohn88:
Generating test data with various distributions is really a PITA, it'd be really nice to be able to generate random timeseries with different distributions, ie a normal, bimodal, random walk, uniform, exponential etc. 

I think we'd probably also want something to describe the spacing of the time points, so either regular or irregular with some parameters

Then I could insert that into a table and see if the analysis I've been trying to do actually catches what I'm trying to understand and what distributions it works for and what it doesn't. 

The api could look something like:
`gen_random_timeseries(seed, DistributionParams, SeriesParams)`

Where distribution params might be something like:
`normal(avg, stddev)`
And the series might look something like:
`regular(start, end, step)`
`irregular(start, end, num_points)`

I think the early stuff would just do regular timeseries and perhaps just expose the different distributions here: https://docs.rs/rand/0.6.0/rand/distributions/index.html

davidkohn88:
Related to #121

JLockerman:
I'd recomend the parameters just be a something a JSON like `'{normal: {avg:10.0, stddev: 5.0 }}', or whatever PostgresTypes generates. That'll make it a lot easier to add things, and less work to start out.

Issue #121, opened 2021-04-28
Data Generation Functions


davidkohn88:
It would be great to have some simple data generation functions for creating demo/test data for playing around with use cases. I'm not sure what exactly the parameters to this sort of function should be, but probably it would generate a random timeseries, with a step size for the timestamp and then a value that either followed a random walk, with bounds, or had data with a given distribution around a center or something like that, possibly a sinusoidal variation on small scale to simulate noise etc. 

I think this would be very useful for 
a) demonstrating the other functions and showing users how to use them
b) letting users generate their own test data sets for themselves

(Coupling this with UDAs would actually simulate inserts on a relatively frequent basis...interestingly enough).

JLockerman:
@mfreed also has a request for a better `generate_series()` which might be related.

PR #255, opened 2021-10-05
add gen_random_timeseries() prototype

burmecia:
This PR is to add a prototype for `gen_random_timeseries()` regarding #180 and #121. Currently only `Normal` and `Uniform` distributions are added, but can be added more in similar way in the future.

The interface is:

```sql
gen_random_timeseries(
  start::timestamptz,
  stop::timestamptz,
  step::interval,
  [dist_params::json],
  [rnd_seed::bigint]
)
```

- `start`, `stop` and `step` is same as [`generate_series`](https://www.postgresql.org/docs/13/functions-srf.html)
- `dist_param` is to specify distribution name and parameters, for example: `'{"normal": {"mean": 10.0, "std_dev": 2.3}}'`. Default: `'{"uniform": {"low": 0.0, "high": 1.0}}'`
- `rnd_seed` is random seed for [Pcg64Mcg](https://rust-random.github.io/rand/rand_pcg/type.Pcg64Mcg.html)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1343,6 +1343,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "rand_pcg"
+version = "0.3.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "59cad018caf63deb318e5a4586d99a24424a364f40f1e5778c29aca23f4fc73e"
+dependencies = [
+ "rand_core 0.6.2",
+]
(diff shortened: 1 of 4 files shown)


WireBaron:
We'll be looking at this and should get a review back to you early next week.

JLockerman:
I don't this implementation matches our needs for the API, going to close until we have the time to look further into it.

PR #256, opened 2021-10-07
Add `sum()`, `average()`, and `num_vals()` finalizers

JLockerman:
I think these are the finalizers which are usefully called directly on the timeseries; for other ones it's likely better to materialize the stats_agg and use its accessors.

bors r+


JLockerman:
bors retry

PR #257, opened 2021-10-07
Adding pipeline operators for stabilizing features.

WireBaron:
This change allows user to pipe their timevectors into counter_agg or
hyperloglog aggregates.

### DIFF
--- a/extension/src/hyperloglog.rs
+++ b/extension/src/hyperloglog.rs
@@ -271,6 +271,23 @@ pub fn hyperloglog_error<'input>(
     hyperloglogplusplus::error_for_precision(precision)
 }
 
+impl HyperLogLog<'_> {
+    pub fn build_from(size: int, type_id: Oid, collation: Option<Oid>, data: impl Iterator<Item=pg_sys::Datum>)
+    -> HyperLogLog<'static> {
+        unsafe {
+            let b = TryInto::<usize>::try_into(size).unwrap().checked_next_power_of_two().unwrap().trailing_zeros();
+            let hasher = DatumHashBuilder::from_type_id(type_id, collation);
+            let mut logger: HLL<pg_sys::Datum, DatumHashBuilder> = HLL::new(b as u8, hasher);
+                
+            for datum in data {
+                logger.add(&datum);
+            }
+
+            flatten_log(&mut logger)
+        }
+    }
+}
+
 fn flatten_log(hyperloglog: &mut HLL<Datum, DatumHashBuilder>)
 -> HyperLogLog<'static> {
     let (element_type, collation) = {
--- a/extension/src/time_series/pipeline/aggregation.rs
+++ b/extension/src/time_series/pipeline/aggregation.rs
@@ -8,6 +8,8 @@ use super::*;
 use crate::{
     ron_inout_funcs, pg_type, build,
     stats_agg::{self, InternalStatsSummary1D, StatsSummary1D},
+    counter_agg::{InternalCounterSummary, CounterSummary},
+    hyperloglog::HyperLogLog
 };
 
 
@@ -30,9 +32,9 @@ pub mod toolkit_experimental {
     varlena_type!(PipelineThenSum);
     varlena_type!(PipelineThenAverage);
     varlena_type!(PipelineThenNumVals);
+    varlena_type!(PipelineThenCounterAgg);
+    varlena_type!(PipelineThenHyperLogLog);
 }
-
-
 
 #[pg_extern(immutable, parallel_safe, schema="toolkit_experimental")]
 pub fn run_pipeline_then_stats_agg<'s, 'p>(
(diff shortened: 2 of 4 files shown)


extension/src/time_series/pipeline/aggregation.rs
```diff
@@ -410,6 +412,213 @@ extension_sql!(r#"
...
+
+#[pg_extern(immutable, parallel_safe, schema="toolkit_experimental")]
+pub fn run_pipeline_then_counter_agg<'s, 'p>(
+    mut timeseries: toolkit_experimental::TimeSeries<'s>,
+    pipeline: toolkit_experimental::PipelineThenCounterAgg<'p>,
+) -> CounterSummary<'static> {
+    timeseries = run_pipeline_elements(timeseries, pipeline.elements.iter());
+    if timeseries.num_points() == 0 {
+        panic!("Unable to create a counter summary from 0 points.");
+    }
```
JLockerman, line 433: I think this should just return `NULL` instead

```suggestion
    if timeseries.num_points() == 0 {
        return None
    }
```
if there are no relevant elements in a timeseries that shouldn't bring down the entire query.

extension/src/time_series/pipeline/aggregation.rs
```diff
@@ -410,6 +412,213 @@ extension_sql!(r#"
...
+pub fn run_pipeline_then_hyperloglog<'s, 'p>(
+    mut timeseries: toolkit_experimental::TimeSeries<'s>,
+    pipeline: toolkit_experimental::PipelineThenHyperLogLog<'p>,
+) -> HyperLogLog<'static> {
+    unsafe {
+        timeseries = run_pipeline_elements(timeseries, pipeline.elements.iter());
+        let size: usize = pipeline.hll_size.try_into().unwrap();
+        let b = size.checked_next_power_of_two().unwrap().trailing_zeros();
+        let hasher = DatumHashBuilder::from_type_id(PgBuiltInOids::FLOAT8OID as u32, None);
+        let mut logger: HLL<pg_sys::Datum, DatumHashBuilder> = HLL::new(b as u8, hasher);
```
JLockerman, line 536: I think this should be a function in hll; I think `DatumHashBuilder` is an internal implementation detail that shouldn't really be exposed...

extension/src/time_series/pipeline/aggregation.rs
```diff
@@ -491,19 +700,19 @@ mod tests {
...
+                ('2020-01-02 UTC'::TIMESTAMPTZ, 15.0), \
+                ('2020-01-05 UTC'::TIMESTAMPTZ, 30.0)) as v(time, value)";
+
+        let val = client.select(
+            &format!("SELECT (series -> sum())::TEXT FROM ({}) s", create_series),
+            None,
+            None
+        )
+            .first()
+            .get_one::<String>();
```
JLockerman, line 715: indent on this got messed up

extension/src/time_series/pipeline/aggregation.rs
```diff
@@ -670,4 +879,85 @@ mod tests {
...
+            let val = client.select(
+                &format!("SELECT series -> hyperloglog(100) -> distinct_count() FROM ({}) s", create_series),
+                None,
+                None
+            )
+                .first()
+                .get_one::<i32>().unwrap();
+            assert_eq!(val, 7);
+        })
+    }
```
JLockerman, line 962: If we're adding the pipeline-folding optimizations we should also add tests that they're working.

JLockerman (commented):
Generally looks. One semantic change that I think is important, and a couple of nits.

extension/src/time_series/pipeline/aggregation.rs
WireBaron: Hmm, I was thinking that `DatumHashBuilder` is eventually going to become its own public bit of infrastructure used in several different places (notably, I'm planning on making use of it for TopN).  However, I agree that here I'm using far too much HyperLogLog internal stuff, so I wrapped 533-542 into a `HyperLogLog::from_timeseries`.  I couldn't make it a proper `From` due to the size parameter, if you can think of a workaround there I'd love to see it.

JLockerman: I kinda meant just `533-536` just a `Hyperloglog::with_size(size: usize, type: Oid)`; I don't think hll should need to know anything about timevectors much like timevectors shouldn't need to know anything about hll sizing contraints.

extension/src/time_series/pipeline/aggregation.rs
JLockerman: there shouldn't be

extension/src/hyperloglog.rs
```diff
@@ -271,7 +272,25 @@ pub fn hyperloglog_error<'input>(
...
 
-pub fn flatten_log(hyperloglog: &mut HLL<Datum, DatumHashBuilder>)
+impl HyperLogLog<'_> {
+    // Can't quite implement a proper From here, since a size is also needed
+    pub fn from_timeseries(size: int, timeseries: &TimeSeries)
+    -> HyperLogLog<'static> {
+        unsafe {
+            let b = TryInto::<usize>::try_into(size).unwrap().checked_next_power_of_two().unwrap().trailing_zeros();
+            let hasher = DatumHashBuilder::from_type_id(PgBuiltInOids::FLOAT8OID as u32, None);
+            let mut logger: HLL<pg_sys::Datum, DatumHashBuilder> = HLL::new(b as u8, hasher);
```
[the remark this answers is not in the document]
WireBaron: After further discussion, implemented this as a `HyperLogLog::build_from(size, type, collation, iterator)`

PR #260, opened 2021-10-11
Rename `timeseries` to `timevector`

JLockerman:
Discussions with users suggest that `timeseries` is a bit too overloaded and can be confusing when also used for our datastructure.


WireBaron (approved):
Still not a fan of the way Timevector looks in camelcase, but I agree it's probably the more correct way of writing it.

PR #264, opened 2021-10-12
Adding new pipeline element `fill_to`

WireBaron:
This new element will take a method type and an interval.  It will then
scan through the passed timevector for any gaps larger than the given
interval, and will fill them according to the method type.

### DIFF
--- a/extension/src/time_series/pipeline.rs
+++ b/extension/src/time_series/pipeline.rs
@@ -1,5 +1,6 @@
 
 mod fill_holes;
+mod fill_to;
 mod resample_to_rate;
 mod sort;
 mod delta;
@@ -22,7 +23,12 @@ use crate::{
 
 use fill_holes::{
     fill_holes,
-    FillMethod,
+    FillHolesMethod,
+};
+
+use fill_to::{
+    fill_to,
+    FillToMethod,
 };
 
 use resample_to_rate::{
@@ -66,7 +72,7 @@ flat_serialize_macro::flat_serialize! {
             snap_to_rate: i64, // padded bool
         },
         FillHoles: 3 {
-            fill_method: FillMethod,
+            fill_method: FillHolesMethod,
         },
         Sort: 4 {
         },
@@ -89,6 +95,10 @@ flat_serialize_macro::flat_serialize! {
         },
         FilterLambda: 10 {
             lambda: lambda::LambdaData<'input>,
+        },
+        FillTo: 11 {
+            interval: i64,
+            fill_method: FillToMethod,
         },
     }
 }
@@ -176,6 +186,8 @@ pub fn execute_pipeline_element<'s, 'e>(
             return filter::apply_lambda_to(timevector, lambda),
         Element::Arithmetic{ function, rhs } =>
             return arithmetic::apply(timevector, *function, *rhs),
+        Element::FillTo{..} =>
+            return fill_to(timevector, &element),
     }
 }
 
--- a/extension/src/time_series/pipeline/fill_to.rs
+++ b/extension/src/time_series/pipeline/fill_to.rs
@@ -1,0 +1,238 @@
+
+use pgx::*;
+
+use flat_serialize_macro::FlatSerializable;
+
+use serde::{Deserialize, Serialize};
+
+use super::*;
+
+type Interval = pg_sys::Datum;
+
+// TODO: there are one or two other gapfill objects in this extension, these should be unified
+#[derive(Clone, Serialize, Deserialize, PartialEq, Debug, FlatSerializable)]
+#[repr(u64)]
+pub enum FillToMethod {
+    LOCF,
+    Interpolate,
+    Nearest,
+}
+
+impl FillToMethod {
+    pub fn fill_point(&self, lhs: &TSPoint, rhs: &TSPoint, target_ts: i64) -> TSPoint {
+        match *self {
+            FillToMethod::LOCF => TSPoint{ts: target_ts, val: lhs.val},
+            FillToMethod::Interpolate => {
+                let interval = rhs.ts as f64 - lhs.ts as f64;
+                let left_wt = 1. - (target_ts - lhs.ts) as f64 / interval;
+                let right_wt = 1. - (rhs.ts - target_ts) as f64 / interval;
+                TSPoint{ts: target_ts, val: lhs.val * left_wt + rhs.val * right_wt}
+            },
+            FillToMethod::Nearest =>
+                if rhs.ts - target_ts >= target_ts - lhs.ts {
+                    TSPoint{ts: target_ts, val: lhs.val}
+                } else {
+                    TSPoint{ts: target_ts, val: rhs.val}
+                },
+        }
+    }
+}
+
+// TODO is (immutable, parallel_safe) correct?
+#[pg_extern(
+    immutable,
+    parallel_safe,
+    name="fill_to",
+    schema="toolkit_experimental"
+)]
+pub fn fillto_pipeline_element<'e> (
+    interval: Interval,
+    fill_method: String,
+) -> toolkit_experimental::UnstableTimevectorPipeline<'e> {
+    unsafe {
+        let interval = interval as *const pg_sys::Interval;
+        // TODO: store the postgres interval object and use postgres timestamp/interval functions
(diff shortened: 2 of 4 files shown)


extension/src/time_series/pipeline/fill_to.rs
```diff
@@ -0,0 +1,248 @@
...
+        SeriesType::NormalSeries{step_interval, ..} => if step_interval <= interval { return series.clone(); },
+        _ => ()
+    }
+
+    let mut result: Option<Vec<TSPoint>> = None;
+    let mut it = series.iter().peekable();
+    let mut current = it.next();
+
+    while current.is_some() {
+        let lhs = current.unwrap();
```
JLockerman, line 94:
```suggestion
    while let (lhs) = current {
```

extension/src/time_series/pipeline/fill_to.rs
```diff
@@ -0,0 +1,248 @@
...
+            if result.is_none() {
+                result = Some(series.iter().collect());
+            }
+            let result = result.as_mut().unwrap();
+            let mut target = lhs.ts + interval;
+            while target < rhs.ts {
+                result.push(method.fill_point(&lhs, &rhs, target));
+                target += interval;
+            }
+        }
```
JLockerman, line 108:
```suggestion
        match it.peek() {
            Some(rhs) if rhs.ts - lhs.ts > interval => {
                let result = match &mut result {
                    Some(result) => result,
                    None => {
                        *result = Some(series.iter().collect());
                        result.as_mut().unwrap()
                    },
                }
                let mut target = lhs.ts + interval;
                while target < rhs.ts {
                    result.push(method.fill_point(&lhs, &rhs, target));
                    target += interval;
                }
            }
            _ => {},
        }
```

extension/src/time_series/pipeline/fill_to.rs
```diff
@@ -0,0 +1,248 @@
...
+        let lhs = current.unwrap();
+        let rhs = it.peek();
+
+        if rhs.is_some() && rhs.unwrap().ts - lhs.ts > interval {
+            let rhs = rhs.unwrap();
+            if result.is_none() {
+                result = Some(series.iter().collect());
+            }
+            let result = result.as_mut().unwrap();
+            let mut target = lhs.ts + interval;
```
JLockerman, line 103: we should add a
```
// FIXME use postgres interval ops if the interval contains anything larger than an hour
```

extension/src/time_series/pipeline/fill_to.rs
```diff
@@ -0,0 +1,248 @@
...
+
+        current = it.next();
+    }
+
+    if result.is_some() {
+        let mut result = result.unwrap();
+        result.sort_unstable_by_key(|p| p.ts);
+        build!{
+            Timevector {
+                series: SeriesType::ExplicitSeries{
```
JLockerman, line 118:
```suggestion
                series: SeriesType::SortedSeries{
```

extension/src/time_series/pipeline/fill_to.rs
```diff
@@ -0,0 +1,248 @@
...
+                target += interval;
+            }
+        }
+
+        current = it.next();
+    }
+
+    if result.is_some() {
+        let mut result = result.unwrap();
+        result.sort_unstable_by_key(|p| p.ts);
```
JLockerman, line 115: 1. do we need this? is this guaranteed to be sorted by the fact that we have sorted input?
2. Unless we deduplicate, I think we should use a stable sort, that we if there's meaning to the current order we'll preserve it
```suggestion
        result.sort_by_key(|p| p.ts);
```

JLockerman: > is this guaranteed to be sorted...
Oh, I see: we push the new elements onto the end of the vector

extension/src/time_series/pipeline/fill_to.rs
```diff
@@ -0,0 +1,246 @@
...
+            let result = result.as_mut().unwrap();
+            let mut target = lhs.ts + interval;
+            while target < rhs.ts {
+                result.push(method.fill_point(&lhs, &rhs, target));
+                target += interval;
+            }
+        }
+
+        current = it.next();
+    }
```
JLockerman, line 109: Just for fun, if we were only working on slices we could do
```suggestion
    let mut result = vec![];
    let mut windows = series.points.windows(2);

    for window in windows {
        if window[1].ts - window[0].ts > interval {
            let result = result.as_mut().unwrap();
            let mut target = window[0].ts + interval;
            while target < window[1].ts {
                result.push(method.fill_point(&window[0], &window[1], target));
                target += interval;
            }
        }
    }
```

extension/src/time_series/pipeline/fill_to.rs
WireBaron: added above

extension/src/time_series/pipeline/fill_to.rs
WireBaron: nice, knew there had to be a better way to do this.

extension/src/time_series/pipeline/fill_to.rs
WireBaron: much nicer!

extension/src/time_series/pipeline/fill_to.rs
WireBaron: Ooops, that's an important fix

extension/src/time_series/pipeline/fill_to.rs
WireBaron: fixed

extension/src/time_series/pipeline/fill_to.rs
```diff
@@ -0,0 +1,241 @@
...
+        match it.peek() {
+            Some(rhs) if rhs.ts - lhs.ts > interval => {
+                let mut target = lhs.ts + interval;
+                while target < rhs.ts {
+                    result.push(method.fill_point(&lhs, &rhs, target));
+                    target += interval;
+                }
+            },
+            _ => {},
+        }
```
JLockerman, line 101:
```suggestion
    while let (Some(lhs), Some(rhs) = (current, it.peek()) {
        if rhs.ts - lhs.ts > interval => {
            let mut target = lhs.ts + interval;
            while target < rhs.ts {
                result.push(method.fill_point(&lhs, &rhs, target));
                target += interval;
            }
        }
```

extension/src/time_series/pipeline.rs
```diff
@@ -176,6 +186,8 @@ pub fn execute_pipeline_element<'s, 'e>(
             return filter::apply_lambda_to(timevector, lambda),
         Element::Arithmetic{ function, rhs } =>
             return arithmetic::apply(timevector, *function, *rhs),
+        Element::FillTo{..} =>
+            return fill_to(&timevector, &element),
```
JLockerman, line 190:
```suggestion
            return fill_to(timevector, &element),
```

extension/src/time_series/pipeline/fill_to.rs
```diff
@@ -0,0 +1,241 @@
...
+
+        Element::FillTo {
+            interval,
+            fill_method
+        }.flatten()
+    }
+}
+
+pub fn fill_to<'s>(
+    series: &toolkit_experimental::Timevector<'s>,
```
JLockerman, line 73:
```suggestion
    series: toolkit_experimental::Timevector<'s>,
```

extension/src/time_series/pipeline/fill_to.rs
```diff
@@ -0,0 +1,241 @@
...
+            Timevector {
+                series: SeriesType::SortedSeries{
+                    num_points: result.len() as _,
+                    points: result.into(),
+                }
+            }
+        }
+    } else {
+        series.clone()
+    }
```
JLockerman, line 119:
```suggestion
    
    if result.is_empty() {
        return series
    }
            let mut result: Vec<TSPoint> = series.iter().chain(result.into_iter()).collect();
        result.sort_by_key(|p| p.ts);
        build!{
            Timevector {
                series: SeriesType::SortedSeries{
                    num_points: result.len() as _,
                    points: result.into(),
                }
            }
        }

```

PR #282, opened 2021-10-20
Update comment in control file

svenklemm:
The comment in the control file is shown as the extension description
inside postgres, so this patch replaces it with a longer description.


JLockerman:
bors merge

Issue #177, opened 2021-07-13
[tracking issue] Switch to upstream PGX


eeeebbbbrrrr:
Good work and congrats on the release!  🎉

I noticed that y'all are using a fork of `pgx` that @JLockerman is maintaining.  

I looked at the commits and they mostly look reasonable.  We'd appreciate some PRs so we can get your changes merged upstream.  Then y'all won't need to ask your users install a version of pgx that's not compatible with upstream.

Thanks and let me know how we can help.

JLockerman:
Thank you!

We would very much like to be on stock upstream, and I have plans to open PRs for the remaining stuff we're relying on. I've just been a bit too overwhelmed recently to translating our stuff into things that are actually worth upstreaming 😓

eeeebbbbrrrr:
We're working on a new type resolution system that is replacing cargo-pgx' `schema.rs`.  

Long story short -- it builds a `bin/sql-generator.rs` file based on how things in your extension code are annotated.  And that builds a dependency graph and spits out nicely formatted and commented SQL in the proper order for `CREATE EXTENSION` to succeed.  It's already working against ZDB, which is quite complex.

@Hoverbear is the one doing this work and might drop by and talk a little more about it.

It can even output graphviz .dot files so you can visualize your schema dependencies.

Then we've got plans for auto-generating extension schema upgrade scripts, and then, the cool thing is we're gonna actually parse your `extension_sql!()` and get them to tie into the dependency graph.  `extension_sql!()` can do that now as opaque blocks, which is killer by itself.

I said all this because looking through your fork, I couldn't quite figure out the `varlena_type!` macro. I suspect we'll be breaking that so we need to figure out a proper solution.

But please, send us PRs.  Happy to merge and publish to crates.io ASAP.  I don't want a situation where we've fractured the pgx community and then users get confused by pgx-proper and pgx-tsdb.

Hoverbear:
@JLockerman Oh my gosh another test case for the new code! I'd love to chat with you about this @JLockerman ! Are you on our discord? Feel free to ping me and we can discuss! I wanna make sure the new generator can handle the features you need, too!

JLockerman:
@eeeebbbbrrrr oh that is very interesting!

[...]

PR #295, opened 2021-11-05
Update `pgx` to `0.2`

JLockerman:
This commit updates our pgx version to the `0.2`. This brings us back to using stock pgx, and opens the way for Postgres 14 support.

The big change for the `0.2` series, and the cause of the version-bump along with most of the changes seen in this commit, is a change to how install scripts are generated: whereas they were previously generated by `cargo pgx` reading through all the Rust files looking for pgx's markers, it is now generated by the pgx-macros themselves. The macros that generate the Postgres shims–`#[pg_extern]`, `#[derive(PostgresType)`, etc–now also generate functions that describe the meaning of these objects. This info, along with a catalog of well-known types are used to generate the equivalent SQL when the extension is run as a binary. There are more details to this than explained here, but the important change for our purposes is that while SQL generation for functions used to be entirely nominal–the generator would look at the names of the argument types and put that in the SQL–it is now type-based. This leads us to a number of changes:
  1. Instead of using `type Foo = pg_sys::Datum` when we want raw access to a Datum of a given type, we now have a `raw` module containing slightly more principled raw types.
  2. We need to migrate to the upstream `Internal` type, instead of our existing polymorphic one. In order to maximize type safety, and minimize the amount of churn, we provide a new type `Inner<...>` which serves the same role as our old `Internal<...>` type, and add shims to communicate with the SQL generator. Long-term we should communicate with pgx and look into providing richer internal types, or an aggregate generator that creates the shims on its own.
  3. `varlena_type!()` is obsolete! Since SQL generation is type-based it can see through our macros and generate the correct SQL despite the actual types not being visible in the top-level files.
  4. As part of this we need to move the actual type-creation into the schema-generating modules, we cannot just have the marker there.

Something that this commit does not take advantage of, that we may want to consider in the future, is removing the `toolkit_experimental::` qualifications from arguments and return types of experimental functions. Since type schema-qualification is now based on the type and not the path, it should be possible to do so, and should make the code more readable.

fixes https://github.com/timescale/timescaledb-toolkit/issues/177


davidkohn88 (commented):
LGTM! I'd probably stick to the raw strings for agg definitions, but I think you said something about it looking better in the generated SQL or something...

PR #299, opened 2021-11-15
Start testing on Postgres 14

JLockerman:
With pgx 2.4 we have [all the headers](https://github.com/zombodb/pgx/pull/289) we need to run the Toolkit on pg14! This commit starts running CI tests on it. This isn't _quite_ complete yet since TimescaleDB does not yet support pg14, so we can't run our continuous aggregate tests on it. Once that's done we should be able to just throw a switch and release.

Also removes some now-unneeded external definitions.


JLockerman:
bors retry

PR #306, opened 2021-11-24
Update rust edition and clean up clippy warnings

WireBaron:


crates/t-digest/src/lib.rs
```diff
@@ -579,9 +579,10 @@ impl TDigest {
         let centroid_weight = (rank - t as f64) / self.centroids[pos].weight() as f64;
 
         // Now we use that location to interpolate the desired value between the centroid mean and the weighted midpoint between the next centroid in the direction of the target rank.
-        return if centroid_weight == 0.5 {
+        let diff = centroid_weight - 0.5;
+        return if diff.abs() < f64::EPSILON {
             self.centroids[pos].mean()
-        } else if centroid_weight < 0.5 {
+        } else if diff < 0. {
```
JLockerman, line 585: should be `is_neg()`?

JLockerman (approved):

crates/t-digest/src/lib.rs
WireBaron: looks like it's `is_sign_negative()`...I like your version better

made the change.

PR #310, opened 2021-12-02
Update rust to 1.57

JLockerman:
This allows us to finally use real error messages in the flat_serialize
alignment checks!
Also requires cleaning up some new clippy lints.


WireBaron (approved):
Looks good so far, but still failing clippy

PR #312, opened 2021-12-07
Reduce duplication and add extension load configuration.

epgts:


JLockerman (approved):
🎉 
Looks like it needs a rebase (sorry, I updated the CI image yesterday, it's not normally required 😅) then you can tell bors it's good to go.

Readme.md
```diff
@@ -56,6 +55,12 @@ cargo pgx install --release && \
 cargo run --manifest-path ../tools/post-install/Cargo.toml -- pg_config
 ```
 
+To start PostgreSQL with the extension loaded, add `-c shared_preload_libraries=timescaledb_toolkit`
+to the startup options, or add this to the `postgresql.conf` file:
+```
+shared_preload_libraries = timescaledb_toolkit
+```
+
```
JLockerman, line 63: I think the better advice here is
> After the extension is installed it can be added to a database by running
> ```SQL
> CREATE EXTENSION timescaledb_toolkit;
> ```
> in `psql`

PR #313, opened 2021-12-09
aggregate builder

JLockerman:
This PR adds an aggregate builder to toolkit in order to

1. Prevent the possibility that someone gets the inner transition state type wrong between different functions making up and aggregate.
2. Ensure that the aggregate context is switched to in, and only in, the correct places.
3. Reduce the likelihood of errors in `CREATE AGGREGATE` statements.
4. Focus aggregate reviewers on the important semantic components of the aggregate implementation instead of the boilerplate.
5. Reduce the effort involved in writing aggregates.

The syntax we eventually settled on looks something like the following
```rust
#[aggregate] impl aggregate_name {
    type State = InternalTransitionType;

    fn transition(
        state: Option<Inner<State>>,
        #[sql_type("sql_type")] argument: RustType, // can have an arbitrary number of args
    ) -> Option<Inner<State>> {
        // transition function function body goes here
    }

    fn finally(state: Option<Inner<State>>) -> Option<ResultType> {
        // final function function body goes here
    }

    // the remaining items are optional

    // parallel-safety marker if desireable
    const PARALLEL_SAFE: bool = true;

    fn serialize(state: Inner<State>) -> bytea {
        // serialize function body goes here
    }

    fn deserialize(bytes: bytea) -> Inner<State> {
        // deserialize function body goes here
    }

    fn combine(state1: Option<Inner<State>>, state2: Option<Inner<State>>) -> Option<Inner<State>> {
        // combine function body goes here
    }
}
```
for example, the simplest aggregate, `arbitrary()`, that returns an element from the aggregated values, looks like
```rust
#[aggregate] impl arbitrary {
    type State = String;

    fn transition(
        state: Option<Inner<State>>,
        #[sql_type("text")] value: String,
    ) -> Option<Inner<State>> {
        match state {
            Some(value) => Some(value),
            None => Some(value.into()),
        }
    }

    fn finally(mut state: Option<Inner<State>>) -> Option<String> {
        match &mut state {
            None => None,
            Some(state) => Some(state.clone()),
        }
    }
}
```

This syntax has a couple of advantages compared to our earlier attemtps: it both looks reasonably familiar to rust users as well as looking like `CREATE AGGREGATE`; the semantics of items created by the macro largely behave as a rust user would expect them to; it has syntactic space for the aggregate's name; it should be easily extensible to moving-aggregates, notably the moving aggregate transition state should fit in naturally as something like `type MovingState`.

----

Related to https://github.com/zombodb/pgx/issues/194, though taking a slightly different strategy. Depending on where that ends up going, this will either be obviated by that, be upstreamed into that effort, or exist alongside it.

### DIFF
--- a/crates/aggregate_builder/Readme.md
+++ b/crates/aggregate_builder/Readme.md
@@ -1,0 +1,82 @@
+# Aggregate Builder #
+
+Library for building Postgres [aggregate functions](https://www.postgresql.org/docs/current/xaggr.html)
+that imitates
+[`CREATE AGGREGATE`](https://www.postgresql.org/docs/current/sql-createaggregate.html).
+
+## Syntax ##
+
+Current syntax looks something like like
+
+```rust
+#[aggregate] impl aggregate_name {
+    type State = InternalTransitionType;
+
+    fn transition(
+        state: Option<Inner<State>>,
+        #[sql_type("sql_type")] argument: RustType, // can have an arbitrary number of args
+    ) -> Option<Inner<State>> {
+        // transition function function body goes here
+    }
+
+    fn finally(state: Option<Inner<State>>) -> Option<ResultType> {
+        // final function function body goes here
+    }
+
+    // the remaining items are optional
+
+    // parallel-safety marker if desireable
+    const PARALLEL_SAFE: bool = true;
+
+    fn serialize(state: Inner<State>) -> bytea {
+        // serialize function body goes here
+    }
+
+    fn deserialize(bytes: bytea) -> Inner<State> {
+        // deserialize function body goes here
+    }
+
+    fn combine(state1: Option<Inner<State>>, state2: Option<Inner<State>>) -> Option<Inner<State>> {
+        // combine function body goes here
+    }
+}
+```
+
+All items except for `type State`, `fn transition()`, and `fn finally()` are
+optional. The SQL for the aggregate and its functions will be created
+automatically, and any necessary memory context switching is handled
+automatically for most cases¹.
+
+¹It will switch to the aggregate memory context before calling the transition
+function body and the combine function body. Looking through `array_agg()`'s
+code this seems to be the correct places to do so. Note that if you want to
+allocate in the aggregate memory context in the final function other work may
+be needed.
+
+## Example ##
+
+Below is a complete example of an `arbitrary()` aggregate that returns one of
+the aggregated values.
+
+```rust
+#[aggregate] impl arbitrary {
+    type State = String;
+
+    fn transition(
+        state: Option<Inner<State>>,
+        #[sql_type("text")] value: String,
+    ) -> Option<Inner<State>> {
+        match state {
+            Some(value) => Some(value),
+            None => Some(value.into()),
+        }
+    }
+
+    fn finally(mut state: Option<Inner<State>>) -> Option<String> {
+        match &mut state {
+            None => None,
+            Some(state) => Some(state.clone()),
+        }
+    }
+}
+```
--- a/crates/aggregate_builder/Readme.md
+++ b/crates/aggregate_builder/Readme.md
@@ -13,9 +13,9 @@ Current syntax looks something like like
     type State = InternalTransitionType;
 
     fn transition(
-        state: Option<Inner<State>>,
+        state: Option<State>,
         #[sql_type("sql_type")] argument: RustType, // can have an arbitrary number of args
-    ) -> Option<Inner<State>> {
+    ) -> Option<State> {
         // transition function function body goes here
     }
 
@@ -32,11 +32,11 @@ Current syntax looks something like like
         // serialize function body goes here
     }
 
-    fn deserialize(bytes: bytea) -> Inner<State> {
+    fn deserialize(bytes: bytea) -> State {
         // deserialize function body goes here
     }
 
-    fn combine(state1: Option<&State>, state2: Option<&State>) -> Option<Inner<State>> {
+    fn combine(state1: Option<&State>, state2: Option<&State>) -> Option<State> {
         // combine function body goes here
     }
 }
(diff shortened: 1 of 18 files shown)


JLockerman:
(soliciting reviews from @WireBaron for code-quality, and @epgts for the DSL's readability)

extension/src/arbitrary.rs
```diff
@@ -0,0 +1,53 @@
+
+use aggregate_builder::aggregate;
+
+use pgx::*;
+
+use crate::{
+    palloc::{Internal, InternalAsValue, Inner, ToInternal},
```
epgts, line 7: The only thing I wonder is whether we can make these macros stand alone. It's very surprising that in order to use `#[aggregate]` I have to add a bunch of `use` that I'm not using, and presumably could grow or shrink with time!

I looked into it a bit and it seems at least some of it comes directly from pgx, but maybe that's something they'd accept?

JLockerman: I have mixed feelings. On the one hand, I'd like this to work as much like an `impl` as possible, on the other hand I definitely agree that things that aren't visible shouldn't require `use`s. I think I'll see if I can have those use absolute paths...

epgts: It looks pretty involved, anyway. I don't see any way to easily use these macros without leading with `use pgx::*` and a smattering of others.

We do have a couple stylistic issues we could address though:

- the outer layer of braces is unneeded:  `use crate::palloc::{Inner, Internal, InternalAsValue, ToInternal};`
- the two matches couple be replaced with `state.map(|s| s.into()).or(value)` and `state.map(|s| s.into())`, respectively. Yeah, this is only sample code, but sample code is expected to be copied and pasted, so arguably ought to be even nittier-pickier! :). Also clippy caught one of them and it would be nice to be clippy-clean.

JLockerman: Removing the `use pgx::*` isn't going to be doable anytime soon; IIRC that requires some work in the expansion of `pg_extern`. The other ones however are unneeded as of the most recent commit! This allows things such as using `type AggregatePointer = Option<crate::palloc::Inner<AggState>>` and using `AggregatePointer` as the transition type, not that I'd recommend doing that.

> the two matches could be replaced with ... Also clippy caught one...
Ah, clippy. I actually disagree with that style; I think the explicit match is more readable, doesn't take much more space, and performs better (well, at compile time), without taking much more space; but the reason we added clippy to CI is to prevent style fights, so what it says goes. (In general this PR isn't clippy-clean yet, so I'd expect there to be a lot of such nits)

Hoverbear:
@JLockerman We should chat and unify efforts!

extension/src/aggregate_builder_tests.rs
```diff
@@ -0,0 +1,219 @@
...
+
+    const PARALLEL_SAFE: bool = true;
+
+    fn serialize(state: Inner<State>) -> bytea {
+        crate::do_serialize!(state)
+    }
+
+    fn deserialize(bytes: bytea) -> Inner<State> {
+        crate::do_deserialize!(bytes, State)
+    }
```
WireBaron, line 89: Doesn't this need a combine to be truly parallel?  Should marking an aggregate `PARALLEL_SAFE` without providing the required support functions generate a compile time error?

extension/src/tdigest.rs
```diff
@@ -634,22 +569,21 @@ mod tests {
     #[pg_test]
     fn test_tdigest_byte_io() {
         unsafe {
-            use std::ptr;
-            let state = tdigest_trans_inner(None, 100, Some(14.0), ptr::null_mut());
-            let state = tdigest_trans_inner(state, 100, Some(18.0), ptr::null_mut());
-            let state = tdigest_trans_inner(state, 100, Some(22.7), ptr::null_mut());
-            let state = tdigest_trans_inner(state, 100, Some(39.42), ptr::null_mut());
-            let state = tdigest_trans_inner(state, 100, Some(-43.0), ptr::null_mut());
+            let state = super::tdigest::transition(None,  100, Some(14.0));
```
WireBaron, line 572: Should we be using `super::*`...I almost do this habitually in tests now.

crates/aggregate_builder/src/lib.rs
```diff
@@ -0,0 +1,749 @@
...
+    match idx {
+        None => None,
+        Some((idx, _)) => {
+            let attribute = attrs.remove(idx);
+            Some(attribute)
+        }
+    }
+}
+
+//
```
WireBaron, line 343: Bit of cleanup needed here.

WireBaron (approved):
Looks good overall.

extension/src/aggregate_builder_tests.rs
JLockerman: 🤔 y'know, I've never tried that. I wonder what postgres will do...

JLockerman: Apparently that is valid as far as `CREATE AGGREGATE` is concerned, but it won't be parallelizable. Since I currently also don't have checks that an aggregate has either both or neither of `serialize` and `deserialize` I might leave this to a future PR...

extension/src/tdigest.rs
JLockerman: I generally consider it better style to use more precise imports, though I admit that tests can be a grey area.

crates/aggregate_builder/src/lib.rs
JLockerman: That's entirely deliberate: it's there to serve as a visual marker when scrolling looking for the expansion part of the proc-macro. I'll turn it into a real comment.

extension/src/aggregate_builder_tests.rs
JLockerman: Well, implementing the check found a bug (I forgot to implement `fn combine()` for the test aggregates) so I think this is worth it.

JLockerman:
re-requesting Brian's review for the new changes

crates/aggregate_builder/Readme.md
```diff
@@ -28,15 +28,15 @@ Current syntax looks something like like
...
+    fn serialize(state: &State) -> bytea {
         // serialize function body goes here
     }
 
     fn deserialize(bytes: bytea) -> Inner<State> {
         // deserialize function body goes here
     }
 
-    fn combine(state1: Option<Inner<State>>, state2: Option<Inner<State>>) -> Option<Inner<State>> {
+    fn combine(state1: Option<&State>, state2: Option<&State>) -> Option<Inner<State>> {
```
WireBaron, line 39: Given that this return is the only surviving `Inner`, I'd try to wrap this to hide that layer entirely.

WireBaron (approved):
One last suggestion

crates/aggregate_builder/Readme.md
JLockerman: 🤔 I didn't believe I could last time I checked, but maybe it'll work now...

JLockerman: ah, this is `fn combine()` not `fn transition()` I think I actually can hide it here. That said, we're also using it in `fn combine()` and `fn transition()` and I'm not sure if we should do one without the others.

---

The issue with `fn transition()`: in _theory_ it would be nice to give `fn transition()` a signature like
```rust
fn transition(state: Option<State>) -> Option<State>
```
unfortunately, this runs into weird issues around `panic!()`s: desugared, the call procedure for the inner function would look something like
```rust
fn transition_like(state: Option<Inner<State>>) -> Option<Inner<State>> {
    match state {
        None => inner(None).map(|s| s.into()),
        Some(mut state) => {
            let output = inner(Some(*state));
            match output {
                None => None,
                Some(new) => {
                    *state = new;
                    Some(state)
                }
            }
        }
    }
}

fn inner(state: Option<String>) -> Option<String> { /* elided */ }
```
now this _almost_ works, and the equivalent code is safe for box ([example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=82e3747cd9f2c89cdbd289481332f28f)) the issue we run into is that we want `Inner<State>` to be recursively freed when the aggregate memory context is discarded. If this occurs due to an error in `inner(Some(state))` we need some way to communicate to `Inner`'s `Drop` impl that `state` has been moved out and it can only drop the pointer, not its contents. Right now I only know of one way to do this: store a dynamic drop flag. That would make the transition function desugar to something more like:

```rust
fn transition_like(state: Option<Inner<Option<String>>>) -> Option<Inner<Option<String>>> {
    match state {
        None => inner(None).map(|s| s.into()),
        Some(mut state) => {
            let inner_state = state.take();
            let output = inner(Some(inner_state));
            match output {
                None => None,
                Some(new) => {
                    *state = Some(new);
                    Some(state)
                }
            }
        }
    }
}
```
Is it worth storing the extra flag so we don't need to think about `Inner<>` as much? I'm not really sure. Realistically, I don't think the performance would change enough to matter, and the complexity isn't _too_ bad, but then, I don't see the complexity being saved as all that much either...

JLockerman: Note that there _may_ be a future compatibility hazard with window functions; they may require a special input to get the tuplestore out of. I don't think this really matters here though.

PR #316, opened 2021-12-15
Remove unused symbols in spacesaving::tests .

epgts:


WireBaron (approved):
Looks good, I should make another pass through these tests to see if there's any more messes I should have cleaned up the last time I modified them.

PR #320, opened 2021-12-21
Adding frequency aggregate

WireBaron:
This change adds a aggregate found to use the most
commonly occuring element for any column.


extension/src/datum_utils.rs
```diff
@@ -198,55 +198,98 @@ flat_serialize_macro::flat_serialize! {
...
+            where
+                A: SeqAccess<'de>,
+            {
+                // borrowed from serialization/types.rs
+                unsafe fn get_struct<T>(tuple: pg_sys::HeapTuple) -> *mut T {
+                    //((char *) ((TUP)->t_data) + (TUP)->t_data->t_hoff)
+                    let t_data: *mut u8 = (*tuple).t_data.cast();
+                    let t_hoff = (*(*tuple).t_data).t_hoff;
+                    t_data.add(t_hoff as usize).cast()
+                }
```
JLockerman, line 253: I think we should only have one `fn get_struct()`, probably in `serialization`

extension/src/datum_utils.rs
```diff
@@ -198,55 +198,98 @@ flat_serialize_macro::flat_serialize! {
...
+                    let tuple =
+                        pg_sys::SearchSysCache1(pg_sys::SysCacheIdentifier_TYPEOID as _, oid as _);
+                    if tuple.is_null() {
+                        pgx::error!("no type info for oid {}", oid);
+                    }
+        
+                    let type_tuple: pg_sys::Form_pg_type = get_struct(tuple);
+                    typmod = (*type_tuple).typmodin;
+                    pg_sys::ReleaseSysCache(tuple);
+                }
```
JLockerman, line 277: You shouldn't need typmod, IIRC it's only used for choosing between different output formats, is largely ignored, and `-1` is always a valid default.

extension/src/datum_utils.rs
```diff
@@ -198,55 +198,98 @@ flat_serialize_macro::flat_serialize! {
...
+        }
+        let count = self.iter().count();
+        let mut seq = serializer.serialize_seq(Some(count + 1))?;
+        seq.serialize_element(&self.type_oid.0)?;
+        for element in self.iter() {
+            let chars = unsafe{pg_sys::OutputFunctionCall(&mut flinfo, element)};
+            let cstr = unsafe {
+                std::ffi::CStr::from_ptr(chars)
+            };
+            seq.serialize_element(cstr.to_str().unwrap())?;
```
JLockerman, line 223: We should add a
```
// TODO currently this always serializes inner data as text. When we start working on more-efficient network serialization, or we start using this in a transition state, we should use the binary format if we don't need human-readable output.
```

extension/src/datum_utils.rs
```diff
@@ -0,0 +1,619 @@
...
+                        pgx::error!("no type info for oid {}", oid);
+                    }
+        
+                    let type_tuple: pg_sys::Form_pg_type = get_struct(tuple);
+                    typmod = (*type_tuple).typmodin;
+                    pg_sys::ReleaseSysCache(tuple);
+                }
+
+                let mut data = vec![];
+                while let Some(next) = seq.next_element::<String>()? {
```
JLockerman, line 280:
```suggestion
                while let Some(next) = seq.next_element::<&str>()? {
```
or 
```suggestion
                while let Some(next) = seq.next_element::<&'de str>()? {
```

extension/src/datum_utils.rs
```diff
@@ -198,55 +198,98 @@ flat_serialize_macro::flat_serialize! {
...
+                    let next = std::ffi::CString::new(next).unwrap();  // TODO: error handling
```
JLockerman, line 281: annoying that pg forces an extra allocation here, but them's the breaks

extension/src/datum_utils.rs
WireBaron: Looks good, taking verbatim

extension/src/datum_utils.rs
WireBaron: Agreed, I'll make the serialization one crate public.

extension/src/datum_utils.rs
WireBaron: Good to know, this was something of a pain to get.

extension/src/datum_utils.rs
WireBaron: Actually this is no longer needed now that I'm not getting typmod, so I'm not bothering making get_struct available.

extension/src/datum_utils.rs
WireBaron: Yeah, I was annoyed with this as well.

extension/src/datum_utils.rs
WireBaron: Went with `&str`

PR #324, opened 2022-01-10
Prototype `state_agg` and `duration_in`.

epgts:
I think I've addressed everything we discussed previously. A question or two remains and I have a new question: how do I test `combine` and parallelism?

Thanks!

### DIFF
--- a/extension/src/state_aggregate.rs
+++ b/extension/src/state_aggregate.rs
@@ -1,0 +1,624 @@
+//! SELECT duration_in('STOPPED', states) as run_time, duration_in('ERROR', states) as error_time FROM (
+//!   SELECT state_agg(time, state) as states FROM ...
+//! );
+//!
+//! Currently requires loading all data into memory in order to sort it by time.
+
+#![allow(non_camel_case_types)]
+
+use pgx::*;
+use serde::{Deserialize, Serialize};
+
+use aggregate_builder::aggregate;
+use flat_serialize::*;
+use flat_serialize_macro::FlatSerializable;
+
+use crate::{
+    flatten,
+    palloc::{Inner, Internal},
+    pg_type,
+    raw::{bytea, TimestampTz},
+    ron_inout_funcs,
+};
+
+use toolkit_experimental::StateAgg;
+
+#[pg_schema]
+pub mod toolkit_experimental {
+    use super::*;
+
+    pg_type! {
+        #[derive(Debug)]
+        struct StateAgg<'input> {
+            states_len: u64, // TODO JOSH this and durations_len can be 32
+            durations_len: u64,
+            durations: [DurationInState; self.durations_len],
+            states: [u8; self.states_len],
+        }
+    }
+
+    impl StateAgg<'_> {
+        pub fn new(states: String, durations: Vec<DurationInState>) -> Self {
+            let states_len = states.len() as u64;
+            let durations_len = durations.len() as u64;
+            unsafe {
+                flatten!(StateAgg {
+                    states_len,
+                    states: states.into_bytes().into(),
+                    durations_len,
+                    durations: (&*durations).into(),
+                })
+            }
+        }
+
+        pub fn get(&self, state: &str) -> Option<i64> {
+            for record in self.durations.iter() {
+                if self.state_str(&record) == state {
+                    return Some(record.duration);
+                }
+            }
+            None
+        }
+
+        fn state_str(&self, record: &DurationInState) -> &str {
+            let beg = record.state_beg as usize;
+            let end = record.state_end as usize;
+            let state = &self.states.as_slice()[beg..end];
+            // SAFETY: came from a String in `new` a few lines up
+            unsafe { std::str::from_utf8_unchecked(state) }
+        }
+    }
+
+    ron_inout_funcs!(StateAgg);
+}
+
+#[aggregate]
+impl toolkit_experimental::state_agg {
+    type State = StateAggTransState;
+
+    const PARALLEL_SAFE: bool = true;
+
+    fn transition(
+        state: Option<State>,
+        #[sql_type("timestamptz")] ts: TimestampTz,
+        #[sql_type("text")] value: Option<String>,
+    ) -> Option<State> {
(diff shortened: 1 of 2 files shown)


crates/state-aggregate/src/lib.rs
```diff
@@ -0,0 +1,85 @@
...
+    last_state: Option<(String, i64)>,
+    // TODO using BTreeMap so lazy prototype tests can depend on stable iteration order
+    durations: std::collections::BTreeMap<String, i64>,
+}
+impl DurationState {
+    fn handle_record(&mut self, state: String, raw_time: i64) {
+        if let Some((last_state, last_time)) = self.last_state.take() {
+            if last_state == state {
+                // Same state recorded twice:  preserve the older timestamp.
+                self.last_state = Some((state, last_time));
```
WireBaron, line 57: What if this is the last record?  It looks like if the last N entries have the same state, that interval won't be reflected in the result of `to_duration_map`

crates/state-aggregate/src/lib.rs
```diff
@@ -0,0 +1,85 @@
+use serde::{Deserialize, Serialize};
+
+use flat_serialize_macro::FlatSerializable;
+
+// Intermediate state kept in postgres.
```
WireBaron, line 5: The `/crates` area is more intended to store postgres agnostic algorithm implementations.  The trans state, and the rest of the objects here should go in `extension\src\state_aggregate.rs` as it isn't really useful outside that context.

extension/src/state_aggregate.rs
```diff
@@ -0,0 +1,342 @@
...
+        state: Option<State>,
+        #[sql_type("timestamptz")] ts: TimestampTz,
+        #[sql_type("text")] value: Option<String>,
+    ) -> Option<State> {
+        let value = match value {
+            None => return state,
+            Some(value) => value,
+        };
+        let mut state = match state {
+            None => state_aggregate::TransState::new().into(),
```
WireBaron, line 39: I don't think this `into` is needed.  I believe clippy will complain about this (check with `cargo clippy --features 'pg13 pg_test'`).

extension/src/state_aggregate.rs
```diff
@@ -0,0 +1,342 @@
...
+pub fn duration_in(state: String, aggregate: Option<StateAgg>) -> i64 {
+    aggregate
+        .map(|aggregate| aggregate.get(&state))
+        .flatten()
+        .unwrap_or(0)
+}
+
+pg_type! {
+    #[derive(Debug)]
+    struct StateAgg<'input> {
```
WireBaron, line 101: Ultimately you'll likely need to also store the start and end records for this range in order to implement `rollup` (combining multiple completed StateAggs).

WireBaron (changes requested):

crates/state-aggregate/src/lib.rs
epgts: I added a test for this, then fixed it, and then realized the bug was only caused by unnecessary complexity here anyway. I simply deleted the whole part about checking for two states in a row and all tests pass including the new one covering this case.

extension/src/state_aggregate.rs
```diff
@@ -0,0 +1,464 @@
...
+//! );
+//!
+//! Currently requires loading all data into memory in order to sort it by time.
+
+// TODO How to mark everything experimental?  I can't seem to get these 3:
+// unexpectedly released features: [
+//     "function stateagg_in(cstring)",
+//     "function stateagg_out(stateagg)",
+//     "type stateagg",
+// ]
```
JLockerman, line 12: for these you put the type definition and `ron_inout_funcs!` in a `#[pg_schema]` module [like so](https://github.com/timescale/timescaledb-toolkit/blob/1c98deac2121402962e6284b9d5534c780d15c38/extension/src/topn.rs#L37-L54)

extension/src/state_aggregate.rs
```diff
@@ -0,0 +1,618 @@
...
+            match (a, b) {
+                (None, None) => COMBINE_NONE.fetch_add(1, Relaxed),
+                (Some(_), None) => COMBINE_A.fetch_add(1, Relaxed),
+                (None, Some(_)) => COMBINE_B.fetch_add(1, Relaxed),
+                (Some(_), Some(_)) => COMBINE_BOTH.fetch_add(1, Relaxed),
+            };
+        }
+        match (a, b) {
+            (None, None) => None,                     // TODO untested
+            (Some(only), None) => Some(only.clone()), // TODO untested
```
epgts, line 109: It would be nice to be able to reach these arms, but I don't think it needs to block merge.

extension/src/state_aggregate.rs
```diff
@@ -0,0 +1,618 @@
...
+        Spi::execute(|client| {
+            client.select("CREATE TABLE test(ts timestamptz, state TEXT)", None, None);
+            client.select(
+                r#"INSERT INTO test VALUES
+                        ('2020-01-01 00:00:00+00', 'START'),
+                        ('2020-01-01 00:00:00+00', 'ERROR')"#,
+                None,
+                None,
+            );
+            // TODO why isn't should_panic working?
```
epgts, line 598: This doesn't need to block merge either; I can just remove the test for now.

If we expect to report errors to the user, we ought to test them, so I don't want to lose this forever.

extension/src/state_aggregate.rs
```diff
@@ -0,0 +1,618 @@
...
+        .flatten()
+        .unwrap_or(0)
+}
+
+#[derive(Clone, Debug, Deserialize, Eq, FlatSerializable, PartialEq, Serialize)]
+#[repr(C)]
+pub struct DurationInState {
+    duration: i64,
+    state_beg: u64,
+    state_end: u64,
```
JLockerman, line 203: if you want these can be u32s; pg datatypes can't be larger than ~2GB anyway

epgts:
Code review TODOs are here for the next phase of revisions post-merge, and the if lets are converted to matches.

Should be ready to go!

Thanks!

PR #325, opened 2022-01-12
Improve error message for counter_agg

JLockerman:
We should have a more expressive error message than the default unwrap one.

### DIFF
--- a/extension/src/counter_agg.rs
+++ b/extension/src/counter_agg.rs
@@ -118,7 +118,7 @@ impl CounterSummaryTransState {
         let mut iter = self.point_buffer.iter();
         let mut summary = InternalCounterSummary::new( iter.next().unwrap(), self.bounds);
         for p in iter {
-            summary.add_point(p).unwrap();
+            summary.add_point(p).unwrap_or_else(|e| pgx::error!("{}", e));
         }
         self.point_buffer.clear();
         // check bounds only after we've combined all the points, so we aren't doing it all the time.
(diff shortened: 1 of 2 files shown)


extension/src/counter_agg.rs
```diff
@@ -101,6 +101,15 @@ pub struct CounterSummaryTransState {
     summary_buffer: Vec<InternalCounterSummary>,
 }
 
+macro_rules! unwrap {
+    ($e:expr) => {
+        match $e {
+            Ok(val) => val,
+            Err(e) => pgx::error!("{}", e),
```
epgts, line 108: pgx::error looks like a panic wrapper of dubious value; I think it's clearer to use panic directly.

Also, what do you think about using something a little more close to normal Rust error handling? Something like this:

```
fn abort<T, E: std::fmt::Display>(e: E) -> Result<T, E> {
    panic!("{}", e)
}
summary.add_point(p).or_else(abort).unwrap();
```

It looks like we only need it in a few spots so it won't be too noisy, but if it does start to look too noisy, those methods can be changed to return Result of CounterError, too, and the function one level out can use abort.

davidkohn88 (commented):
I think @epgts comments are good. Otherwise LGTM.

extension/src/counter_agg.rs
JLockerman: > pgx::error looks like a panic wrapper of dubious value; I think it's clearer to use panic directly.

yeah, kinda is now. I still prefer using it b/c postgres errors are a bit richer than rust `panic!()` (they can have details and hints in addition to the message) and I have hope that we'll be able to add them to the `error!()` macro. Using it now will hopefully make it easier to grep for them once we make the needed changes upstream.


Honestly, what's really embarrassing is that I forgot about `unwrap_or_else()` the following should work
```rust
summary.add_point(p).unwrap_or_else(|| pgx::error!("{}", e))
```
in fact, since we don't need the return value other wise, just `or_else()` should work
```rust
summary.add_point(p).or_else(|| pgx::error!("{}", e))
```

JLockerman: type-inference issues in the plain `or_else()` version. I suppose there's an argument to be made that `unwrap_or_else()` expresses intent better anyway 🤔

PR #332, opened 2022-01-27
Speed up builds by fixing github action cache and cargo build cache

epgts:
* Changes

** Add comments throughout ci.yml explaining how it works.

** Correct path in CARGO_HOME cache.

We were attempting to cache two non-existent directories
`/home/postgres/.cargo` (`~/.cargo`).  Our actual CARGO_HOME is
`/usr/local/cargo`, so instead we'll cache those.

** Comment out restore-keys and leave TODO for the plan.

It's not clear why we had so many.  The github action cache is write
ONCE, which means over time it becomes colder and colder.  We need to
increment the cache keys periodically to warm it back up.  When we do
that, we can avoid dropping all the way down to cold cache performance
by moving the previous cacche keys to restore-keys, but it shouldn't
make sense for restore-keys to list more than one previous key.

** Move all commands to new tools/build script.

This both reduces duplication across all the blocks in ci.yml as well as
allowing us to build and test the same way locally.

** Replace sudo with complex and unnecessary flags with su.

Before:
sudo -HEsu postgres sh -c "..."
- -E - preserve environment - su does this already
- -H - force HOME to target user's home directory - su does this already
- -s - run command through shell rather than directly.  And then we ran
  ANOTHER shell under that.  Both were unnecessary.
- -u postgres - have to specify the user

After:
su postgres -c ...'

** Specify PGVERSION in each job's environment to reduce duplication.

** Run crates tests only once and separate from pgx tests.

These tests do not depend on postgresql so running them more than once
buys us nothing.

Worse, not running the pgx tests under the extension subdirectory causes
pgx to thrash cargo's build cache and build the world twice.  We don't
fully understand this, but document what we do know in tools/build.

** TODO Add questions about many of the environment variables.

** TODO Consider running post-install / doc tests separately.

* Timings

Cold cache: 18m20s
https://github.com/timescale/timescaledb-toolkit/runs/5007577229?check_suite_focus=true
Warm cache:  9m14s
https://github.com/timescale/timescaledb-toolkit/runs/5007862209?check_suite_focus=true

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,10 +18,22 @@ jobs:
     container:
       image: timescaledev/rust-pgx:latest
       env:
+        # Must keep in sync with `path: target` in cache entry below
+        CARGO_TARGET_DIR_NAME: target
+        # TODO Why?  Cargo default is to pass `-C incremental` to rustc; why don't we want that?
+        #   https://doc.rust-lang.org/rustc/codegen-options/index.html#incremental
+        #   Well turning it off takes the extension target size down from 3G to 2G...
         CARGO_INCREMENTAL: 0
+        # TODO Why?  If we're concerned about trouble fetching crates, why not
+        #  just fetch them once at the time we select a dependency?
+        #  Errors fetching crates are probably rare enough that we don't see the
+        #  need to bother, but then why not just let the build fail?
         CARGO_NET_RETRY: 10
+        # TODO What reads this?  It's not listed on
+        #  https://doc.rust-lang.org/cargo/reference/environment-variables.html
         CI: 1
         RUST_BACKTRACE: short
+        # TODO We don't seem to run rustup, nor does it seem like we should.
         RUSTUP_MAX_RETRIES: 10
 
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -30,41 +42,50 @@ jobs:
       with:
         ref: ${{ github.event.pull_request.head.sha }}
 
+    # ~postgres/.pgx/config.toml comes in on the image.
+    # Setting HOME=~postgres (which is /home/postgres) in cargo's environment
+    # would work, but sticking with this approach for now.
```
JLockerman, line 47: This does seem more elegant. I _think_ I ran into permissions issues when I tried to just run in the existing directory without changing ownership 🤔

.github/workflows/ci.yml
```diff
@@ -79,50 +100,42 @@ jobs:
...
+        key: TESTING6-${{ runner.os }}-test-pg${{ env.PGVERSION }}-target
 
-    - name: Run PG 12 Tests
-      run: sudo -HEsu postgres sh -c "/usr/local/cargo/bin/cargo test --workspace --features 'pg12 pg_test'"
+    - name: Run pgx tests
+      run: su postgres -c 'sh tools/build -pg$PGVERSION test-extension 2>&1'
 
+    # TODO post-install + doc test adds a good 90s.  We could probably put
+    # them together run serially in a separate job to shave that 90s off the
+    # long pole.
```
JLockerman, line 118: `post-install` is testing update scripts, which should be safe to just do once. doc tests are mostly testing continuous aggregates, which I _think_ we want to do on every version 🤔

tools/build
```diff
@@ -0,0 +1,128 @@
...
+
+# For reasons we don't yet understand, pgx prevents the cargo cache from
+# working across runs UNLESS all pgx commands are run from the 'extension'
+# subdirectory of the project.  Any cargo command we need to run from that
+# subdirectory, we use CARGO_TARGET_DIR="$extension_CARGO_TARGET_DIR", and for
+# all others, CARGO_TARGET_DIR="$top_CARGO_TARGET_DIR".
+#
+# If each entry in ci.yml `jobs` is guaranteed to be run in isolation, always
+# with the target dir freshly copied from the specified cache key, then we
+# don't need to bother with CARGO_TARGET_DIR here.
```
JLockerman, line 48: I _think_ this holds. If I understand correctly, each job [will run in its own VM](https://docs.github.com/en/actions/learn-github-actions/understanding-github-actions#the-components-of-github-actions) which is isolated from the VMs running the other jobs. Don't know if that will simplify anything.

tools/build
```diff
@@ -0,0 +1,128 @@
...
+#
+# However, this script is derived from one I was using for my local builds
+# anyway, and apparently everyone is suffering the slow build times locally,
+# so why have more than one way to build?  Why not use the same tool locally
+# as the official build uses?
+#
+# CARGO_TARGET_DIR_NAME allows me to call my target directory '.target'.
+# TODO Can we simplify and just force .target here?  It works fine, but I didn't want to surprise people...
```
JLockerman, line 56: do you mean `./target`? If so, I think that should be safe to use. If we really wanted to be principled, I guess we could try to extract `target_directory` it from `cargo metadata` 🤔

JLockerman: (users _can_ override the default target directory using a [`.cargo/config.toml`](https://doc.rust-lang.org/cargo/reference/config.html))

.github/workflows/ci.yml
epgts: Here I'm only explaining what we're doing and why, not proposing a change.  The actions run as root, so if you had permission issues, you probably were running as postgres with data writable only by root.

tools/build
epgts: I think so too, I just wanted to capture the conditions.

tools/build
epgts: I mean `.target`. I always prefix my build directories with `.` so I don't trip over them all the time.

I experimented with the setting in `~/.cargo/config.toml` and it doesn't seem to be very useful: it overrides for ALL projects and JUST ONE DIRECTORY and resolves it relative to the home directory! Setting `build.target-dir = "target"` causes `/home/epg/target` to be used for all Rust builds which works about as well as you'd expect :)

PR #333, opened 2022-01-27
Revert tdigest aggregate builder

JLockerman:
This changed names in the public API we didn't want.

### DIFF
--- a/extension/src/tdigest.rs
+++ b/extension/src/tdigest.rs
@@ -1,8 +1,6 @@
 
 use std::{convert::TryInto, mem::take, ops::Deref};
 
-use aggregate_builder::aggregate;
-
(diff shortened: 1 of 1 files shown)


extension/src/tdigest.rs
```diff
@@ -1,8 +1,6 @@
 
 use std::{convert::TryInto, mem::take, ops::Deref};
 
-use aggregate_builder::aggregate;
```
epgts, line 4: Line 53 still has `#[aggregate]`...

JLockerman: yeah, my first attempt at reversion didn't work right. I redid it more incrementally, hopefully it'll work this time...

epgts (approved):
bors r+

PR #345, opened 2022-02-17
Start on testrunner

JLockerman:
Add a update testrunner. It'll install every version of the extension marked as `upgradeable_from` in `timescaledb_toolkit.control` and test that updates to the current version work correctly. At a high level:

1. For each version in `upgradeable_from`
    1. Checkout the corresponding tag in git,
    2. Build and install the extension at that tag,
    3. Set git back to the original state.
2. Build and install the extension at the original git state.
3. For each version in `upgradeable_from`
    1. create a database,
    2. install the old version of the extension,
    3. install some `timescaledb_toolkit` objects,
    4. update the extension,
    5. validate the extension is in the expected state.

**NOTE:** This version of the update testrunner _will_ move git's `HEAD`. Though git will warn on conflicts, and we do our best to reset the tree state before the script exits, we recommend only using it on a clean tree.

---

Some future work I'd like to do:
 - [ ] per-version update tests.
 - [ ] cache the old binaries and install scripts in CI so we don't have to rebuild them each time

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -308,5 +308,12 @@ jobs:
         key: ${{ runner.os }}-test-updates-target-1.6_a
         #restore-keys: ${{ runner.os }}-test-crates-target-PREVIOUS
 
+    - name: Cache old versions dir
+      uses: actions/cache@v2
+      with:
+        path: old-versions
+        key: ${{ runner.os }}-test-old-versions-1.6_a
+        #restore-keys: ${{ runner.os }}-test-old-versions-PREVIOUS
+
     - name: Run Update Tests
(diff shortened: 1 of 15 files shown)


epgts (approved):
I haven't read it deeply but from a high level it looks good and, as you say, the amount of time it takes is really nothing to worry about right now.

A few formatting nits I saw, nothing that need block merge:
- looks like at least some of the Rust files could use a rustfmt; i'd hit 'em all
- some extra newlines
- missing final newline in ci.yaml

JLockerman:
@epgts good catch! Should be fixed now
(someday we'll add format tests...)

epgts (approved):

PR #361, opened 2022-03-08
Versioned update scripts

JLockerman:
This commit redos how we do update scripts. Whereas before we would use essentially the same script for updating to a version as we'd use for installing the extension from scratch, with this commit we make behave differently for updates in order to make sure we create all and only those objects that are new to the version, and the other objects are only updated. We do this to handle name collisions better: if the user creates a function `foo()` and the extension also creates a function `foo()` in version `x.y.z` updating the extension to version `x.y.z` should _not_ overwrite the user function, and will now just error.

The strategy we use to achieve this is:

1. We keep lists of new objects for each version in `extension/lib/stabilization_info.rs` (this will also be useful for stabilization meta-tests).
2. When generating the update scripts we check if each object is new in the target version, if it is we `CREATE` the object, if not we update the object if needed.

As of this commit the logic is really only done for functions, aggregates, and types as these are the only kinds of stable objects we have. The commit also contains a skeleton for operators; these and casts will need to be finished before we stabilize timevector pipelines.

---

This commit also includes the commits from PR https://github.com/timescale/timescaledb-toolkit/pull/363


JLockerman:
(review was done in person)

PR #367, opened 2022-03-10
Switch stabilization tests to new info

JLockerman:
This commit switches the stabilization tests to the new stabilization info file, meaning there's exactly one central location that decides if things are stable. It also cleans the actual stabilization test code, and documents what invariants we're trying to maintain.

### DIFF
--- a/extension/src/hyperloglog.rs
+++ b/extension/src/hyperloglog.rs
@@ -194,7 +194,7 @@ fn hyperloglog_final_inner(
 }
 
 extension_sql!("\n\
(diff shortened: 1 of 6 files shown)


rtwalker (commented):
There is also a use of `INT` here
https://github.com/timescale/timescaledb-toolkit/blob/a6326345c30fdcdbbfa4f712901d296d38af6339/extension/src/lttb.rs#L97

WireBaron (approved):

PR #368, opened 2022-03-11
Implement into_table function for state_aggregate .

epgts:
Usage:
```
SELECT state, duration FROM toolkit_experimental.into_table(
    (SELECT toolkit_experimental.state_agg(ts, state) FROM states_test))
    ORDER BY state, duration;
 state | duration
-------+-----------
 ERROR |   3000000
 OK    | 106000000
 START |  11000000
```

### DIFF
--- a/extension/src/state_aggregate.rs
+++ b/extension/src/state_aggregate.rs
@@ -60,12 +60,16 @@ pub mod toolkit_experimental {
             None
         }
 
+        pub(super) fn states_as_str(&self) -> &str {
+            let states: &[u8] = self.states.as_slice();
+            // SAFETY: came from a String in `new` a few lines up
+            unsafe { std::str::from_utf8_unchecked(states) }
+        }
+
         fn state_str(&self, record: &DurationInState) -> &str {
             let beg = record.state_beg as usize;
             let end = record.state_end as usize;
-            let state = &self.states.as_slice()[beg..end];
-            // SAFETY: came from a String in `new` a few lines up
-            unsafe { std::str::from_utf8_unchecked(state) }
+            &self.states_as_str()[beg..end]
         }
     }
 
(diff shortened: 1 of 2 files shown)


extension/src/state_aggregate.rs
```diff
@@ -185,6 +189,18 @@ pub fn duration_in(state: String, aggregate: Option<StateAgg>) -> i64 {
         .unwrap_or(0)
 }
 
+#[pg_extern(immutable, parallel_safe, schema = "toolkit_experimental")]
+pub fn values(
```
WireBaron, line 193: Recommend changing the name to `display` (or if you come up with anything better, just not values).

extension/src/state_aggregate.rs
```diff
@@ -185,6 +189,18 @@ pub fn duration_in(state: String, aggregate: Option<StateAgg>) -> i64 {
...
+    agg: StateAgg<'_>,
+) -> impl std::iter::Iterator<Item = (name!(state, String), name!(duration, i64))> + '_ {
```
WireBaron, line 195: It would be awesome if we could make both this and `DurationInState` return a postgres `Interval` type instead of microseconds, but that's a bit of a heavier lift.

WireBaron (approved):
Rename `values` and this is good to go.

extension/src/state_aggregate.rs
jerryxwu: I also wonder whether `display` is good choice either since it might suggest for display only. I don't have a good suggestion either. The best that I came up is `unpack`.

rtwalker: `compute`? `capture`? `obtain`? `manifest`? `embody`? `occur`? `appear`? `emerge`? `unfold`? `objectify`?

Agree with @jerryxwu that `display` connotes a particular context that might suggest function is more limited that it actually is.

extension/src/state_aggregate.rs
WireBaron: Yeah, that's probably easier than the `AnyElement` task

WireBaron (approved):

PR #369, opened 2022-03-14
Add prototype gauge_agg and a few functions.

epgts:
The primary difference between a counter and a gauge is that counters
only increase, so apparent decreases are treated as resets followed by
continued increments from the pre-reset value.

To facilitate this, we move the reset accounting into a new
`CounterSummaryBuilder` and add a `GaugeSummaryBuilder` which does not
handle resets specially.

Functions:
- delta
- idelta_left
- idelta_right
- rollup

Also expand counter_agg's minimal tests just a bit.  Both aggregates
still need much more thorough testing.

TODO:
- consider deeper changes to `CounterSummary` and its Builders to
  protect invariants (currently bug-prone as innards are exposed)

Resolves issue #375.

### DIFF
--- a/crates/counter-agg/src/lib.rs
+++ b/crates/counter-agg/src/lib.rs
@@ -6,6 +6,8 @@ use std::fmt;
 
 
 pub mod range;
+
+#[cfg(test)]
 mod tests;
 
 #[derive(Debug, PartialEq)]
@@ -14,16 +16,33 @@ pub enum CounterError{
     BoundsInvalid,
 }
 
+// TODO Intent is for this to be immutable with mutations going through (and
+//  internal consistency protected by) the builders below.  But, we allow raw
+//  access to the extension to allow it to (de)serialize, so the separation is
+//  but a fiction for now.  If the only consequence of corruption is
+//  nonsensical results rather than unsound behavior, garbage in garbage out.
+//  But much better if we can validate at deserialization.  We can do that in
+//  the builder if we want.
 #[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
-pub struct CounterSummary {
+pub struct MetricSummary {
+    // TODO invariants?
     pub first: TSPoint,
     pub second: TSPoint,
     pub penultimate: TSPoint,
     pub last: TSPoint,
+    // Invariants:
+    // - num_changes > 0 if num_resets > 0
+    // - num_resets > 0 if num_changes > 0
+    // - reset_sum > 0 if num_resets > 0
+    // - num_resets > 0 if reset_sum > 0
     pub reset_sum: f64,
     pub num_resets: u64,
     pub num_changes: u64,
+    // TODO Protect from deserialization?  Is there any risk other than giving
+    //  nonsensical results?  If so, maybe it's fine to just accept garbage
+    //  out upon garbage in.
     pub stats: StatsSummary2D,
+    // TODO See TODOs in I64Range about protecting from deserialization.
     pub bounds: Option<range::I64Range>,
 }
 
@@ -43,12 +62,12 @@ fn to_seconds(t: f64)-> f64{
     t / 1_000_000_f64 // by default postgres timestamps have microsecond precision
 }
 
-/// CounterSummary tracks monotonically increasing counters that may reset, ie every time the value decreases
+/// MetricSummary tracks monotonically increasing counters that may reset, ie every time the value decreases
 /// it is treated as a reset of the counter and the previous value is added to the "true value" of the
 /// counter at that timestamp.
-impl CounterSummary {
-    pub fn new(pt: &TSPoint, bounds:Option<range::I64Range>) -> CounterSummary {
-        let mut n = CounterSummary{
+impl MetricSummary {
+    pub fn new(pt: &TSPoint, bounds:Option<range::I64Range>) -> MetricSummary {
+        let mut n = MetricSummary{
             first: *pt,
             second: *pt,
             penultimate: *pt,
@@ -63,8 +82,15 @@ impl CounterSummary {
         n
     }
 
+    fn reset(&mut self, incoming: &TSPoint) {
+        if  incoming.val < self.last.val {
+            self.reset_sum += self.last.val;
+            self.num_resets += 1;
+        }
+    }
+
     // expects time-ordered input
-    pub fn add_point(&mut self, incoming: &TSPoint) -> Result<(), CounterError>{
+    fn add_point(&mut self, incoming: &TSPoint) -> Result<(), CounterError>{
 
         if incoming.ts < self.last.ts {
             return Err(CounterError::OrderError);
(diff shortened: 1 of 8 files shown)


crates/counter-agg/src/lib.rs
```diff
@@ -14,16 +16,32 @@ pub enum CounterError{
     BoundsInvalid,
 }
 
+// TODO Intent is for this to be immutable with mutations going through (and
```
epgts, line 19: All  the TODOs on this struct are for post-merge consideration.

crates/counter-agg/src/lib.rs
```diff
@@ -96,23 +116,18 @@ impl CounterSummary {
     }
 
     fn single_value(&self) -> bool {
```
epgts, line 118: @davidkohn88 I'd particularly appreciate your thoughts on this change.

last == first does not mean single value on a gauge.  However, self.stats.n is only 1 when there is a single value if I'm reading it correctly. I added a test for gauge that fails in the expected manner without this fix. I can make existing counter tests fail by having this method return either true or false unconditionally, which suggests the existing tests do demonstrate this works.

WireBaron: I think `last == first` is still a valid test on a gauge.  First and last are `TSPoint`s which is a value/timestamp pair, so even if the value ends back where it started, the timestamp should be different.  That being said, `self.stats.n` might be a more direct check.

extension/src/counter_agg/gauge.rs
```diff
@@ -0,0 +1,579 @@
...
+#[pg_extern(immutable, parallel_safe, schema = "toolkit_experimental")]
+pub fn gauge_agg_summary_trans(
+    state: Internal,
+    value: Option<CounterSummary>,
+    fcinfo: pg_sys::FunctionCallInfo,
+) -> Internal {
+    gauge_agg_summary_trans_inner(unsafe { state.to_inner() }, value, fcinfo).internal()
+}
+fn gauge_agg_summary_trans_inner(
+    state: Option<Inner<CounterSummaryTransState>>,
```
davidkohn88, line 196: This feels weird still where it's a CounterSummary...

davidkohn88 (commented):
I mentioned to @epgts that I would love to talk through a couple things, but I'll leave what I have here, so the comments don't sit in limbo. 

The other main question I have is around how we tell which one we're working on and whether the old stays the same on disk or if we need to futz with the current on-disk format and invoke the versioning procedure. ("Invoke the versioning procedure!" I am imagining it being said by someone in military dress on a warship of some sort). 

Anyway, I do think we should have some way of identifying which one we're doing from the data structure on disk? But we might just need a different name for the type and it doesn't matter if they're the same internally? I dunno. But I might be totally off base.

crates/counter-agg/src/lib.rs
```diff
@@ -284,4 +299,90 @@ impl fmt::Display for CounterError {
...
         }
     }
-}
\ No newline at end of file
+}
+
+#[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
+pub struct CounterSummaryBuilder(CounterSummary);
+
+impl CounterSummaryBuilder {
```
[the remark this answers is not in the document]
epgts: The great unsolved problem in our field :)

Your suggestion of `MetricSummary` is for the underlying type which I've currently left named `CounterSummary`, right? So your suggestion is:
- MetricSummary
- CounterSummaryBuilder
- GaugeSummaryBuilder

Those look fine to me. I'll wait for Brian and Ryan to weigh in on our second naming issue in two days before I make any changes :)

extension/src/counter_agg/gauge.rs
epgts: Is this about the name or something else? Other than the name, I think this is the right structure. Happy to discuss more.

crates/counter-agg/src/lib.rs
epgts: Yep the original check was fine! I reverted this change.

extension/src/counter_agg.rs
```diff
@@ -1,3 +1,5 @@
+pub mod gauge;
```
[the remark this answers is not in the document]
epgts: Can you elaborate?

For the sake of keeping the change minimal, I've built this `gauge` mod on the existing `counter_agg` mod, so this needs to be contained within.

As a follow-up, I think we should factor the common bits out of `counter_agg` into a third mod (within the extension crate) both these modules would build on. We could make `gauge_agg` a sibling rather than child of `counter_agg` at that point.

Would that get us where you want to be?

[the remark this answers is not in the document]
epgts: With the further separation of the aggregates we discussed last week, this became the obvious move after all. Done now. Thanks!

extension/src/counter_agg/gauge.rs
epgts: Made the changes we discussed.

docs/gauge_agg.md
```diff
@@ -0,0 +1,76 @@
+# Gauge Aggregates [<sup><mark>experimental</mark></sup>](/docs/README.md#tag-notes)
+
+A gauge is a metric similar to a counter, with the primary difference being
+that it measures a value that varies up and down over time, rather than an
+ever-increasing COUNT of the number of times something happened.
+Examples include resource utilization metrics, precipitation levels,
+or temperatures.
+
+`gauge_agg` currently shares implementation with `counter_agg` but without the
```
davidkohn88, line 9: This feels inside baseball, which is fine here, but it would be good to figure out some more user facing explanations here at some point.

docs/gauge_agg.md
```diff
@@ -0,0 +1,76 @@
+# Gauge Aggregates [<sup><mark>experimental</mark></sup>](/docs/README.md#tag-notes)
+
+A gauge is a metric similar to a counter, with the primary difference being
+that it measures a value that varies up and down over time, rather than an
+ever-increasing COUNT of the number of times something happened.
+Examples include resource utilization metrics, precipitation levels,
```
davidkohn88, line 6: Do we want a specific example of each and a comparison? This is a good explanation, but I think an illustration like the number of visitors to a website is a counter, because x and this is a gauge because why and this is how you can tell the difference.  It feels like that would help but I might be wrong.

docs/gauge_agg.md
```diff
@@ -0,0 +1,76 @@
...
+unordered gauge aggregate in the future.
+
+# Test table
+
+Examples below are tested against the following table:
+
+```SQL ,non-transactional
+SET TIME ZONE 'UTC';
+CREATE TABLE gauge_test (
+    measure_id      BIGINT,
```
davidkohn88, line 21: Maybe we can use a real example with some more representative names than the testing one? If we develop the example above about diffs between gauges and counters that might help. I just find sometimes the concreteness can help with understanding.

davidkohn88 (commented):
Added some docs comments, but no need to address now, more things to consider as we move forward. Otherwise LGTM.

Overarching docs comment is we may want to prototype a section on metrics and distinguishing between counters and gauges, maybe even ask Promscale for some help on that so that folks can better understand what we're doing here. We'd take point on explaining that we've tried to keep the interfaces more similar than Prometheus etc. But if they want difft interfaces they should just ask (ie `increase()`)?

docs/gauge_agg.md
WireBaron: @davidkohn88 I think the expectation is that we're using this section more as developer level docs and as a driver for our doctester, and less as anything user facing.  The goal is to not be having the same material duplicated here and in the on-line docs, but of course, this is very much an evolving design.

WireBaron (approved):
Looks like all of the feedback has been addressed.  We do still need to figure out what to do with the docs area, but it's good to see what you had suggested for that.

PR #372, opened 2022-03-16
Improve tools/build flexibility for local builds.

epgts:
- Add parameters for path to pg_config and server TCP port.
- Run local sql-doctester, not whatever is installed.

### DIFF
--- a/tools/build
+++ b/tools/build
@@ -19,20 +19,23 @@ usage() {
     die 'build [ -n -pg1[234] ] ( test-crates | test-extension | test-post-install | test-doc | clippy )'
 }
 
-require_pgversion() {
-    [ -n "$pgversion" ] || die 'specify one of -pg12 | -pg13 | -pg14'
+require_pg_version() {
+    [ -n "$pg_version" ] || die 'specify one of -pg12 | -pg13 | -pg14'
 }
 
-find_pgconfig() {
-    require_pgversion
-    full_version=$(
-        cd "$HOME/.pgx"
-        set -- $pgversion.*/pgx-install
-        [ "$1" = "$pgversion.*/pgx-install" ] && die "$pgversion not installed in $HOME/.pgx"
-        set -- $(dirname $*)
-        [ $# -eq 1 ] || die "too many installations for $pgversion, found: $*"
-        echo $1)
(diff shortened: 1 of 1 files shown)


tools/build
```diff
@@ -19,20 +19,23 @@ usage() {
     die 'build [ -n -pg1[234] ] ( test-crates | test-extension | test-post-install | test-doc | clippy )'
 }
 
-require_pgversion() {
-    [ -n "$pgversion" ] || die 'specify one of -pg12 | -pg13 | -pg14'
+require_pg_version() {
+    [ -n "$pg_version" ] || die 'specify one of -pg12 | -pg13 | -pg14'
```
epgts, line 23: Here we have pure renames

tools/build
```diff
@@ -19,20 +19,23 @@ usage() {
...
-    full_version=$(
-        cd "$HOME/.pgx"
-        set -- $pgversion.*/pgx-install
-        [ "$1" = "$pgversion.*/pgx-install" ] && die "$pgversion not installed in $HOME/.pgx"
-        set -- $(dirname $*)
-        [ $# -eq 1 ] || die "too many installations for $pgversion, found: $*"
-        echo $1)
-    pg_config="$HOME/.pgx/$full_version/pgx-install/bin/pg_config"
+find_pg_config() {
+    if [ -z "$pg_config" ]; then
```
epgts, line 27: Here we add a check if pg_config is already specified before looking for it in ~/.pgx (which remains the default and what we use under the github action).

tools/build
```diff
@@ -69,10 +72,20 @@ while [ $# -gt 0 ]; do
...
+            shift
+            ;;
+
         -pg1[234])
-            pgversion=${arg#-pg}
-            pg=pg$pgversion
-            pgport=288$pgversion
+            pg_version=${arg#-pg}
+            pg=pg$pg_version
+            [ -z "$pg_port" ] && pg_port=288$pg_version
```
epgts, line 88: Similarly, only use pgx's port numbering scheme if port wasn't already specified.

epgts (commented):
Adding explanatory comments in case that helps y'all with the review :)

WireBaron (approved):

PR #378, opened 2022-03-28
Return INTERVAL from duration_in(TEXT, StateAgg).

epgts:


extension/src/state_aggregate.rs
```diff
@@ -458,14 +469,17 @@ mod tests {
...
                     None,
                     None,
                 )
                 .first()
                 .get_one()
                 .unwrap();
-            assert_eq!(60000000, result);
+            // XXX what have i done wrong?  it worked interactively...
+            //  left: `60`,
+            // right: `4633641066610819072
```
epgts, line 481: @WireBaron any ideas?

extension/src/state_aggregate.rs
```diff
@@ -248,6 +249,16 @@ struct Record {
     time: i64,
 }
 
+fn microseconds_to_interval(micros: i64) -> *const pg_sys::Interval {
```
epgts, line 252: It took me a while to come up with this, and it finally came as basically a wild guess. I couldn't find any example of us doing this already. It does work locally, though I'm not sure whether that's by accident, nor am I certain it's not a leak...

JLockerman: [Here's](https://doxygen.postgresql.org/backend_2utils_2adt_2timestamp_8c_source.html#l01497) the canonical postgres code to create an interval from a bunch of integers. I think the code you have looks correct, though it could perhaps be simplified to
```rust
let i = pg_sys::palloc0(std::mem::size_of::<pg_sys::Interval>()) as *mut pg_sys::Interval;
(*i).time = micros;
i
```
or even
```rust
let i = pg_sys::palloc0(std::mem::size_of::<pg_sys::Interval>()) as *mut pg_sys::Interval;
*i = pg_sys::Interval {
    month: 0,
    day: 0,
    time: 0,
};
i
```

extension/src/state_aggregate.rs
JLockerman: 1. The problem is that `result: i64` when the return type of `EXTRACT()` is `Numeric` which is still a pointer-type. If you want this version of the code to work you need to do `EXTRACT(...)::bigint`.
2. Generally for these kinds of errors I find that text-based comparisons can make things easier to spot. If you do `EXTRACT(...)::text` it will return as a `String` and you can get the ground truth of what Postgres thinks is going on and avoid all cross-language errors.
3. Doing `duration_in(...)::text` might be better than `EXTRACT(...)::bigint` here because it will tell us the entire Interval, though I guess us since epoch does basically that 🤔 

(My first guess would be timezones, thought I don't _think_ that should be applicable here)

epgts: Thanks!

extension/src/state_aggregate.rs
epgts: Yeah I didn't think of assigning the whole struct, which is the obvious move and let's me have rustc make sure I don't forget to set any fields.

I guess `palloc0` is analogous to `calloc` in that it allocates and zeroes, but that's not generally safe and I prefer to avoid it. That point of contention is much older than my career and I'd rather not litigate it here, just explaining why I avoided palloc0 :). The postgres code you linked to is also explicit rather than implicit :)

Anyway it's much better now: thanks!

extension/src/state_aggregate.rs
```diff
@@ -279,11 +292,13 @@ mod tests {
                 None,
             );
             assert_eq!(
-                31536120000000,
+                "8760:02:00",
```
WireBaron, line 295: Presumably we wouldn't want to call out to postgres for every timestamptz calculation, but it is worth considering that this a bit different from the postgres result:
```
postgres=# select ('2020-12-31 00:02:00+00'::timestamptz - '2020-01-01 00:00:00+00'::timestamptz)::interval;
     interval      
-------------------
 365 days 00:02:00
(1 row)
```

WireBaron (commented):
Timestamps are a bit weird, and this does return different results from what postgres does for some calculations.  I think this is actually fine, but I do want to think on this a little more.

extension/src/state_aggregate.rs
epgts: It took me a while to figure out how and where postgresql is doing this, but I did find it, and I've documented it here, and I'm calling the same function.

Thanks!

WireBaron (approved):
Looks good!

PR #379, opened 2022-03-30
Handle NULL output from our aggregates

### LINKED ISSUE #362: Incorrect NULL handling in hyperloglog
version: `1.6.0-dev` (c49a737, 2022-03-09)

```SQL
SELECT hyperloglog(1, null::int);
```
fails with an error
```
ERROR:  returned Datum was NULL
CONTEXT:  extension/src/hyperloglog.rs:32:1
```
while it should return a NULL hyperloglog.

JLockerman:
`pgx::Internal` is inconsistent in that a NULL input to a function taking `Internal` works fine and results in an `Internal` containing `None` while returning an Internal containing `None` results in an error. This commit works around the issue by making our functions return an `Option<Internal>` if they can return a SQL NULL.

fixes #362

### DIFF
--- a/crates/aggregate_builder/Readme.md
+++ b/crates/aggregate_builder/Readme.md
@@ -98,7 +98,7 @@ pub mod aggregate_name {
         __inner: pgx::Internal,
         value: RustType,
         __fcinfo: pg_sys::FunctionCallInfo,
-    ) -> Internal {
+    ) -> Option<Internal> {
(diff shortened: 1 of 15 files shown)


epgts (approved):

WireBaron (approved):
This looks good.  I would like to see a test for proper handling of `timestamp, NULL` in counter_agg, but that isn't necessary to close this issue.

PR #380, opened 2022-03-31
Prepare for the `1.7.0` cycle

JLockerman:


jerryxwu (approved):

rtwalker:
Perhaps unrelated to this PR, but I've been wondering what this version number corresponds to?
https://github.com/timescale/timescaledb-toolkit/blob/e3ade4038310dc2ea04a3c86402673df5f55c9f1/extension/Cargo.toml#L3

JLockerman:
@rtwalker the Cargo.toml version corresponds to absolutely nothing! If we were to use this thing as a non-git dependency of something else that's what Cargo would use to determine the version, but as is it's meaningless, and we only have it there b/c IIUC it's required.

Issue #373, opened 2022-03-22
Implement an Approximate Count aggregate using Count-Min Sketch


rtwalker:
#### Description ####
An implementation of `approx_count` originally proposed in #6. The proposed implementation would sit atop a new aggregate (`count_agg`?) that uses the Count-Min Sketch data structure. Interrogating the count aggregate via `approx_count` would give an estimated frequency. 

The Count-Min Sketch is an array of counters parameterized by the number of rows `d` in the array and by the number of buckets in a row `w`. If we take `w = 2/ε` and `d = ln(1/δ)`, then we have that the estimated frequency of a given item has error at most `εN` with probability at least `1 − δ`, where `N` is the number of elements present in the data stream. I think we could make the `ε` and `δ` optional parameters to `count_agg` rather than hard-coding in a level of performance guarantees (and space complexity) for all sketches.

#### Proposed API ####
To get the count aggregate use
```sql
SELECT count_agg(value, 0.01, 0.01) FROM tbl;
```
and to query the estimated count of a given item "foo" in the table use
```sql
SELECT approx_count(count_agg(value), "foo") FROM tbl;
```
I think as a first pass I'll just get this to work on `text` columns before exploring supporting a broader number of types.

#### Alternatives ####
Could use a Count-Sketch (which does a median calculation instead of a min). 

From pg. 16 of http://dimacs.rutgers.edu/~graham/

> **1.3.2.1 Refining Count-Sketch and Count-Min Sketch guarantees**
We have seen that the Count-Sketch and Count-Min sketch both allow f (i) to
be approximated via somewhat similar data structures. They differ in provid-
ing distinct space/accuracy trade-offs: the Count sketch gives ε√F2 error with
O(1/ε2) space, whereas the Count-Min sketch gives εN error with O(1/ε)
space. In general, these bounds cannot be compared: there exist some fre-
quency vectors where (given the same overall space budget) one guarantee
is preferable, and others where the other dominates. Indeed, various experi-
mental studies have shown that over real data it is not always clear which is
preferable [23].

There are fewer hash functions to keep up with in a Count-Min sketch, so I'm inclined to try it first.

EDIT: I think it might also be worth seeing how well we can do using the SpaceSaving algorithm/data structure underlying`freq_agg`. I think it would make it much easier on users if we could consolidate into one aggregate here instead of offering both a `freq_agg` and a `count_agg` that support different functions. Could use `approx_freq` instead of `approx_count` for consistency if we go that route.

[...]

PR #388, opened 2022-04-13
Create Count-Min Sketch crate

rtwalker:
Implement Count-Min sketch probabilistic data structure with basic add, subtract, and (point) estimate functions.

Addresses some of #373

### DIFF
--- a/crates/count-min-sketch/src/lib.rs
+++ b/crates/count-min-sketch/src/lib.rs
@@ -1,0 +1,257 @@
+//! Count-Min Sketch implementation in Rust
+//!
+//! Based on the paper:
+//! <http://dimacs.rutgers.edu/~graham/pubs/papers/cm-full.pdf>
+
+use std::{
+    fmt,
+    hash::{Hash, Hasher},
+};
+
+#[allow(deprecated)]
+use std::hash::SipHasher;
+
+use serde::{Deserialize, Serialize};
+
+/// The CountMinHashFn is a data structure used to hash items that are being
+/// added to a Count-Min Sketch.
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
+#[repr(C)]
+pub struct CountMinHashFn {
+    key: u64,
+}
+
+const SEED: u64 = 0x517cc1b727220a95; // from FxHash
+
+impl CountMinHashFn {
+
+    /// Creates a new CountMinHashFn whose hash function key is equal to `key`.
+    pub fn with_key(key: u64) -> Self {
+        Self { key }
+    }
+
+    /// Computes the hash of `item` according to the hash function and returns
+    /// the bucket index corresponding to the hashed value.
+    ///
+    /// The returned value will be between 0 and (`nbuckets` - 1).
+    #[allow(deprecated)]
+    pub fn hash_into_buckets<T: Hash>(&self, item: &T, nbuckets: usize) -> usize {
+        let (key1, key2) = (self.key, SEED);
+        let mut hasher = SipHasher::new_with_keys(key1, key2);
+        item.hash(&mut hasher);
+        let hash_val = hasher.finish();
+        (hash_val % (nbuckets as u64)) as usize
+    }
+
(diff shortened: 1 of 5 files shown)


crates/count-min-sketch/src/lib.rs
```diff
@@ -0,0 +1,240 @@
+//! Count-Min Sketch implementation in Rust
+//! Based on the paper:  http://dimacs.rutgers.edu/~graham/pubs/papers/cmencyc.pdf
```
epgts, line 2: nit: add angle-brackets around this (cargo doc complains)

crates/count-min-sketch/src/lib.rs
```diff
@@ -0,0 +1,240 @@
...
+    counters: Vec<Vec<isize>>,
+    hashfuncs: Vec<HashFunction>,
+    width: usize,
+    depth: usize,
+}
+
+impl CountMinSketch {
+    pub fn new(epsilon: f64, delta: f64) -> Self {
+        assert!(epsilon.lt(&1f64) && epsilon.gt(&0f64));
+        assert!(delta.lt(&1f64) && delta.gt(&0f64));
```
epgts, line 51: Why `delta.lt(&1f64)` rather than `delta < 1.0`? Also consider writing numeric comparisons using only < and ascending left to right so: `0.0 < delta && delta < 1.0`

rtwalker: I think this happened because I had the two lines below in view as I wrote these and having to use `.exp()` and `.ln()` there made this the subconscious choice. 

Unquestionably, `0.0 < delta && delta < 1.0` is better.

crates/count-min-sketch/src/lib.rs
```diff
@@ -0,0 +1,314 @@
...
+/// to insertions or deletions, or other transactions), it is capable of working
+/// over streams of updates, at high rates.[1]
+///
+/// [1]: <http://dimacs.rutgers.edu/~graham/pubs/papers/cmencyc.pdf>
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct CountMinSketch {
+    width: usize,
+    depth: usize,
+    hashfuncs: Vec<CountMinHashFn>,
+    counters: Vec<Vec<i64>>,
```
epgts, line 64: Is it worth specifically calling out right here that hashfuncs and counters must both be at least `depth` long? I see both places you construct it now enforce it, and I guess there's little chance we could introduce a third construction that wouldn't explode in testing but... well call me paranoid :)

crates/count-min-sketch/src/lib.rs
```diff
@@ -0,0 +1,314 @@
...
+
+        write!(f, "+------++")?;
+        for _ in 0..self.width {
+            write!(f, "--------+")?;
+        }
+        writeln!(f)
+    }
+}
+
+#[cfg(test)]
```
epgts, line 252: Looks like these are all purely in terms of public interface (good) and therefore can move to `crates/count-min-sketch/tests/lib.rs`

epgts (commented):
Thanks!

crates/count-min-sketch/src/lib.rs
rtwalker: I don't mind enabling a little paranoia :)

Do you want a comment here? in the docstring? 

I spent a little while trying to make something like this work:
```rust
struct CountMinSketch<const D: usize, const W: usize> {
    hashfuncs: [CountMinHashFn; D],
    counters: [[i64; W]; D]
}
```
but I don't think that'd work in the extension.

epgts: I don't think you can get around run-time assertions. I just like to have documentation too.

In the doc string is good as callers can see it live if their editor is smart enough, and it also allows you to reference other fields in a way that `cargo doc` will check in case we miss the comments on a rename (e.g. `/// must be at least [Self::depth] long`).

Thanks!

epgts (approved):

PR #389, opened 2022-04-13
Create typed specialization for freq_agg and topn_agg

### LINKED ISSUE #376: Specialize freq_agg and topn_agg for bigints and text rather than supporting AnyElement
Postgres functions that return values of type `AnyElement` are inherently ugly due to the requirement that the user provide a value exactly matching the return type as an input to the function.  This makes for an ugly and unpleasant user experience.  The goal of supporting `AnyElement` in these aggregates was to free the user from having to worry about the type, and it very much failed in this as now the user has to be exactly aware of the type in every instance where they want to get a value out of the aggregate.

A friendlier alternative would be implementations that support (at least implicit casts from) the types we expect or see customers using.  Ints and strings seem the most like values and should be our first priorities.  It may be worth having an AnyElement specialization available as a fallback for users who absolutely need it, but it should probably get its own unique name (`freq_agg_generic`?) to ensure no one accidentally uses it.

WireBaron:
The `AnyElement` implementation for our various spacesaving implementations are very convenient to allow users to build them using any arbitrary type of data, but turn out to be very ugly to then try to extract data from.

This change creates a 64-bit int specialization and a text specialization of our frequency aggregates.  These can be used without providing a type argument to get data out, which ends up being much cleaner.

The existing `AnyElement` implementations have been moved to `raw_freq_agg` and `raw_topn_agg` to allow easier implicit casts to the new specializations.

The need to provide distinct types to postgres for each specialization has resulted in many similar looking data structures and functions in this code.  I'm very open to hearing suggestions on how this could be done more cleanly from any reviewers.

fixes #376

### DIFF
--- a/extension/src/frequency.rs
+++ b/extension/src/frequency.rs
@@ -24,14 +24,16 @@ use crate::{
     palloc::{Inner, Internal, InternalAsValue, ToInternal},
     pg_any_element::{PgAnyElement, PgAnyElementHashMap},
     pg_type,
-    raw::bytea,
+    raw::{bytea, text},
     ron_inout_funcs,
 };
 
 use spfunc::zeta::zeta;
 use statrs::function::harmonic::gen_harmonic;
 
-use crate::frequency::toolkit_experimental::SpaceSavingAggregate;
+use crate::frequency::toolkit_experimental::{
+    SpaceSavingAggregate, SpaceSavingBigIntAggregate, SpaceSavingTextAggregate,
+};
 
 // Helper functions for zeta distribution
 
@@ -203,7 +205,19 @@ impl SpaceSavingTransState {
         }
     }
 
-    fn topn_agg_from_type_id(skew: f64, nval: u32, typ: pg_sys::Oid, collation: Option<Oid>) -> Self {
+    fn topn_agg_from_type_id(
+        skew: f64,
+        nval: u32,
+        typ: pg_sys::Oid,
+        collation: Option<Oid>,
+    ) -> Self {
+        if nval == 0 {
+            pgx::error!("topn aggregate requires an n value > 0")
+        }
+        if skew <= 1.0 {
+            pgx::error!("topn aggregate requires a skew factor > 1.0")
+        }
+
         let prob_eq_n = zeta_eq_n(skew, nval as u64);
         let prob_lt_n = zeta_le_n(skew, nval as u64 - 1);
 
@@ -212,7 +226,8 @@ impl SpaceSavingTransState {
             indices: PgAnyElementHashMap::new(typ, collation),
             total_vals: 0,
             freq_param: skew,
-            max_size: nval - 1 + SpaceSavingTransState::max_size_for_freq(prob_eq_n / (1.0 - prob_lt_n)),
+            max_size: nval - 1
+                + SpaceSavingTransState::max_size_for_freq(prob_eq_n / (1.0 - prob_lt_n)),
             topn: nval,
         }
     }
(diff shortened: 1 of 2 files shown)


extension/src/frequency.rs
```diff
@@ -399,6 +399,92 @@ pub mod toolkit_experimental {
...
+    pg_type! {
+        #[derive(Debug)]
+        struct SpaceSavingBigIntAggregate<'input> {
+            num_values: u32,
+            topn: u32,
+            values_seen: u64,
+            freq_param: f64,
+            counts: [u64; self.num_values], // JOSH TODO look at AoS instead of SoA at some point
+            overcounts: [u64; self.num_values],
+            datums: [i64; self.num_values],
```
JLockerman, line 412: I believe something like this will work now
```rust
pg_type! {
        #[derive(Debug)]
        struct SpaceSavingBigIntAggregate<'input> {
            num_values: u32,
            topn: u32,
            values_seen: u64,
            freq_param: f64,
            elements: [Entry; self.num_values],
    }
}


flat_serialize! {
    #[derive(Debug)]
    struct Entry {
        count: u64,
        overcount: u64,
        datum: i64,
    }
}
```
Though I can't say if it would be a good idea

extension/src/frequency.rs
```diff
@@ -430,6 +536,56 @@ pub fn topn_agg_with_skew_trans(
...
+) -> Option<Internal> {
+    if n <= 0 {
+        pgx::error!("topn aggregate requires an n value > 0")
+    }
+    if skew <= 1.0 {
+        pgx::error!("topn aggregate requires a skew factor > 1.0")
+    }
+
+    // Important: this variable will hold the lifetime of the varlena until we copy it in freq_agg_trans()
+    let text = value.map(pgx::rust_str_to_text_p);
```
JLockerman, line 579: if we're doing this I'd recommend dust using an `Option<crate::raw::text>` and using `pg_detoast_datum_copy()`

extension/src/frequency.rs
```diff
@@ -445,6 +601,36 @@ pub fn freq_agg_trans(
...
+
+#[pg_extern(schema = "toolkit_experimental", immutable, parallel_safe)]
+pub fn freq_agg_text_trans(
+    state: Internal,
+    freq: f64,
+    value: Option<&str>,
+    fcinfo: pg_sys::FunctionCallInfo,
+) -> Option<Internal> {
+    // Important: this variable will hold the lifetime of the varlena until we copy it in freq_agg_trans()
+    let text = value.map(pgx::rust_str_to_text_p);
```
JLockerman, line 626: if we're doing this I'd recommend dust using an `Option<crate::raw::text>` and using `pg_detoast_datum_copy()`

extension/src/frequency.rs
```diff
@@ -568,6 +806,40 @@ requires = [topn_agg_trans, space_saving_final, space_saving_combine, space_savi
...
+        parallel = safe\n\
+    );\n\
+",
+name = "topn_bigint_agg",
+requires = [topn_agg_bigint_trans, space_saving_bigint_final, space_saving_combine, space_saving_serialize, space_saving_deserialize],
+);
+
+extension_sql!("\n\
+    CREATE AGGREGATE toolkit_experimental.topn_agg(\n\
+        integer, TEXT\n\
```
JLockerman, line 826: the arguments to the aggregates should have names, one because I the update scripts will need them once stable, but more importantly for documentation in `\df`. Actually, can this use the aggregate builder instead?

JLockerman (commented):
Needs some formatting and I have a couple of nits, but over all it looks good.
Not sure I should be the approver anymore, so I'll leave that for someone else 😉

extension/src/frequency.rs
```diff
@@ -712,6 +1171,64 @@ pub fn min_frequency(
     0.
 }
 
+#[pg_extern(immutable, parallel_safe, name="max_frequency", schema = "toolkit_experimental")]
+pub fn max_bigint_frequency(
+    agg: SpaceSavingBigIntAggregate<'_>,
+    value: i64,
+) -> f64 {
+    for (idx, datum) in agg.datums.iter().enumerate() {
+        if value == datum as i64 {
```
epgts, line 1180: I've run out of time, but the same approach can be applied to max and min frequency functions: implement the algorithm in generic function parametrized over data type and conversion function.

extension/src/frequency.rs
```diff
@@ -399,6 +399,92 @@ pub mod toolkit_experimental {
...
+
+    impl<'input> From<&SpaceSavingTransState> for SpaceSavingBigIntAggregate<'input> {
+        fn from(trans: &SpaceSavingTransState) -> Self {
+            assert_eq!(trans.type_oid(), pg_sys::INT8OID);
+
+            let mut values = Vec::new();
+            let mut counts = Vec::new();
+            let mut overcounts = Vec::new();
+
+            for entry in &trans.entries {
```
epgts, line 424: I demonstrate how to genericize this in commit 31e12c3e9f38f07b9fce3ea8279ae5b4a1674ad6

extension/src/frequency.rs
```diff
@@ -430,6 +536,56 @@ pub fn topn_agg_with_skew_trans(
...
 
+#[pg_extern(schema = "toolkit_experimental", immutable, parallel_safe)]
+pub fn topn_agg_with_skew_bigint_trans(
+    state: Internal,
+    n: i32,
+    skew: f64,
+    value: Option<i64>,
+    fcinfo: pg_sys::FunctionCallInfo,
+) -> Option<Internal> {
+    if n <= 0 {
```
epgts, line 547: We can avoid repeating these checks by moving them into topn_agg_from_type_id as show in commit c9eb12fe91eb232ac557500d0c0513921e1f351c

extension/src/frequency.rs
```diff
@@ -682,6 +1049,98 @@ fn topn_internal(
...
+    let n = agg.topn as i32;
+    topn_bigint(agg, n)
+}
+
+fn topn_bigint_internal(
+    agg: SpaceSavingBigIntAggregate<'_>,
+    max_n: i32,
+    min_freq: f64,
+) -> impl std::iter::Iterator<Item = i64> + '_ {
+    let iter = agg
```
epgts, line 1083: This is pretty complex, what with zipping two things together and having 3 different exit criteria. It would be great not to duplicate that algorithm. I show how to factor it out in commit d6e293a9ae9d6a5ae7f1f880feb792d8b5b829a4.

epgts (commented):
I offer a few suggestions for reducing duplication in branch https://github.com/timescale/timescaledb-toolkit/compare/br/remove_freq_anyelem...eg/remove_freq_anyelem?expand=1

The same approach can be applied in a couple other places.

WireBaron (commented):
Incorporated feedback, last change in new batch includes a rustfmt, might be easier to see diff looking at individual changes.

extension/src/frequency.rs
WireBaron: Yeah, even cleaner with the new change.

Issue #386, opened 2022-04-12
MetricSummary::bounds cleanup


epgts:
The `bounds` field of `MetricSummary` represents `unbounded` in at least two different ways (`None`, `Some((None, None))`), and possibly 4.  Furthermore, the underlying `I64Range` needlessly allows invalid states.

### 4 different ways to represent unbounded?

- `None`
- `Some((None, None))`
- `Some((None, _))`
- `Some((_, None))`

If so, then some of `I64Range`'s methods disagree on the last two states, e.g. `has_infinite` returns `true` for either one, but `contains` treats them as unbounded on only one end; other differences may exist.

But maybe the "has" in `has_infinite` means that's right and it really is saying "either end is infinite". Even if so, this complexity goes unused.

### `I64Range` validity

`I64Range` allows invalid states by exposing its field as `pub` and thus needs a `pub is_valid` method which may return false.

#### `I64Range`

- Model unbounded left as `i64::MIN` and right as `i64::MAX`.
- Enforce validity in `extend` method (DONE) and new factory functions `empty` and `infinite`.
- Expose `#[inline]` `left` and `right` methods.
- Privatize `left` and `right` fields and `is_valid` method.

#### `MetricSummary`

- Remove `Option` from `MetricSummary::bounds`, with default of `I64Range::infinite()` instead of `None`.

PR #390, opened 2022-04-14
MetricSummary::bounds cleanup

epgts:
MetricSummary:
- Delegate to I64Range for infinity rather than duplicating the concept.

I64Range
- Cleanup and encapsulate I64Range's handling of infinite bounds.
- Enforce validity in new factory functions `empty` and `infinite`.
  (method `extend` already did)
- Expose #[inline] left and right methods.
- Privatize left and right fields.
- Factor out new method `both` out of various places that needed access
  to finite left and right.

Closes issue https://github.com/timescale/timescaledb-toolkit/issues/386.

### DIFF
--- a/crates/counter-agg/src/lib.rs
+++ b/crates/counter-agg/src/lib.rs
@@ -1,11 +1,10 @@
-
 use time_series::TSPoint;
 use stats_agg::{XYPair, stats2d::StatsSummary2D};
-use serde::{Deserialize, Serialize};
 use std::fmt;
 
 
 pub mod range;
+pub mod stable;
 
 #[cfg(test)]
 mod tests;
@@ -23,7 +22,7 @@ pub enum CounterError{
 //  nonsensical results rather than unsound behavior, garbage in garbage out.
 //  But much better if we can validate at deserialization.  We can do that in
 //  the builder if we want.
-#[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
+#[derive(Clone, Debug, PartialEq)]
 pub struct MetricSummary {
     // TODO invariants?
     pub first: TSPoint,
@@ -43,7 +42,7 @@ pub struct MetricSummary {
     //  out upon garbage in.
     pub stats: StatsSummary2D,
     // TODO See TODOs in I64Range about protecting from deserialization.
-    pub bounds: Option<range::I64Range>,
+    pub bounds: range::I64Range,
 }
 
 // Note that this can lose fidelity with the timestamp, but it would only lose it in the microseconds,
@@ -66,8 +65,8 @@ fn to_seconds(t: f64)-> f64{
 /// it is treated as a reset of the counter and the previous value is added to the "true value" of the
 /// counter at that timestamp.
 impl MetricSummary {
-    pub fn new(pt: &TSPoint, bounds:Option<range::I64Range>) -> MetricSummary {
-        let mut n = MetricSummary{
+    pub fn new(pt: &TSPoint, bounds: range::I64Range) -> MetricSummary {
+        let mut n = MetricSummary {
             first: *pt,
             second: *pt,
             penultimate: *pt,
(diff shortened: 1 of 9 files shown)


crates/counter-agg/src/lib.rs
```diff
@@ -204,42 +203,56 @@ impl MetricSummary {
...
             return Err(CounterError::BoundsInvalid);
         }
+        let (left, right);
+        match self.bounds.both() {
+            None => return Err(CounterError::BoundsInvalid),
+            Some((l, r)) => {
+                left = l;
+                right = r;
+            }
+        }
```
WireBaron, line 241: Seems there really should be a cleaner way to do this `unwrap_or_return` behavior, but if there is I don't know it.

crates/counter-agg/src/stable.rs
```diff
@@ -0,0 +1,62 @@
+//! These data structures are stable, meaning they may not change even in
+//! their layout in memory, as the raw bytes-in-memory are serialized and
+//! exchanged by PostgreSQL.
```
WireBaron, line 3: @JLockerman can correct me if I have this wrong, but this is not actually the case for these objects.  The `MetricSummary` is only used in the internal state, which is never persisted across version upgrades.  The objects that aren't allowed to change are the `pg_type` declared final objects, `CounterSummary` and `FlatSummary` in this case.

crates/counter-agg/src/stable.rs
```diff
@@ -0,0 +1,62 @@
...
+
+use stats_agg::stats2d::StatsSummary2D;
+use time_series::TSPoint;
+
+use crate::range::I64Range;
+
+#[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
+// TODO Our one serialization test (counter_byte_io) passes, but is that just luck?
+//#[repr(C)]
+pub struct MetricSummary {
```
WireBaron, line 20: I don't love creating this new struct that only differs by `bounds` being an option.  Is this because you thought the original `MetricSummary` wasn't something that could be changed?  In this case, I'd prefer changing the original object.  Failing that, maybe we could make `bounds` non-public and add a `bounds()` accessor that returns an option?

crates/counter-agg/src/stable.rs
epgts: Oh, that would be a relief! If that's so, why do we test it? If we don't actually require stability of that serialization, we're just paying a tax every time we change it:

https://github.com/timescale/timescaledb-toolkit/blob/main/extension/src/counter_agg.rs#L972

JLockerman: TLDR: you're both partially right, the situation is confusing, and the problem goes away once we stop caring about timescaledb versions older than 2.7.0

Ah this is the confusing bit: postgres has two different methods of serialziation

1. `varlena` types must exist as a flat memory buffer, and are written to disk in the same format as they exist in memory.
2. If you want to use an aggregate in parallel execution, you need to be able to serialize the transition state into some (machine-agnostic?) wire format for IPC. Notably, we used to store the output of this in continuous aggregates, though that's slated to change in timescaledb 2.7.0

@WireBaron is saying that these types aren't the first kind; they only exist in transition functions–ephemerally in memory–and therefore they can change between version. The test is checking that we can serialize this to disk stably in continuous aggregates, which is a _slightly_ weaker invariant; we can change the in-memory representation freely as long as we can keep deserializing old partials. Hopefully we'll be able to get everyone onto the new continuous aggregates format soon, and we'll be able to stop caring about the second form of stability.

JLockerman: Notably this means that the following comment
```rust
//! Note that [MetricSummary] is already in violation, as it does not lock in
//! a memory representation and the Rust project makes no guarantees to
//! preserve this across releases of the compiler.  We should bump its
//! serialization version and repr(C) the new one.
```
is incorrect: the rust-memory layout doesn't really matter here, what matters is that serde guarantees layout-stability for the serialized form of a given struct definition.

epgts: (the incorrect leap i was making was from "we directly derive the serialization form from the in-memory layout" (true) to "we serialize the actual in-memory layout" (false); the latter necessarily leads to "we need repr(C)", it's just not so! :)

Josh points out on a call that what I'm doing here is backwards: adding a second struct from which to derive (de)serialize code from when instead I should stop deriving and implement them directly. Which I'll do when I return to this, likely next week.

Thanks!

PR #404, opened 2022-05-02
Refactor TimeVector to greatly simplify structure

### LINKED ISSUE #384: Refactor and simplify TimeVector
Right now the time vector object is in a complex state where it's has many half realized ideas and premature optimizations baked into it.  It should be possible to reduce it to a much simpler object which is easier to develop around and move toward stability.

The following are some suggested subtasks for this:

- [ ] Refactor time-series crate (hopefully remove)
  - [ ] Remove any timevector objects here, they're superseded by the objects in timeseries.md
  - [ ] move gapfill objects and code needed by asap into asap.md (under extension/src)
  - [ ] if it's worth keeping the crate just for the remaining tspoint struct, rename it at least, but try to find a more elegant solution
- [ ] Rename any timeseries vestiges in comments or filenames to timevector
- [ ] Remove the enum variants of the Timevector object, leave just the explicit format with a flag to indicate sorted (we may want to add the enum back to handler another representation in the future, but this can be easily handled with our versioning and isn't worth carrying the complexity until then)

WireBaron:
This changes the TimeVector to remove the SeriesType enum which was adding a lot of complexity, but was not providing any added benefit (as all TimeVectors were being serialized as Explicit type anyways).

This change also finishes up renaming a few remaining `time_series` files and directories and removes the dead code from the time_series crate (the TSPoint code is moved to its own crate).

Due to the large amount of renaming, it may be easier for reviewers to look at this PR change by change rather than as a large single change.

Fixes #384


crates/tspoint/src/lib.rs
```diff
@@ -0,0 +1,161 @@
...
+        S: serde::Serializer {
+        if serializer.is_human_readable() {
+            // FIXME ugly hack to use postgres functions in an non-postgres library
+            extern "C" {
+                fn _ts_toolkit_encode_timestamptz(dt: i64, buf: &mut [u8; 128]);
+            }
+            let mut ts = [0; 128];
+            unsafe {
+                _ts_toolkit_encode_timestamptz(self.ts, &mut ts);
+            }
```
JLockerman, line 44: (This code is _entirely_ my fault, and this comment is meant to cast aspersions on me, not Brian's wonderful PR)
I just realized we do might be able to do this more sanely by dynamically by storing a global function pointer (well, `AtomicUsize` but same diff), and setting it lazily from somewhere. This would also allow us to to have a default version of this function for testing outside of Postgres. It may pay for us to do this in a followup PR.

extension/src/time_vector/pipeline.rs
```diff
@@ -73,14 +61,8 @@ pub mod toolkit_experimental {
...
-            ResampleToRate: 2 {
-                interval: i64,
-                resample_method: ResampleMethod,
-                snap_to_rate: i64, // padded bool
-            },
-            FillHoles: 3 {
-                fill_method: FillHolesMethod,
-            },
+            // 2 was for resample_to_rate
+            // 3 was for fill_holes
```
JLockerman, line 65: Note that we can replace these if we so choose; the update script will make sure that the gaps aren't used. (We don't need to now, but we should probably do so for when we add new elements)

extension/src/time_vector/pipeline.rs
```diff
@@ -477,7 +455,7 @@ mod tests {
                 (ts:\"2020-04-10 00:00:00+00\",val:5.8046),\
                 (ts:\"2020-04-14 00:00:00+00\",val:7.195),\
                 (ts:\"2020-04-20 00:00:00+00\",val:10.0221)\
-            ]");
+            ],is_sorted:true)");
```
JLockerman, line 458: This is interesting: I was deliberate when implementing the timevector text format so that it would _only_ contain the data, and not the metadata such as the `num_points` or `is_sorted`. I did this because this seemed like one of the few types where working with the text form of the type could be useful, so keeping only the data minimized the possibility of data validation errors. That said,I guess none of this matters yet, so we can punt on this until later a long a we record the decision.

JLockerman (approved):
Oh, wow. I'm really impressed with how much simpler this makes things, and pleasantly surprised how readable the PR is. I left some comments, but they're really there "for the record", mostly containing Gratuitous Style Nits than anything I think is necessarily worth changing before the PR is merged.

crates/tspoint/src/lib.rs
WireBaron: Sounds promising, but we'll leave this for a future change.

extension/src/asap.rs
```diff
@@ -135,23 +135,24 @@ fn asap_final_inner(
...
 
             // Drop the last value to match the reference implementation
             normal.pop();
             let values = asap_smooth(&normal, state.resolution as u32);
 
+            let mut points = Vec::new();
+            let interval = downsample_interval * normal.len() as i64 / values.len() as i64;
+            for i in 0..values.len() {
+                points.push(TSPoint{ts: start_ts + i as i64 * interval, val: values[i]});
+            }
```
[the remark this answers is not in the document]
WireBaron: This looks good, made the change

extension/src/asap.rs
```diff
@@ -160,61 +161,58 @@ fn asap_final_inner(
...
     // Drop the last value to match the reference implementation
     normal.pop();
 
     let result = asap_smooth(&normal, resolution as u32);
 
+    let mut points = Vec::new();
+    let interval = downsample_interval * normal.len() as i64 / result.len() as i64;
+    for i in 0..result.len() {
+        points.push(TSPoint{ts: start_ts + i as i64 * interval, val: result[i]});
+    }
```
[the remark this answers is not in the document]
WireBaron: Adopted same change

extension/src/asap.rs
```diff
@@ -160,61 +161,58 @@ fn asap_final_inner(
...
-            }
+            num_points: points.len() as u64,
+            points: points.into(), 
+            is_sorted: true,
         }
     })
 }
 
+// Adds the given number of points to the end of a non-empty NormalTimevector
+pub fn fill_normalized_series_gap(values: &mut Vec<f64>, points: i32, post_gap_val: f64) {
```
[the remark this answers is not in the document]
WireBaron: No, this definitely should not be `pub` now that it's been moved to this file.

extension/src/time_vector.rs
```diff
@@ -0,0 +1,297 @@
...
+#[pg_schema]
+pub mod toolkit_experimental {
+    use super::*;
+    pg_type! {
+        #[derive(Debug)]
+        struct Timevector<'input> {
+            num_points: u64,  // required to be aligned
+            points: [TSPoint; self.num_points],
+            is_sorted: bool,
+        }
```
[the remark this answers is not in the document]
WireBaron: Seemed like an easy enough change, though it did end up changing all the test output.

extension/src/time_vector/pipeline.rs
WireBaron: I think we'll ultimately want to bring back resample and fill_holes once we plan them out better, though of course we can fill these gaps with anything

extension/src/time_vector/pipeline.rs
WireBaron: I still think it makes sense to call `unnest` to get a readable form.  Even without the metadata the points aren't that clean.

crates/tspoint/src/lib.rs
```diff
@@ -0,0 +1,161 @@
+
```
epgts, line 1: How about rustfmt the added files?

extension/src/time_vector/iter.rs
```diff
@@ -0,0 +1,36 @@
...
+    fn next(&mut self) -> Option<Self::Item> {
+        match self {
+            Slice{iter} => iter.next(),
+        }
+    }
+
+    // XXX the functions below, `last()` and `count()` in particular rely on
+    //     this being precise and accurate, with both elements of the tuple
+    //     being the same as the actual yielded number of elements, if this
+    //     changes those will also nee to change
```
epgts, line 23: "need"

extension/src/time_vector/pipeline/map.rs
```diff
@@ -152,11 +102,11 @@ pub fn check_user_function_type(function: pg_sys::regproc) {
         error!("invalid number of mapping function arguments, expected fn(timevector) RETURNS timevector")
     }
 
-    if unsafe { *argtypes } != *crate::time_series::TIMEVECTOR_OID {
+    if unsafe { *argtypes } != *crate::time_vector::TIMEVECTOR_OID {
```
epgts, line 105: Can we `assert!(!nargs.is_null())` before dereferencing? I'd always NULL-check in C too. I prefer surprising "impossible NULL at file:NNN" in my logs to surprise segfaults :)

docs/timeseries_pipeline_elements.md
```diff
@@ -51,16 +51,14 @@ CREATE VIEW daily_delta AS
...
             ->  toolkit_experimental.delta()) AS deltas
     FROM test_data
     GROUP BY device;
 ```
 
-This command creates a timevector from the time and temperature columns (grouped by device), sorts them in increasing time, aggregates them as a daily average, interpolates the values for any missing days, and computes the deltas between days.  Now we can look at the deltas for a specific device:
+This command creates a timevector from the time and temperature columns (grouped by device), sorts them in increasing time, and computes the deltas between values.  Now we can look at the deltas for a specific device.  Note that the output for this test is inaccurate as we've removed some of the pipeline elements for the moment.
 
-```SQL
+```SQL,ignore-output
```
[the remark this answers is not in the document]
WireBaron: In this particular case, the problem is that without the `resample_to_rate` operator the output jumps from 30 lines to 10000.  I would like to see this `ignore-output` tag removed once we come up with a better designed resampling operator.

In general we mark sql as `ignore` if we're just using it as documentation .  For instance if I just wanted to illustrate how to use an API without actually setting up the underlying tables and data that I'd need to actually run it.  This was obviously a lot more common when the primary purpose of the files in this directory was to serve as our primary documentation.

extension/src/time_vector/iter.rs
WireBaron: fixed

extension/src/time_vector/pipeline/map.rs
WireBaron: I assume you mean `argtypes` rather than `nargs` here.  Added the assert.

crates/tspoint/src/lib.rs
WireBaron: That's actually a fair point.  There are no truly new files in this PR, but anything which has been sufficiently mangled after renaming will appear new once I squash it, so it's a good time to apply formatting.

PR #410, opened 2022-05-05
Include both pgx 0.2.x and pgx 0.4.x in CI image and modify update-tester crate

rtwalker:
We're updating our pgx dependency to 0.4.3, and using `cargo-pgx install` with pgx 0.4.x inside the update-tester causes problems.

Starting in 0.4, pgx changed it's SQL generation and gives the following error when running `cargo pgx install` inside the update tester with previous versions of the toolkit checked out:
```
  found `pgx` 0.2-0.3 series SQL generation while using `cargo-pgx` 0.4 series.
```
As a solution, we're changing the docker image used in our CI setup to include 0.2 series version of pgx installed at /pgx-0.2 as well as a 0.4 series version of pgx installed at /pgx-0.4.

With both pgx versions installed, for each toolkit release that is checked out while running the update tester, we can infer the appropriate cargo pgx subcommand to use by inspecting the Cargo.toml.

### DIFF
--- a/docker/ci/Dockerfile
+++ b/docker/ci/Dockerfile
@@ -4,13 +4,18 @@ RUN apt-get update \
     && apt-get install -y clang libclang1 sudo bash cmake \
     && rm -rf /var/lib/apt/lists/*
 
+# install cargo pgx
+RUN cargo install cargo-pgx --version '^0.2' --root /pgx/0.2 \
+    && cargo install cargo-pgx --version '^0.4' --root /pgx/0.4
+
+ENV PATH "/pgx/0.2/bin/:${PATH}"
+
+# install doctester
+RUN cargo install --git https://github.com/timescale/timescaledb-toolkit.git --branch main sql-doctester
+
 RUN useradd -ms /bin/bash postgres
 USER postgres
 
-# install cargo pgx
-RUN cargo install cargo-pgx --version '^0.2'
-
-# only use pg12 for now timescaledb doesn't support 13
 RUN set -ex \
     && cargo pgx init --pg12 download --pg13 download --pg14 download \
     && cargo pgx start pg12 \
@@ -19,6 +24,7 @@ RUN set -ex \
     && cargo pgx stop  pg13 \
     && cargo pgx start pg14 \
     && cargo pgx stop  pg14
+
 
 # install timescaledb
 # TODO make seperate image from ^
@@ -49,9 +55,6 @@ RUN set -ex \
         && echo "shared_preload_libraries = 'timescaledb'" >> ~/.pgx/data-14/postgresql.conf \
     && cd ~ \
     && rm -rf ~/timescaledb
-
-# install doctester
-RUN cargo install --git https://github.com/timescale/timescaledb-toolkit.git --branch main sql-doctester
 
 # add clippy
 RUN rustup component add clippy
--- a/tools/update-tester/src/installer.rs
+++ b/tools/update-tester/src/installer.rs
@@ -1,27 +1,30 @@
-use std::{
-    collections::HashSet,
-    path::Path,
-};
+use std::{collections::HashSet, path::Path};
 
 use colored::Colorize;
-
+use semver::Version;
+use toml_edit::Document;
 use xshell::{cmd, cp, mkdir_p, pushd, pushenv, read_dir};
 
 use crate::{defer, quietly_run};
 
+#[allow(clippy::too_many_arguments)]
 pub fn install_all_versions(
     root_dir: &str,
     cache_dir: Option<&str>,
     pg_config: &str,
+    cargo_pgx: &str,
+    cargo_pgx_old: &str,
     current_version: &str,
     old_versions: &[String],
     reinstall: &HashSet<&str>,
 ) -> xshell::Result<()> {
     let extension_dir = path!(root_dir / "extension");
-    let install_toolkit = || -> xshell::Result<()> {
-        let _d = pushd(&extension_dir)?;
+    let install_toolkit = |pgx_version: Version| -> xshell::Result<()> {
         let _e = pushenv("CARGO_TARGET_DIR", "../target/extension");
-        quietly_run(cmd!("cargo pgx install -c {pg_config}"))
+        match pgx_version >= Version::new(0, 4, 0) {
+            true => quietly_run(cmd!("{cargo_pgx} pgx install -c {pg_config}")),
+            false => quietly_run(cmd!("{cargo_pgx_old} pgx install -c {pg_config}")),
+        }
     };
     let post_install = || -> xshell::Result<()> {
         let _d = pushd(root_dir)?;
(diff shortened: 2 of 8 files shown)


tools/update-tester/src/installer.rs
```diff
@@ -77,6 +90,36 @@ fn get_current_checkout() -> xshell::Result<String> {
...
 }
 
+fn get_pgx_version(cargo_toml_contents: &str) -> String {
+    let mut version = get_field_val(cargo_toml_contents, "pgx").to_string();
+    version.truncate(3); // only care about the major.minor version
+    version
+}
+
+// find a `<field name> = "<field value>"` and extract `<field value>`
+fn get_field_val<'a>(contents: &'a str, field_name: &str) -> &'a str {
```
rtwalker, line 100: this is a duplicate of [`tools/post-install/src/main.rs:get_field_val`](https://github.com/timescale/timescaledb-toolkit/blob/8183d126e4898d64892ee358b3b48b8851f7abbf/tools/post-install/src/main.rs#L265-L271)

let me know if I should just use that function instead.

tools/update-tester/src/installer.rs
```diff
@@ -77,6 +90,36 @@ fn get_current_checkout() -> xshell::Result<String> {
...
+    contents
+        .lines()
+        .filter(|line| line.starts_with(field_name))
+        .map(get_quoted_field)
+        .next()
+        .unwrap_or_else(|| panic!("cannot read field `{}` in file", field_name))
+}
+
+// given a `<field name> = "<field value>"` extract `<field value>`
+fn get_quoted_field(line: &str) -> &str {
```
rtwalker, line 110: this is a near duplicate of [`tools/post-install/src/main.rs:get_quoted_field`](https://github.com/timescale/timescaledb-toolkit/blob/8183d126e4898d64892ee358b3b48b8851f7abbf/tools/post-install/src/main.rs#L274-L283)

Uses `"` instead of `'`, but if I'm going into the `post-install` crate for the other function, I might just change the function there to work for `"` too.

tools/update-tester/src/installer.rs
```diff
@@ -18,10 +18,11 @@ pub fn install_all_versions(
     reinstall: &HashSet<&str>,
 ) -> xshell::Result<()> {
     let extension_dir = path!(root_dir / "extension");
-    let install_toolkit = || -> xshell::Result<()> {
-        let _d = pushd(&extension_dir)?;
+    let install_toolkit = |pgx_version| -> xshell::Result<()> {
         let _e = pushenv("CARGO_TARGET_DIR", "../target/extension");
-        quietly_run(cmd!("cargo pgx install -c {pg_config}"))
+        quietly_run(cmd!(
+            "/pgx-{pgx_version}/bin/cargo-pgx pgx install -c {pg_config}"
```
rtwalker, line 24: A downside of this approach is likely not being able to run the update tester locally before pushing. I'm not sure whether that's currently part of anyone's workflow, but it would be broken until you installed `pgx` into those same locations on your machine.

rtwalker:
Is there a way to determine whether the changes to `docker/ci/Dockerfile` triggered a new build of the CI image? Hoping the failure is due to a new image incorporating the changes not being built first.

tools/update-tester/src/installer.rs
JLockerman: you can switch it to the ones in [`scripting-utilities`](https://github.com/timescale/timescaledb-toolkit/blob/8183d126e4898d64892ee358b3b48b8851f7abbf/crates/scripting-utilities/control_file_reader/src/lib.rs)

(background: I really wanted `post-install` install to be fast to compile and run b/c it's part of the normal workflow, so when I started writing other scripts that used similar logic I just copy/pasted the code. However _this_ code is small enough and important enough that I figure it's worth having one crate that defines the canonical layout. However, I never had the chance to go back and change the others 😓 )

tools/update-tester/src/installer.rs
JLockerman: 🤔

JLockerman:
> Is there a way to determine whether the changes to docker/ci/Dockerfile triggered a new build of the CI image?

Right now that image is only built manually, so it definitely didn't change the build 😅.
We should probably add a rule to automatically build the image whenever we change the docker file (like [we do](https://github.com/timescale/timescaledb-toolkit/blob/8183d126e4898d64892ee358b3b48b8851f7abbf/.github/workflows/patch_build.yml#L4-L7) for the patch image!) but I never figured out how to run the other tests on the newly built image.

tools/update-tester/src/installer.rs
JLockerman: That used to be part of my workflow on occasion when I needed to debug update issues, but really that only happened when debugging the update scripts or the meta-tests, which honestly wasn't all that frequent.

I think the way to support local builds would be to let the user pass in the root dir in which the `pgx-{pgx_version}`s reside. But I think we can leave that for a followup PR if @WireBaron and @epgts find it acceptable (maybe the next person to break updates should fix it 😈)

tools/update-tester/src/installer.rs
JLockerman: Oh, wait, these aren't actually identical, the other one looks for things in  single quotes `'`, while this one looks for things in double quotes `"`.

JLockerman: I have mixed feelings about this; on the one hand this is nice and simple and should work now. On the other hand the TOML syntax is much more involved than the `.control` one, and this fits us to one specific version. On the gripping hand, this version auto-discovery won't work for the fancier cargo features like git dependencies anyway, so whatever we choose we'll probably have to change if the code if we do that. 

In other words: you can use a real toml library if you want; this is always going to compile the current version of Toolkit, so the additional overhead shouldn't be that high, but don't feel like you have to. If you do decide to go that route, [`toml_edit`](https://docs.rs/toml_edit/latest/toml_edit/index.html) seems to be the one cargo is using right now, so that should be a safe choice.

tools/update-tester/src/installer.rs
JLockerman: Oh, nevermind, I was wrong see https://github.com/timescale/timescaledb-toolkit/pull/410#discussion_r866934570

JLockerman:
Anyway, I'm 👍 on this, but I feel like I shouldn't be the sole reviewer, so I'm going to give @WireBaron and @epgts a chance to review.

docker/ci/Dockerfile
```diff
@@ -8,17 +8,17 @@ RUN useradd -ms /bin/bash postgres
 USER postgres
 
 # install cargo pgx
-RUN cargo install cargo-pgx --version '^0.2'
+RUN cargo install cargo-pgx --version '^0.2' --root /pgx-0.2 \
+    && cargo install cargo-pgx --version '^0.4' --root /pgx-0.4
```
JLockerman, line 12: note that [the CI scripts](https://github.com/timescale/timescaledb-toolkit/blob/8183d126e4898d64892ee358b3b48b8851f7abbf/.github/workflows/ci.yml) (well [here](https://github.com/timescale/timescaledb-toolkit/blob/main/tools/build)) will look for `cargo pgx` in the path, so we need to add at least one of these to the path to ensure they keep working. My reccomendation would be to add the `0.2` one then switch it in the same PR we update the extension to `0.4`

JLockerman: (I don't love how many layers of indirection CI has, but I don't know of any good way to fix that 🤔)

tools/update-tester/src/installer.rs
rtwalker: Sounds like we are on the same page. I was a little torn about hand-rolling my own `Cargo.toml` inspection but was also unsure about introducing a dependency for what seemed like a straightforward enough task (for now at least). 

`toml_edit` looks like good choice to me!

docker/ci/Dockerfile
rtwalker: Good catch! Will update using your recommendation

tools/update-tester/src/installer.rs
epgts: I made running all the tests easy with `tools/build` specifically so that we COULD run them all early and often! And I do :)

That said, I don't think installing a couple versions of pgx into a special location is so onerous (assuming that is in fact all that's necessary; I had other problems, see below). Just please document that in the README.

Thanks!

epgts (changes requested):

docker/ci/Dockerfile
```diff
@@ -8,17 +8,17 @@ RUN useradd -ms /bin/bash postgres
 USER postgres
 
 # install cargo pgx
-RUN cargo install cargo-pgx --version '^0.2'
+RUN cargo install cargo-pgx --version '^0.2' --root /pgx-0.2 \
```
epgts, line 11: Let's use `/pbx/0.2` and `/pgx/0.4` instead (both children of `/pgx`).

tools/update-tester/src/installer.rs
epgts: I see what's wrong; have a look at commit dbd10eb4357ffd6e039ce798d5a8824b7faa1956 .

Still would be nice to fix that error handling...

rtwalker: I'd been unaware/forgotten about `tools/build`! I think rather than breaking it, I'd like to see it keep working and start using it myself 😜

I'm open to other solutions still, but assuming `pgx`s are installed at `/pgx/0.X` is the best I can come up with at the moment, so good to know it isn't too onerous. Duly noted about making others aware in the README.

From at your commit:
> Also we don't need to read and parse Cargo.toml repeatedly.
 
Don't we? Is there another (better) way to know which `cargo pgx` we want to call? Where the `pgx` we want will depend on which release we have checked out

epgts: Hah, of course you're right, we do! My bad. I think something like the scoping changes are required though.

rtwalker:
Closing this PR and splitting the changes into two. Seems like the "CI / Test Updates" job runs before the "Build CI Image" job and so, the updates test (which is built with the new CI image in mind) fails. 

It looks like there might be a way to specify workflow dependencies and workflows that trigger other workflows but the examples in the GitHub workflows docs look frightening to me. I'd rather just do the simple thing.

rtwalker:
Closing this PR and splitting the changes into two. Seems like the "CI / Test Updates" job runs before the "Build CI Image" job and so, the updates test (which is built with the new CI image in mind) fails.

It looks like there might be a way to specify workflow dependencies and workflows that trigger other workflows but the examples in the GitHub workflows docs look frightening to me. I'd rather just do the simple thing.

PR #411, opened 2022-05-06
Create reportpackagingfailures.yml

jerryxwu:
Add a Github Action to report on "Build packages" Action run failures (failure, canceled, etc). The initial version also report on CI run failures for testing purpose since we don't build packages very often.

### DIFF
--- a/.github/workflows/reportpackagingfailures.yml
+++ b/.github/workflows/reportpackagingfailures.yml
@@ -40,4 +40,3 @@ jobs:
         env:
           SLACK_WEBHOOK_URL: ${{ secrets.SLACK_NOTIFY_WEBHOOK_URL }}
           SLACK_WEBHOOK_TYPE: INCOMING_WEBHOOK
-Loading
--- a/.github/workflows/reportpackagingfailures.yml
+++ b/.github/workflows/reportpackagingfailures.yml
@@ -1,0 +1,43 @@
+name: Report Build Package Failures
+on:
+  workflow_run:
+    workflows: [Build packages, CI]
+    types: [completed]
+
+jobs:
+  on-failure:
+    runs-on: ubuntu-latest
+    if: ${{ github.event.workflow_run.conclusion != 'success' }}
+    steps:
+      - name: slack-send
+        uses: slackapi/slack-github-action@v1.19.0
+        with:
+          payload: |
+            {
+              "blocks": [
+                {
+                  "type": "section",
+                  "text": {
+                    "type": "mrkdwn",
+                    "text": "Workflow run <${{ github.event.workflow_run.html_url }}|${{ github.event.workflow.name}}#${{ github.event.workflow_run.run_number }}>"
+                  }
+                },
+                {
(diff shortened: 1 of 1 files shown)


.github/workflows/reportpackagingfailures.yml
```diff
@@ -0,0 +1,43 @@
...
+                      "text": "*Triggered By*\n<${{ github.event.sender.html_url }}|${{ github.event.sender.login }}>"
+                    }
+                  ]
+                }
+              ]
+            }
+        env:
+          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_NOTIFY_WEBHOOK_URL }}
+          SLACK_WEBHOOK_TYPE: INCOMING_WEBHOOK
+Loading
```
rtwalker, line 43: I'm not super familiar with GitHub Actions, does this line do anything?

JLockerman: it [looks like](https://github.com/timescale/timescaledb-toolkit/actions/runs/2278995924) it's an error: CC @jerryxwu

.github/workflows/reportpackagingfailures.yml
```diff
@@ -0,0 +1,43 @@
+name: Report Build Package Failures
+on:
+  workflow_run:
+    workflows: [Build packages, CI]
```
JLockerman, line 4: I'm not convinced we should report every time CI fails: it's pretty common for it fail on PRs, and I think notifying each time will add too much noise to the channel. It might be better to only send the notification if it failed when run against the `main` or `staging` branches; the former means our nightly checks failed, while the latter means something failed in Bors

JLockerman: using `${{ github.event.workflow_run.conclusion != 'success'  && github.event.workflow_run. event != 'pull_request'}}` in the `if:` on line `10` might do it 🤔

jerryxwu: @JLockerman The intention is to remove CI failures reporting later. Keeping it there initially so we see how this works for us since we don't build packages very often.

JLockerman (approved):
One other comment (https://github.com/timescale/timescaledb-toolkit/pull/411/files#r866971745) then I think I'm 👍 (by which I mean: I don't understand the slack part of this at all, but the rest looks good 😋)

.github/workflows/reportpackagingfailures.yml
JLockerman: @jerryxwu That makes sense, 👍 sound like a plan

rtwalker (approved):

.github/workflows/reportpackagingfailures.yml
jerryxwu: > using `${{ github.event.workflow_run.conclusion != 'success' && github.event.workflow_run.event != 'pull_request'}}` in the `if:` on line `10` might do it 🤔

Good suggestion. I will make the change to reduce noise.

PR #412, opened 2022-05-06
Allow building CI image in Actions

JLockerman:
This commit adds a GitHub action to build and deploy our CI image. This will any of us to rebuild and deploy the CI image, and test it better before deployment.

### DIFF
--- a/.github/workflows/ci_image_build.yml
+++ b/.github/workflows/ci_image_build.yml
@@ -1,0 +1,56 @@
+name: Build CI Image
+
+on:
+  pull_request:
+    paths:
+      - 'docker/ci/**'
+      - '.github/workflows/ci_image_build.yml'
+  workflow_dispatch:
+
+jobs:
+  build:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout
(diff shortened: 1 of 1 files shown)


.github/workflows/ci_image_build.yml
```diff
@@ -0,0 +1,56 @@
...
+on:
+  pull_request:
+    paths:
+      - 'docker/ci/**'
+      - '.github/workflows/ci_image_build.yml'
+  workflow_dispatch:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
```
jerryxwu, line 12: @JLockerman Is this the place that we should be explicit about version of ubuntu that we want to stick to rather than the latest?

JLockerman: @jerryxwu 🤔 this is only ever used to build the docker image, so I'm not sure how much it matters either way

JLockerman: @jerryxwu I'll pin `20.04` under the assumption that it's slightly more likely to keep working that way

rtwalker (approved):

PR #414, opened 2022-05-06
Specify Ubuntu 20.04 instead of 'latest' in github configuration.

epgts:
'ubuntu-latest' is currently documented as being 20.04:
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idruns-on

But let's lock that in now rather than surrender our update schedule
to GitHub.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ on:
 jobs:
   testpostgres:
     name: Test Postgres
-    runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


epgts:
Let's wait for @jerryxwu 's #413 rename of this file to merge first.

rtwalker (approved):

epgts:
@WireBaron this is ready to go now.

WireBaron (approved):

PR #417, opened 2022-05-11
Include both pgx 0.2.x and pgx 0.4.x in CI image

rtwalker:
We're updating our pgx dependency to 0.4 series, and using `cargo-pgx install` with pgx 0.4.x inside the update-tester causes problems.

Starting in 0.4, pgx changed it's SQL generation and gives the following error when running `cargo pgx install` inside the update tester with previous versions of the toolkit checked out:

  found `pgx` 0.2-0.3 series SQL generation while using `cargo-pgx` 0.4 series.

As a solution, we're changing the docker image used in our CI setup to include 0.2 series version of pgx installed at /pgx/0.2 as well as a 0.4 series version of pgx installed at /pgx/0.4 so that we can run the update tester with the appropriate `cargo-pgx` subcommand.

### DIFF
--- a/docker/ci/Dockerfile
+++ b/docker/ci/Dockerfile
@@ -4,13 +4,18 @@ RUN apt-get update \
     && apt-get install -y clang libclang1 sudo bash cmake \
     && rm -rf /var/lib/apt/lists/*
 
+# install cargo pgx
+RUN cargo install cargo-pgx --version '^0.2' --root /pgx/0.2 \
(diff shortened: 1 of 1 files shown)


JLockerman (approved):
Looks right to me 👍 
(and the CI builder test [seems to be running](https://github.com/timescale/timescaledb-toolkit/runs/6394421072?check_suite_focus=true) 😋)

PR #408, opened 2022-05-03
Upgrade to pgx 0.4.3

rtwalker:
### Upgrading to the latest pgx
Closes #387 

Went from 0.2.4 -> 0.3.0 -> 0.4.0 -> 0.4.3, editing the Cargo.toml, building, and testing, and the experience was largely unremarkable.

Most of the changes come from followng the "Upgrading" section of the [0.4.0](https://github.com/tcdi/pgx/releases/tag/v0.4.0) release notes.

Only thing to note is that I ran into a *lot* of occurrences of the following error:
```
 error[E0053]: method `input` has an incompatible type for trait
    --> extension/src/type_builder.rs:261:29
     |
 261 |             fn input(input: &std::ffi::CStr) -> $name<'input>
     |                             ^^^^^^^^^^^^^^^
     |                             |
     |                             expected struct `pgx::cstr_core::CStr`, found struct `std::ffi::CStr`
     |                             help: change the parameter type to match the trait: `&pgx::cstr_core::CStr`
```

However, just blindly swapping in the `pgx::cstr_core` `CStr(ing)`s for the `std::ffi` ones made Rust much happier. 
Please yell at me if that's too irresponsible.

### macOS support on Apple M1 Macs
I think we should be OK to close #401 as well. 

`pgx` claims M1 support, and by virtue of being able to do 
```
$ cargo pgx run pg14
    Stopping Postgres v14
building extension with features `pg14`
"cargo" "build" "--features" "pg14" "--no-default-features" "--message-format=json-render-diagnostics"
    Finished dev [unoptimized + debuginfo] target(s) in 0.07s

installing extension
     Copying control file to /Users/rwalker/.pgx/14.2/pgx-install/share/postgresql/extension/timescaledb_toolkit.control
     Copying shared library to /Users/rwalker/.pgx/14.2/pgx-install/lib/postgresql/timescaledb_toolkit.so
 Discovering SQL entities
  Discovered 655 SQL entities: 9 schemas (1 unique), 521 functions, 71 types, 0 enums, 54 sqls, 0 ords, 0 hashes, 0 aggregates
     Writing SQL entities to /Users/rwalker/.pgx/14.2/pgx-install/share/postgresql/extension/timescaledb_toolkit--1.7.0-dev.sql
    Finished installing timescaledb_toolkit
    Starting Postgres v14 on port 28814
    Re-using existing database timescaledb_toolkit
psql (14.2)
Type "help" for help.

timescaledb_toolkit=# 
```
and
```
[...]

PR #418, opened 2022-05-11
`update-tester` requires two versions of `cargo-pgx`

rtwalker:
Now that #417 is merged, we can change the `update-tester` crate to require two versions of `pgx` to be installed: 
 * a 0.2-0.3 series pgx aka "old"
 * a 0.4 series or newer pgx (we don't actually use this one yet, but we will after #408)

Rather than insisting on having the `pgx`s installed in a specific location, the update tester takes in paths to the `*/bin/cargo-pgx` subcommands as arguments and then inspects the `Cargo.toml` during the checkout of each release to determine which one to invoke. 

I've also updated the `tools/build` script and the `.github/workflows/ci.yml` to use the new update tester.

Went ahead and bumped the `update-tester` version to `0.2.0` so I could more easily tell the difference while testing locally, though it probably makes little practical difference.

### DIFF
--- a/tools/build
+++ b/tools/build
@@ -16,7 +16,7 @@ die() {
 }
 
 usage() {
-    die 'build [ -n -pg1[234] ] ( test-crates | test-extension | test-post-install | test-doc | clippy )'
+    die 'build [ -n -pg1[234] ] ( test-crates | test-extension | test-post-install | test-doc | test-updates | clippy)'
 }
 
 require_pg_version() {
@@ -37,6 +37,15 @@ find_pg_config() {
     fi
     [ -x "$pg_config" ] || die "$pg_config not executable"
 }
+
+require_cargo_pgx() {
+    [ -n "$cargo_pgx" ] || die 'specify path to cargo-pgx (0.4 series or newer)'
+}
+
+require_cargo_pgx_old() {
+    [ -n "$cargo_pgx_old" ] || die 'specify path to cargo-pgx (0.2-0.3 series)'
+}
+
 
 [ $# -ge 1 ] || usage
 
@@ -74,6 +83,16 @@ while [ $# -gt 0 ]; do
 
         -pgconfig)
             pg_config="$1"
+            shift
+            ;;
+
+        -cargo-pgx)
+            cargo_pgx="$1"
+            shift
+            ;;
+
+        -cargo-pgx-old)
+            cargo_pgx_old="$1"
             shift
             ;;
 
@@ -139,6 +158,8 @@ while [ $# -gt 0 ]; do
         test-updates)
             require_pg_version
             find_pg_config
+            require_cargo_pgx
+            require_cargo_pgx_old
             (
                 export CARGO_TARGET_DIR="$top_CARGO_TARGET_DIR"
                 $nop cargo pgx start $pg
@@ -147,7 +168,9 @@ while [ $# -gt 0 ]; do
                  -p $pg_port \
                  --cache old-versions \
                  . \
-                 "$pg_config"
+                 "$pg_config" \
+                 "$cargo_pgx" \
+                 "$cargo_pgx_old"
             )
             ;;
 
--- a/tools/update-tester/src/installer.rs
+++ b/tools/update-tester/src/installer.rs
@@ -1,27 +1,31 @@
-use std::{
-    collections::HashSet,
-    path::Path,
-};
+use std::{collections::HashSet, path::Path};
 
 use colored::Colorize;
-
+use semver::Version;
+use toml_edit::Document;
 use xshell::{cmd, cp, mkdir_p, pushd, pushenv, read_dir};
 
 use crate::{defer, quietly_run};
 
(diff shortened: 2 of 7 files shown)


tools/update-tester/src/installer.rs
```diff
@@ -1,27 +1,30 @@
...
 
 use colored::Colorize;
-
+use semver::Version;
+use toml_edit::Document;
 use xshell::{cmd, cp, mkdir_p, pushd, pushenv, read_dir};
 
 use crate::{defer, quietly_run};
 
+#[allow(clippy::too_many_arguments)]
```
JLockerman, line 10: 😄

JLockerman (approved):
I'd like Eric to review the `tools/build` changes but over all I think I'm 👍 on this.
There's probably a more general way to pass in multiple pgx versions, but we can probably procrastinate on that until we have more than 2.

I also left a code style comment on `get_pgx_version()`. Again, not sure if the changes are worth the effort, so only make them if you feel they help (or want to play around with that code some more 😋)

tools/build
```diff
@@ -77,6 +86,16 @@ while [ $# -gt 0 ]; do
             shift
             ;;
 
+        --cargo-pgx)
```
epgts, line 89: Either we need to change the -pg* options to use double-dash or change these two to use single. I prefer the latter, but won't put up a fight over it. As long as we're consistent.

tools/build
```diff
@@ -147,7 +168,9 @@ while [ $# -gt 0 ]; do
                  -p $pg_port \
                  --cache old-versions \
                  . \
-                 "$pg_config"
+                 "$pg_config" \
+		 "$cargo_pgx" \
+		 "$cargo_pgx_old"
```
epgts, line 173: Line these up with the above.

epgts (commented):
I want to read installer.rs and have a look at Josh's question but have run out of time. I'll look tomorrow morning.

tools/build
rtwalker: Apologies for being dense, but what does that look like?

```
                $nop cargo run --manifest-path tools/update-tester/Cargo.toml -- \
                 -h localhost \
                 -p $pg_port \
                 --cache old-versions \
                 . \
                 "$pg_config" \
		 "$cargo_pgx" \
		 "$cargo_pgx_old"
```
seems lined up to me?

tools/build
rtwalker: I don't have any taste when it comes to shell scripts. Does single-dash mean we have to do `-cargopgx` and `-cargopgxold`? Or is `-cargo-pgx-old` possible?

tools/update-tester/src/installer.rs
```diff
@@ -77,6 +94,15 @@ fn get_current_checkout() -> xshell::Result<String> {
     cmd!("git rev-parse --verify HEAD").read()
 }
 
+fn get_pgx_version(cargo_toml_contents: &str) -> Version {
+    let cargo = cargo_toml_contents
+        .parse::<Document>()
+        .expect("invalid Cargo.toml");
+    let mut version = cargo["dependencies"]["pgx"].to_string();
+    version = version.trim().trim_matches('"').to_string();
```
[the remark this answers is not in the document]
rtwalker: > In fact, the following should work as the function body

It does!

Good point about not always being able to rely on simply having
```
pgx = "X.Y.Z"
```
don't want to address now, but at least we *might* remember to fix here whenever we make that change in the `Cargo.toml`

tools/update-tester/src/installer.rs
```diff
@@ -35,32 +38,46 @@ pub fn install_all_versions(
...
-    // Install the versions in reverse-time order.
-    // Since later versions tend to be supersets of old versions,
-    // I expect compilation to be faster this way - Josh
-    for version in old_versions.iter().rev() {
-        let force_reinstall = reinstall.contains(&**version);
-        if !force_reinstall && version_is_installed(pg_config, version)? {
-            eprintln!("{} {}", "Already Installed".blue(), version);
-            continue
+    {
+        let _ext = pushd(&extension_dir)?;
```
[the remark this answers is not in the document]
epgts: Yep, this looks good. It's unfortunate that this is how pgx is. At some point we should see if we can fix or replace pgx :)

For now:
- for all crates actions, operate at project root
- for all extension actions, operate at extension sub-directory with separate target directory

And it looks like this is done right, and I verified locally.

tools/build
epgts: `-cargo-pgx-old` is fine. Feel free to adjust the others for consistency. `-pgconfig` looks fine but `-cargopgxold` is definitely too many words smooshed together :)

tools/build
```diff
@@ -139,6 +158,8 @@ while [ $# -gt 0 ]; do
         test-updates)
             require_pg_version
             find_pg_config
+	    require_cargo_pgx
+	    require_cargo_pgx_old
```
epgts, line 162: these two are still misindented

Issue #422, opened 2022-05-16
Support aarch64-unknown-linux-gnu


epgts:
We're seeing increased demand to support aarch64 for Linux platforms, in both private and public (e.g. issue #382). We plan to support a binary release (e.g. Docker image), but a first step is to make the toolkit work on aarch64-unknown-linux-gnu at all.

Acceptance criteria:
- github action installed to test every merge request under Ubuntu 20.04 / aarch64

Out of scope:
- Production of any binary release

epgts:
Current status:

Many tests crash.

I arbitrarily picked uddsketch::tests::pg_uddsketch_io_test to start with. Backtrace:

```
#0  core::fmt::Formatter::alternate () at library/core/src/fmt/mod.rs:1880
#1  core::fmt::Formatter::pad_integral () at library/core/src/fmt/mod.rs:1354
#2  0x0000fffc8116d30c in core::fmt::num::fmt_u128 () at library/core/src/fmt/num.rs:641
#3  0x0000fffc8116da60 in core::fmt::write () at library/core/src/fmt/mod.rs:1190
#4  0x0000fffc811ed5c0 in std::io::Write::write_fmt (self=0xffffe2581998, fmt=...)
    at /rustc/7737e0b5c4103216d6fd8cf941b7ab9bdbaace7c/library/std/src/io/mod.rs:1657
#5  0x0000fffc80b1f53c in <&mut ron::ser::Serializer<W> as serde::ser::Serializer>::serialize_u128 (self=0xffffe2581998, v=1)
    at /home/epg/.cargo/registry/src/github.com-1ecc6299db9ec823/ron-0.7.0/src/ser/mod.rs:409
#6  0x0000fffc80b1f40c in <&mut ron::ser::Serializer<W> as serde::ser::Serializer>::serialize_u8 (self=0xffffe2581998, v=1)
    at /home/epg/.cargo/registry/src/github.com-1ecc6299db9ec823/ron-0.7.0/src/ser/mod.rs:393
#7  0x0000fffc80b94bc8 in serde::ser::impls::<impl serde::ser::Serialize for u8>::serialize (
    self=0xffffe2581ea0 "\001)X\342\377\377\000", serializer=0xffffe2581998)
    at /home/epg/.cargo/registry/src/github.com-1ecc6299db9ec823/serde-1.0.137/src/ser/impls.rs:15
[...]

PR #424, opened 2022-05-17
Make some changes to support aarch64-unknown-linux-gnu

epgts:
- Change `lto` back to default of `false`.
  - It became "thin" in commit d18ebe6e8fdc2d91245b3dfe6d3c07431152979a .
  - Fixes segfaults on aarch64-unknown-linux-gnu
    (https://github.com/ron-rs/ron/issues/379)
- Change stats_agg::tests::pg_stats_agg_fuzz to allow more variance.
  - aarch64 varies a lot more on "covar_pop" and "covar_samp".
    We investigated a bit and found that the postgresql equivalents
    we were comparing against varied further from amd64 behavior than
    toolkit's.  Possibly differences between gcc and llvm code-gen?
- Also drop redundant default overrides from profiles in Cargo.toml .

for issue #422

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,13 +24,7 @@ members = [
 
 ]
 
-[profile.dev]
-panic = "unwind"
-lto = "thin"
-
 [profile.release]
-panic = "unwind"
-opt-level = 3
 lto = "fat"
 debug = true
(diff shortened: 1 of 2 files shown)


epgts:
dang github that was rude

PR #426, opened 2022-05-23
Build timescaledev/rust-pgx image for linux/arm64 (aarch64).

epgts:
- Raise timeout from the default of 6 hours to 12.
  aarch64 build seems to have been almost done when it got killed at 6 hours:
  it had gotten all the way to the doctester build which is almost the last
  step (and probably the last compiler run).  A later run took only 4 hours.
  Still, we're emulating a CPU; cut it some slack!
- Only setup QEMU for arm64 not a bunch of other platforms we don't use.
- Stop setting load: true .
- Stop building doctester.
  We've built it from source since commit 0b92688 .

for issue #422

### DIFF
--- a/.github/workflows/ci_image_build.yml
+++ b/.github/workflows/ci_image_build.yml
@@ -10,12 +10,19 @@ on:
 jobs:
   build:
     runs-on: ubuntu-20.04
+    # aarch64 build seems to have been almost done at the default 6 hour
+    # timeout: it had gotten all the way to the doctester build which is
+    # almost the last step (and probably the last compiler run?)
+    # We're emulating a CPU; cut it some slack:
+    timeout-minutes: 720
     steps:
       - name: Checkout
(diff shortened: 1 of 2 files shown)


epgts:
@JLockerman how can I try this out before merge? Thanks!

rtwalker:
> @JLockerman how can I try this out before merge? Thanks!

I think you just did :) 

see
```yaml
on:
  pull_request:
    paths:
      - 'docker/ci/**'
      - '.github/workflows/ci_image_build.yml'
```

If you're talking about using the image that was just built from the changes in this PR as the container in which all of our tests are run, that's an issue I've been having with #408. I've been playing around with GitHub Actions and have some ideas for how we could do that.

JLockerman:
@epgts yup, Ryan got it. The image automatically builds now in the PR checks.

PR #429, opened 2022-05-24
Support explicit NULL values in timevectors

### LINKED ISSUE #425: Support Explicit NULL values in Timevectors
Right now we simply ignore entries with a value of NULL when adding to a timevector.  While this may be fine for most use cases, it's not difficult to imagine a user case where the explicit presence of a NULL value is valuable information.

Note that this will change the behavior of the timevector for cases containing explicit NULLs, and therefore this should be fixed before timevectors are eligible for stabilization.

WireBaron:
This change modifies the timevector structure by replacing the is_sorted field with a set of bit flags covering both whether a timevector is known to be sorted and whether it may contain NULL values.  It also adds a bitvector with one bit per entry indicating whether that entry is NULL.

Fixes #425


rtwalker:
I think that the update tests failed because this branch is too far behind `main`. We've landed several changes to the `update-tester` crate in the meantime

epgts (commented):
Looks good to me but do you want to add some test cases with null values? I may have missed them; I thought I only saw existing tests without nulls updated.

Also could use a rustfmt :)

epgts (approved):

rtwalker:
Looks like you need another rebase onto `main` now that we've upgraded pgx

rtwalker (approved):

extension/src/lttb.rs
```diff
@@ -81,14 +83,14 @@ pub fn lttb_final_inner(
...
-                    internal_padding: [0; 3],
-                    points: (&*downsampled).into(),
-                }
-            ).into()
+            flatten!(Timevector {
+                num_points: downsampled.len() as u32,
+                flags: time_vector::FLAG_IS_SORTED,
+                internal_padding: [0; 3],
+                points: (&*downsampled).into(),
+                null_val: std::vec::from_elem(0_u8, (downsampled.len() + 7) / 8).into()
```
JLockerman, line 91: What _is_ `from_elem()`? I've never heard of the function before, and I can't find any refrences to it in the docs...

If it's what I think it is, it's probably more idiomatic to do
```rust
null_val: vec![0_u8; (downsampled.len() + 7) / 8].into()
```

extension/src/time_vector.rs
```diff
@@ -182,10 +212,13 @@ pub fn inner_compound_trans<'b>(
...
-                state.is_sorted = state.is_sorted
-                    && series.is_sorted
-                    && state.points.as_slice().last().unwrap().ts
-                        <= series.points.as_slice().first().unwrap().ts;
+                if state.is_sorted()
+                    && (!series.is_sorted()
+                        || state.points.as_slice().last().unwrap().ts
+                            > series.points.as_slice().first().unwrap().ts)
+                {
+                    state.flags ^= FLAG_IS_SORTED
```
JLockerman, line 220: is `^=` really better than `state.flags &= !FLAG_IS_SORTED` here?

extension/src/time_vector.rs
```diff
@@ -30,8 +33,10 @@ pub mod toolkit_experimental {
         struct Timevector<'input> {
             num_points: u32,
             is_sorted: bool,
-            internal_padding: [u8; 3],  // required to be aligned
+            flags: u8,         // extra information about the stored data
+            internal_padding: [u8; 2],  // required to be aligned
             points: [TSPoint; self.num_points],
+            null_val: [u8; (self.num_points + 7)/ 8], // bit vector, must be last element for alignment purposes
```
JLockerman, line 39: Ah, if only [`div_ceil()`](https://doc.rust-lang.org/stable/std/primitive.u64.html#method.div_ceill) was stable, then we could use that instead of doing this manually. I do wonder if we should extract this out into a type, or at least extract out `(num_points + 7)/ 8` into a function

extension/src/time_vector.rs
```diff
@@ -22,6 +22,9 @@ use crate::raw::bytea;
 
 pub use toolkit_experimental::{Timevector, TimevectorData};
 
+// Bit flags stored in Timevector flags
+const FLAG_HAS_NULLS : u8 = 0x01;
```
JLockerman: The [bitflags](https://docs.rs/bitflags/latest/bitflags/) library may be useful for this

JLockerman (approved):
I think we're going to eventually want to encapsulate the nulls bitmap a bit better, and probably clean the text output a bit, but it's good enough for now

PR #430, opened 2022-05-25
Lock pgx versions all the way.

epgts:
Even to the patch level!  Else we get:

    `pgx-0.4.4` shouldn't be used with `cargo-pgx-0.4.5`,
    please use `pgx = "~0.4.5"` in your `Cargo.toml`.


rtwalker:
Do you want to change
https://github.com/timescale/timescaledb-toolkit/blob/14209754c6a0205cec92d303d62deea09a532e7f/tools/update-tester/src/installer.rs#L103-L107
to 
```rust
    cargo["dependencies"]["pgx"]
        .as_str()
        .expect("expected pgx to only have a version")
        .trim_start_matches(&['=', '^', '~'])
        .parse()
        .expect("cannot parse pgx version")
```

rtwalker:
> > ```
> >     .trim_start_matches(&['=', '^', '~'])
> > ```
> 
> 
> Looks like it! Had to phrase it as `['=', '^', '~'].as_slice()` though; `&['a', 'b', 'c']` has type `&[char; 3]` when `&[char]` is required (the size is part of the array type).

hmm weird? https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=2292f156968ec6d31468021a853aaf5d

rtwalker (approved):

PR #432, opened 2022-05-25
remove PATH hack in `tools/build` script

rtwalker:
We also needed to change `tools/build` for #408 to pass CI, but we should remove once it's merged.

### DIFF
--- a/tools/build
+++ b/tools/build
@@ -2,9 +2,6 @@
 
 set -ex
 
-# TODO After we upgrade to 0.4 make this change in docker/ci/Dockerfile and then remove this line.
-export PATH=/home/postgres/pgx/0.4/bin:$PATH
-
 print() {
     printf '%s\n' "$*"
 }


docker/ci/Dockerfile
```diff
@@ -13,7 +13,7 @@ RUN cargo install cargo-pgx --version '=0.2.4' --root /home/postgres/pgx/0.2 \
     && cargo install cargo-pgx --version '=0.4.5' --root /home/postgres/pgx/0.4
 
 # TODO After we upgrade to 0.4 change this to 0.4 and remove the hack in tools/build .
```
epgts, line 15: drop the todo too

tools/build
```diff
@@ -2,9 +2,6 @@
 
 set -ex
 
-# TODO After we upgrade to 0.4 make this change in docker/ci/Dockerfile and then remove this line.
-export PATH=/home/postgres/pgx/0.4/bin:$PATH
```
epgts, line 6: We need to remove this in a separate pull request after this one so each commit is safe and clean on its own.

docker/ci/Dockerfile
```diff
@@ -12,8 +12,7 @@ USER postgres
 RUN cargo install cargo-pgx --version '=0.2.4' --root /home/postgres/pgx/0.2 \
     && cargo install cargo-pgx --version '=0.4.5' --root /home/postgres/pgx/0.4
 
-# TODO After we upgrade to 0.4 change this to 0.4 and remove the hack in tools/build .
-ENV PATH "/home/postgres/pgx/0.2/bin/:${PATH}"
+ENV PATH "/home/postgres/pgx/0.4/bin/:${PATH}"
```
epgts, line 15: nit: remove trailing / in `bin/` while you're here

tools/build
rtwalker: Opened #436 to make the changes separate

epgts (approved):

PR #441, opened 2022-06-06
Relax tolerance in UDDSketch merge assertions

### LINKED ISSUE #396: UDDSketch assertion at crates/udd-sketch/src/lib.rs:290:9
 - OS: Aiven's TimescaleDB provider
 - PostgreSQL version, 13.6
 - TimescaleDB Toolkit version: 1.5.1
 - Installation method: Aiven

**Describe the bug**
I hit an assert when I don't expect to when creating a UDDSketch.

**To Reproduce**
Apologies for the seemingly random test case, but here it goes.

```
CREATE TABLE "public"."timescale_bug" ("n" float8 NOT NULL);
```

Load the attached CSV file into the table. 

Run

```
SELECT
	uddsketch(100, 0.01, CAST(n AS FLOAT))
FROM
	timescale_bug
```

**Expected behavior**
Obvious

**Actual behavior**
```
Query 1 ERROR: ERROR:  assertion failed: (self.gamma.powf(1.0 / f64::powi(2.0, self.compactions as i32)) -\n     other.gamma.powf(1.0 / f64::powi(2.0, other.compactions as i32))).abs() <\n    f64::EPSILON
CONTEXT:  crates/udd-sketch/src/lib.rs:290:9
```

Fudging with the bucket count or error will get it to work, but it's random when the assert is tripped.

[timescale_bug.csv.zip](https://github.com/timescale/timescaledb-toolkit/files/8524661/timescale_bug.csv.zip)

rtwalker:
Fixes #396 by increasing the tolerance in our comparison of the calculated initial error parameters for two `UDDSketches` being merged

### DIFF
--- a/crates/udd-sketch/src/lib.rs
+++ b/crates/udd-sketch/src/lib.rs
@@ -290,7 +290,7 @@ impl UDDSketch {
         assert!(
             (self.gamma.powf(1.0 / f64::powi(2.0, self.compactions as i32)) - 
              other.gamma.powf(1.0 / f64::powi(2.0, other.compactions as i32))
-            ).abs() < f64::EPSILON
+            ).abs() < 1e-9 // f64::EPSILON too small, see issue #396
         );
         assert!(self.max_buckets == other.max_buckets);
 


rtwalker:
Tests passed locally, but perhaps I should take this to indicate that I need to reduce the size of `extension/src/test_data.rs`?
> [ERROR:  could not resize shared memory segment "/PostgreSQL.1297418116" to 133487040 bytes: No space left on device](https://github.com/timescale/timescaledb-toolkit/runs/6760024813?check_suite_focus=true#step:7:426)

crates/udd-sketch/src/lib.rs
```diff
@@ -290,7 +290,7 @@ impl UDDSketch {
         assert!(
             (self.gamma.powf(1.0 / f64::powi(2.0, self.compactions as i32)) - 
              other.gamma.powf(1.0 / f64::powi(2.0, other.compactions as i32))
-            ).abs() < f64::EPSILON
+            ).abs() < 1e-9
```
epgts, line 293: I think a little explanatory comment here would go a long way, and include the bug number.

rtwalker: e.g. `// f64::EPSILON too small, see issue #396`? 

I tried to make the commit message explanatory, since I usually rely on `git blame` for this stuff, but take your point that `1e-9` by itself doesn't give much context for this threshold.

epgts: That comment is great: succinct and pointing to more detail.

epgts (approved):

PR #442, opened 2022-06-06
We missed a place to update the cargo-pgx version.

epgts:
We really need to get a handle on this...

### DIFF
--- a/.github/workflows/patch_build.yml
+++ b/.github/workflows/patch_build.yml
@@ -10,6 +10,8 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
+    env:
+      CARGO_TARGET_DIR_NAME: target
     steps:
       - name: Checkout
         uses: actions/checkout@v2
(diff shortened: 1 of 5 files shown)


epgts:
Why don't we run `.github/workflows/patch_build.yml` on every pull request?
Currently we limit it to
```
    paths:
      - 'docker/patch/**'
      - '.github/workflows/patch_build.yml'
```

but that means we miss breaking changes like the pgx upgrade.

Any change to the inputs to `cargo pgx install` or `cargo run --manifest-path ../tools/post-install/Cargo.toml` need to execute this to detect breakage...  and that's almost everything.

rtwalker (commented):
> Apparently when running sql-doctester under patch_build, the extension
needs to be DROPped first.  Make tools/build DROP and switch to that.

From CI error message:
> thread 'main' panicked at 'could not run init script: Error { ... message: "extension \"timescaledb\" does not exist" ... }', tools/sql-doctester/src/runner.rs:102:14

Do you want `DROP EXTENSION IF EXISTS`?

rtwalker:
> We really need to get a handle on this...

Agreed. Thanks for leaving the comment hints in the appropriate places! Wonder what else we can do for this?

epgts:
Well both extensions are definitely already there :(

So I think I have only papered over a problem we need to instead
understand.

But no one needs this to work right now so I am going to file an issue and
leave it broken for now.

Thanks.

rtwalker (approved):

PR #449, opened 2022-06-14
Try to skip running Clippy if no rust files changed

JLockerman:


rtwalker:
can you add a way to set the [step outcome](https://docs.github.com/en/actions/learn-github-actions/contexts#steps-context) to "success" when we skip clippy so that bors will be happy?

rtwalker (approved):

JLockerman:
It seems to work, it cuts down Clippy time from ≈5 minutes to ≈1 minute if none of the rust files change. Is this worth the complexity?

PR #451, opened 2022-06-15
Improve error message for HyperLogLog

JLockerman:
This commit improves the error message on attempting to build a HLL with and invalid size. It may be better to move the check earlier, so that it errors even on null input, but this commit holds off on doing so as that would be a breaking change.


WireBaron (commented):
From CI, looks like pg_test still doesn't like the expected error.

WireBaron (approved):
Approved once CI passes, maybe comment out failure tests and create issue to investigate.

Issue #423, opened 2022-05-17
Saturating math functions/types


JLockerman:
## What's the functionality you would like to add ##
Functions such as `sub_saturate_to_zero(a, b)` that perform `a - b` but return 0 when `b > a` instead of going negative.

## Example Usage ##
TODO

## Open Questions ##
Which kinds of numbers are these useful on? How to handle naming such as `sub_saturate_to_zero` and `sub_saturate_to_min`? Would 2's complement wrapping functions be useful.

## Alternatives ##
Add a type (eg. `SaturatingNonNeg`) that does this form of operation for the normal math operators.

PR #454, opened 2022-06-23
Saturating Math for i32/integers

thatzopoulos:
Issue : https://github.com/timescale/timescaledb-toolkit/issues/423

This PR adds saturating math functions for addition/subtraction/multiplication for i32.

Each math operator has two functions: 

1. `*_saturate_to_min` saturates at the numeric bounds instead of overflowing, similar to Rusts builtin saturating methods.
2. `*_saturate_to_zero` saturates at 0 for the minimum bound instead of i32::MIN

The function names could probably use some work, I had a difficult time naming them in a way that wasn't too verbose while also being distinct from the builtin Rust `.saturating_*` methods.

### DIFF
--- a/extension/src/saturation.rs
+++ b/extension/src/saturation.rs
@@ -4,13 +4,13 @@ use pgx::*;
 
 /// Computes x+y, saturating at the numeric bounds instead of overflowing
 #[pg_extern(schema = "toolkit_experimental", immutable, parallel_safe)]
-fn addition_saturate_to_min(x: i32, y: i32) -> i32 {
+fn saturating_add(x: i32, y: i32) -> i32 {
     x.saturating_add(y)
 }
 
 /// Computes x+y, saturating at 0 for the minimum bound instead of i32::MIN
 #[pg_extern(schema = "toolkit_experimental", immutable, parallel_safe)]
-fn addition_saturate_to_zero(x: i32, y: i32) -> i32 {
+fn saturating_add_pos(x: i32, y: i32) -> i32 {
     // check to see if abs of y is greater than the abs of x?
     let result = x.saturating_add(y);
     if result > 0 {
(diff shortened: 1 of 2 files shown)


extension/src/saturation.rs
```diff
@@ -0,0 +1,114 @@
+use pgx::*;
+
+/// Saturating Math for Integers
+
+/// Computes x+y, saturating at the numeric bounds instead of overflowing
+#[pg_extern(schema = "toolkit_experimental", immutable, parallel_safe)]
+fn addition_saturate_to_min(x: i32, y: i32) -> i32 {
```
WireBaron, line 7: I'd go with `saturating_add` here.  It's more succinct and makes it more obvious that it's saturating on both high and low ends.

extension/src/saturation.rs
```diff
@@ -0,0 +1,114 @@
...
+    x.saturating_add(y)
+}
+
+/// Computes x+y, saturating at 0 for the minimum bound instead of i32::MIN
+#[pg_extern(schema = "toolkit_experimental", immutable, parallel_safe)]
+fn addition_saturate_to_zero(x: i32, y: i32) -> i32 {
```
WireBaron, line 13: Maybe `saturating_add_pos` or `pos_saturating_add`?

extension/src/saturation.rs
```diff
@@ -0,0 +1,114 @@
...
+
+/// Computes x*y, saturating at the numeric bounds instead of overflowing
+#[pg_extern(schema = "toolkit_experimental", immutable, parallel_safe)]
+fn multiplication_saturate_to_min(x: i32, y: i32) -> i32 {
+    x.saturating_mul(y)
+}
+
+/// Computes x*y, saturating at 0 for the minimum bound instead of i32::MIN
+#[pg_extern(schema = "toolkit_experimental", immutable, parallel_safe)]
+fn multiplication_saturate_to_zero(x: i32, y: i32) -> i32 {
```
WireBaron, line 47: I don't think we actually need a saturate_to_zero for multiplication.  If a user needs this to be positive they probably need the inputs to this to be positive.  If they've checked that, then the result will necessarily be positive.

extension/src/saturation.rs
```diff
@@ -0,0 +1,114 @@
...
+    if result > 0 {
+        result
+    } else {
+        0
+    }
+}
+
+/// Computes x-y, saturating at the numeric bounds instead of overflowing.
+#[pg_extern(schema = "toolkit_experimental", immutable, parallel_safe)]
+fn subtraction_saturate_to_min(x: i32, y: i32) -> i32 {
```
WireBaron, line 25: `saturating_sub`, and `pos_saturating_sub`?

WireBaron (commented):
Code looks good, but we do need to figure out better names.  Maybe @rtwalker has some good suggestions?

Also adding @JLockerman as a reviewer since he requested this.

epgts (commented):
I like Brian's names. Otherwise LGTM.

extension/src/saturation.rs
rtwalker: +1 for `saturating_add_pos`

thatzopoulos:
Updated names to `saturating_***` and `saturating_***_pos`. Removed test for multiplication saturating to zero.

rtwalker (approved):

WireBaron (approved):

PR #458, opened 2022-06-28
Add `count_min_sketch` aggregate and `approx_count` accessor

### LINKED ISSUE #373: Implement an Approximate Count aggregate using Count-Min Sketch

rtwalker:
Add `count_min_sketch` aggregate and `approx_count` accessor. 

The initial implementations work only on `text` columns

Resolves #373

### DIFF
--- a/extension/src/countminsketch.rs
+++ b/extension/src/countminsketch.rs
@@ -1,0 +1,270 @@
+use pgx::*;
+
+use aggregate_builder::aggregate;
+use countminsketch::{CountMinHashFn, CountMinSketch as CountMinSketchInternal};
+use flat_serialize::*;
+
+use crate::{
+    flatten,
+    palloc::{Inner, Internal},
+    pg_type,
+    raw::bytea,
+    ron_inout_funcs,
+};
+
+#[pg_schema]
+pub mod toolkit_experimental {
+    use super::*;
+
+    pg_type! {
+        #[derive(Debug)]
+        struct CountMinSketch<'input> {
+            width: u32,
+            depth: u32,
+            counters: [i64; self.width * self.depth],
+        }
+    }
+
+    impl CountMinSketch<'_> {
+        fn new(width: u32, depth: u32, counters: Vec<i64>) -> Self {
+            let counters_arr = counters.try_into().unwrap();
+            unsafe {
+                flatten!(CountMinSketch {
+                    width,
+                    depth,
+                    counters: counters_arr,
+                })
+            }
+        }
+
+        pub fn to_internal_countminsketch(&self) -> CountMinSketchInternal {
+            let depth: u64 = self.depth.into();
+            let hashfuncs = (1..=depth).map(CountMinHashFn::with_key).collect();
+
+            let mut counters: Vec<Vec<i64>> = Vec::with_capacity(self.depth as usize);
+            let row_width = self.width as usize;
+            for row in 0..self.depth {
+                let row_start = (row * self.width) as usize;
+                counters.push(
+                    self.counters
+                        .iter()
+                        .skip(row_start)
+                        .take(row_width)
+                        .collect(),
+                );
+            }
+
+            CountMinSketchInternal::new(
+                self.width as usize,
+                self.depth as usize,
+                hashfuncs,
+                counters,
+            )
+        }
+
+        pub fn from_internal_countminsketch(sketch: &mut CountMinSketchInternal) -> Self {
+            CountMinSketch::new(
+                sketch.width().try_into().unwrap(),
+                sketch.depth().try_into().unwrap(),
+                sketch.counters().iter().flatten().cloned().collect(),
+            )
+        }
+    }
+
+    ron_inout_funcs!(CountMinSketch);
+}
+
+use toolkit_experimental::CountMinSketch;
+
+#[aggregate]
+impl toolkit_experimental::count_min_sketch {
+    type State = CountMinSketchInternal;
+
+    fn transition(
+        state: Option<State>,
+        #[sql_type("text")] value: Option<String>,
+        #[sql_type("float")] error: f64,
+        #[sql_type("float")] probability: f64,
+    ) -> Option<State> {
+        let value = match value {
+            None => return state,
+            Some(value) => value,
+        };
+
+        let mut state = match state {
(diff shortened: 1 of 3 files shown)


extension/src/countminsketch.rs
```diff
@@ -0,0 +1,143 @@
...
+                Some(a)
+            }
+        }
+    }
+}
+
+#[pg_extern(immutable, parallel_safe, schema = "toolkit_experimental")]
+pub fn approx_count(item: String, aggregate: Option<CountMinSketch>) -> i64 {
+    let sketch = aggregate.unwrap();
+    CountMinSketch::to_internal_countminsketch(&sketch).estimate(item)
```
WireBaron, line 142: This is the antipattern that is my biggest grievance with how we use the crates subdirectory.  Converting the Postgres CountMinSketch object back into the crate object so that we can call a function on it is really not efficient.  We can leave this in place for now, but we'll need to implement the estimate functionality directly on top of the Postgres structure.

extension/src/countminsketch.rs
```diff
@@ -0,0 +1,143 @@
...
+}
```
WireBaron, line 143: We will need some pg_tests showing that this works correctly in Postgres.  You should be able to mostly copy a `SPI::execute` type test from any of the other aggregate rs files under extension.

extension/src/countminsketch.rs
```diff
@@ -0,0 +1,143 @@
...
+#[pg_schema]
+pub mod toolkit_experimental {
+    use super::*;
+
+    pg_type! {
+        #[derive(Debug)]
+        struct CountMinSketch<'input> {
+            width: u32,
+            depth: u32,
+            hash_keys: [u64; self.depth],
```
WireBaron, line 24: Given that we don't allow a user to input these, I'm not sure it's worth actually storing them.  We should be able to rely on these being the set 1..depth.

extension/src/countminsketch.rs
```diff
@@ -0,0 +1,143 @@
...
+            counters: [i64; self.width * self.depth],
+        }
+    }
+
+    impl CountMinSketch<'_> {
+        pub fn new(width: u32, depth: u32, hash_keys: Vec<u64>, counters: Vec<i64>) -> Self {
```
WireBaron, line 30: This seems like it's only useful in `from_internal_countminsketch`.  If that's the case it shouldn't be public.

extension/src/countminsketch.rs
rtwalker: tests added :)

extension/src/countminsketch.rs
```diff
@@ -0,0 +1,275 @@
...
+        Some(state)
+    }
+
+    fn finally(state: Option<&mut State>) -> Option<CountMinSketch<'static>> {
+        let state = match state {
+            None => return None,
+            Some(state) => CountMinSketch::from_internal_countminsketch(state),
+        };
+
+        Some(state)
```
WireBaron, line 109: Can this function just be `state.map(CountMinSketch::from_internal_countminsketch)`

WireBaron (approved):
Looks good, just one last suggestion.

epgts (approved):

extension/src/countminsketch.rs
epgts: Having implemented this two or three times so far myself, I couldn't agree more that it's a mess! But I think the answer is to break `extension` up into more crates, so that we can build business logic directly against pg_type structs without all this destruction and reconstruction and so that we can avoid giving rustc an ever-growing pile of files to handle when I make a one-line change to only one of those files. What stops us now is so much of what goes into a `pg_type` is locked inside the `extension` crate.

We've nibbled at the edges of this discussion a few times... how about I spend a day to prototype something like this when I get back from vacation?

Thanks!

rtwalker: I'd be interested to see that!

Issue #3, opened 2020-12-07
Approximate COUNT DISTINCT


JLockerman:
## What's the functionality you would like to add ##
An approximate method to count the number of distinct elements in a table, such as HyperLogLog.

## How would the function be used ##
```SQL
SELECT approx_distinct(value) FROM data;
 approx_distinct
-----------------
              42

CREATE MATERIALIZED VIEW distinct_per_hour
WITH (timescaledb.continuous) AS
SELECT time_bucket(INTERVAL '1 hour', time) AS bucket,
    approx_distinct(value),
FROM data
GROUP BY bucket;

SELECT * FROM distinct_per_hour LIMIT 4;
         bucket         | approx_distinct 
------------------------+-----------------
   2020-01-03 01:00:00  |              31
   2020-01-03 02:00:00  |              27
   2020-01-03 03:00:00  |              30
   2020-01-03 04:00:00  |               1

```

## Why should this feature be added? ##
1. Without proper indexes `COUNT DISTINCT` can take time and space proportionate to the amount of memory, approximate counting can be done in ~constant space.
2. An approximate `COUNT DISTINCT` can be used with continuous aggregates to provide fast approximate counts for time intervals

## Remaining Work ##

- [x] Merge HyperLogLog++ implementation: https://github.com/timescale/timescale-analytics/pull/170
- [ ] Decide the final sizing API: https://github.com/timescale/timescale-analytics/issues/3#issuecomment-870662375
- [ ] Add Default collation to C collation transform (due to https://github.com/timescale/timescale-analytics/issues/148)

JLockerman:
Am experimental version of this is available in [0.1.0](https://github.com/timescale/timescale-analytics/releases/tag/0.1.0), and should also be available on Forge soon.

JLockerman:
Note: the hyperloglog impelemtation is in the process of being replaced with a Hyperloglog++ implementation on [the `jl/hllpp` branch](https://github.com/timescale/timescale-analytics/tree/jl/hllpp) (work is done on the actual implementation, it just needs to be attached to the SQL type).

JLockerman:
Note from issue https://github.com/timescale/timescale-analytics/issues/148: all deterministic collations (which include all default collations) are treated the same for `text_hash()`, they are treated like binary hash. This means that we may want to default the default collation to always be equivalent to the C collation for HLL purposes, and avoid that entire class of errors. I don't think that will work for cases that care about ordering and not just equality though.

[...]

PR #459, opened 2022-06-30
Th/approx count distinct

thatzopoulos:
Issue : https://github.com/timescale/timescaledb-toolkit/issues/3

This PR adds a new function `approx_count_distinct`, a method to count the number of distinct elements in a table, such as hyperloglog, that does not require the user to specify the number of buckets/registers. The number of buckets/registers is always set to 32,768

### DIFF
--- a/extension/src/hyperloglog.rs
+++ b/extension/src/hyperloglog.rs
@@ -38,13 +38,43 @@ pub fn hyperloglog_trans(
     fc: pg_sys::FunctionCallInfo,
 ) -> Option<Internal> {
     // let state: Internal = Internal::from_datum();
-    hyperloglog_trans_inner(unsafe{ state.to_inner() }, size, value, fc).internal()
+    hyperloglog_trans_inner(
+        unsafe { state.to_inner() },
+        size,
+        value,
+        fc,
+        unsafe { pgx::get_getarg_type(fc, 2) },
+    )
+    .internal()
 }
+
+const APPROX_COUNT_DISTINCT_DEFAULT_SIZE: i32 = 32678;
+
+/// Similar to hyperloglog_trans(), except size is set to a default of 32,678
+#[pg_extern(immutable, parallel_safe,schema = "toolkit_experimental")]
+pub fn approx_count_distinct_trans(
+    state: Internal,
+    // TODO we want to use crate::raw::AnyElement but it doesn't work for some reason...
+    value: Option<AnyElement>,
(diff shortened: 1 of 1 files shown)


WireBaron (approved):

extension/src/hyperloglog.rs
```diff
@@ -38,13 +38,29 @@ pub fn hyperloglog_trans(
...
+
+/// Similar to hyperloglog_trans(), except size is set to a default of 32,678
+#[pg_extern(immutable, parallel_safe,schema = "toolkit_experimental")]
+pub fn approx_count_distinct_trans(
+    state: Internal,
+    // TODO we want to use crate::raw::AnyElement but it doesn't work for some reason...
+    value: Option<AnyElement>,
+    fc: pg_sys::FunctionCallInfo,
+) -> Option<Internal> {
+    // let state: Internal = Internal::from_datum();
```
rtwalker, line 54: Is the commented out code intentional? is it related to the TODO above?

extension/src/hyperloglog.rs
```diff
@@ -38,13 +38,29 @@ pub fn hyperloglog_trans(
...
+    hyperloglog_trans_inner(unsafe{ state.to_inner() }, APPROX_COUNT_DISTINCT_DEFAULT_SIZE, value, fc,unsafe {pgx::get_getarg_type(fc,1)}).internal()
```
rtwalker, line 55: would you indulge a formatting change?
```rust
    hyperloglog_trans_inner(
        unsafe { state.to_inner() },
        APPROX_COUNT_DISTINCT_DEFAULT_SIZE,
        value,
        fc,
        unsafe { pgx::get_getarg_type(fc, 1) },
    )
    .internal()
```

rtwalker (approved):

extension/src/hyperloglog.rs
epgts: can we just defer to rustfmt? :)

extension/src/hyperloglog.rs
thatzopoulos: I kept this comment and the TODO since they were still both in the `hyperloglog_trans` function, I'm unsure if the two are related. I don't mind deleting them from this function

extension/src/hyperloglog.rs
thatzopoulos: I will push up the formatting change

PR #461, opened 2022-07-05
Prepare for 1.9.0 release cycle

rtwalker:


WireBaron (approved):

WireBaron (approved):
From this morning's conversation it sounds like we also want to update the version in cargo.toml as part of this.

PR #465, opened 2022-07-08
Fixing off by one error in state_agg interpolate

### LINKED ISSUE #464: state_agg interpolation hits an error if the interpolated first state doesn't already exist in the agg
**Describe the bug**
When creating an interpolated state, we will add a new duration object to handle the initial state if one doesn't already exist.  Unfortunately there's an off by one error in the object we create, where it's computing the state_end value.  This results in the duration object dropping the last character, which causes it to fail to match the string for the first state when doing the interpolation.

This shows up as the following error:
ERROR:  assertion failed: first_state < durations.len() && last_state < durations.len()
CONTEXT:  extension/src/state_aggregate.rs:92:13

WireBaron:
Fixes #464

### DIFF
--- a/extension/src/state_aggregate.rs
+++ b/extension/src/state_aggregate.rs
@@ -158,7 +158,7 @@ pub mod toolkit_experimental {
                                 durations.push(DurationInState {
                                     duration: start_interval,
                                     state_beg: states.len() as u32,
-                                    state_end: (states.len() + start_state.len() - 1) as u32,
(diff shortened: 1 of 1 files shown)


epgts (commented):
Can we add a test that fails without the fix?

thatzopoulos (approved):

rtwalker (commented):
Confirmed that the test fails without this fix!

rtwalker (approved):

epgts (approved):

PR #467, opened 2022-07-12
Document supported platforms.

epgts:


jerryxwu:
@epgts the issue that you're linking doesn't appear to be correct.

PR #471, opened 2022-07-14
Add a template for feature documentation.

epgts:


docs/template.md
```diff
@@ -0,0 +1,85 @@
+# FEATURE-NAME [<sup><mark>experimental</mark></sup>](/docs/README.md#tag-notes)
+
+- Current status: ( prototype | experimental | stabilizing | stable )
+- Effort remaining:  ( little | some | lots )
```
epgts, line 4: Maybe the right way to gauge "effort" is "how many of our six-week increments might it take"?

rtwalker (approved):

docs/template.md
jerryxwu: If we go with this, are we building assumption that an increment is about six weeks? This could easily change though. Perhaps we can just gauge in "person-month"?

davidkohn88:
We aren't building in that assumption and for me, I'd like to keep it
intentionally vague, there are lots of pitfalls to estimation that I think
we can skirt by doing that, the goal is mainly as an aid for
prioritization, and to get a sense of cost in which I want to speak in very
broad strokes.

WireBaron (approved):

PR #474, opened 2022-07-20
funnel prototype

epgts:


docs/funnel.md
```diff
@@ -0,0 +1,228 @@
...
+Looking for rows where one event occurs within an interval of another, e.g.
+- two LOGIN events within one week
+- BUY event within one hour of SEARCH
+
+#### two LOGIN events within one week
+
+```SQL
+SELECT toolkit_experimental.within_interval(
+	'LOGIN', 'LOGIN', '1 week',
+	toolkit_experimental.funnel_agg("user", event, time))
```
davidkohn88, line 55: I don't know why user is included in the aggregate, I think it should probably just be in the group by so: 

`SELECT user, funnel_agg(event, time) FROM funnel_test GROUP BY user`

docs/funnel.md
```diff
@@ -0,0 +1,228 @@
...
+Name is terrible, this is just a first test.
+
+Looking for rows where one event occurs within an interval of another, e.g.
+- two LOGIN events within one week
+- BUY event within one hour of SEARCH
+
+#### two LOGIN events within one week
+
+```SQL
+SELECT toolkit_experimental.within_interval(
```
davidkohn88, line 53: One overarching question on this for me is whether we want to use accessors for this or if we want to, for instance, adopt some syntax that constructs a matcher for the the funnel and can be dropped into the where clause like so, I'd also say that many operations on these will probably be best done on some sort of materialization:

```SQL
CREATE MATERIALIZED VIEW funnels as SELECT user, toolkit_experimental.funnel_agg(event, time) funnel
FROM funnel_test
GROUP BY user;

-- Then we can search with the WHERE clause:

SELECT * FROM funnels 

WHERE funnel ? event('LOGIN') -> event('SEARCH', within => '1 hour'::interval) 
OR funnel ? event('LOGIN') -> event('LOGIN', withi n=> '1 hour'::interval, consecutive => true);

```

docs/funnel.md
davidkohn88: I also wonder whether we want to add an option to the aggregate to deduplicate the underlying data or not...or maybe we could just compress the whole thing then it wouldn't matter so much....

davidkohn88 (commented):
Looking at this some more one big question I have is whether this is actually something we should think about working with the Promscale team on to develop a more integrated solution where we have some more control over the data model...I think there could be something really interesting here though and some very interesting synergies with compression and other bits. 

The nifty thing for me is that this data structure looks kinda like our compressed segments in a lot of ways and I think it would be interesting to develop ways to push this down to the compressed data directly to do the aggregate and apply the where clause at the same time in at least some cases. 

We also will DEFINITELY need a rollup function for this aggregate.

Anyway, awesome start and really helped get the juices flowing in terms of thinking about this and I'm looking forward to the discussion.

epgts:
Discussion moved to issue #475

PR #484, opened 2022-08-01
Add macro to generate function and struct for common operator accessors.

epgts:


extension/src/accessors.rs
```diff
@@ -1,271 +1,120 @@
...
-    build, flatten,
-    pg_type,
-    ron_inout_funcs,
-};
+use counter_agg::range::I64Range;
+use flat_serialize::FlatSerializable;
+
+use crate::{build, flatten, pg_type, ron_inout_funcs};
+
+macro_rules! accessor {
```
JLockerman, line 10: I _think_ this can be simplified to use

```rust
macro_rules! accessor {
    (
        $name: ident (
            $($field:ident : $typ: tt),* $(,)?
        )
    ) => {
         // ...
    };
}

macro_rules! accessor_fn_impl {
    (
        $name: ident (
            $( $field:ident : $typ: tt ),*
                $(,)?
        )
    ) => {
        // ...
    };
}
```

([playground example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=b5cfef7f574ef7196a200f3d0d3c6a8f))

if you think that makes it simpler.

JLockerman (approved):
LGTM

WireBaron (approved):
I'd suggest modifying the macro as Josh suggested, but this looks good overall.

extension/src/accessors.rs
epgts: I do, and as I indicated on the call, I didn't understand why `vals` was necessary and had tried to get rid of it.  I pushed a commit attempting your suggestion.  Can you tell what is wrong?  Thanks.

extension/src/accessors.rs
```diff
@@ -1,271 +1,101 @@
...
+        $name: ident (
+            $($field:ident : $typ: tt),* $(,)?
+        )
+    ) => {
+        ::paste::paste! {
+            $crate::pg_type!{
+                // TODO Move into pg_type as we don't care to vary it.
+                #[derive(Debug)]
+                struct [<Accessor $name:camel>] {
+                $($field: $typ),*
```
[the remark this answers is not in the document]
epgts: Yep that did it.  Thanks!

PR #487, opened 2022-08-04
Use flat_serialize::FlatSerializable directly where needed.

epgts:
Don't require callers of pg_type! and build! to use it for them.


rtwalker (approved):

WireBaron (approved):
Nice change

PR #488, opened 2022-08-08
Stabilize timevector

WireBaron:
This change moves the timevector function and object out of the toolkit_experimental schema.  It also adds a template issue for feature stabilization.

### DIFF
--- a/extension/src/stabilization_info.rs
+++ b/extension/src/stabilization_info.rs
@@ -14,18 +14,18 @@ crate::functions_stabilized_at! {
     "1.9.0" => {
         accessorunnest_in(cstring),
         accessorunnest_out(accessorunnest),
-        arrow_timevector_unnest(timevector,accessorunnest),
-        rollup(timevector),
+        arrow_timevector_unnest(timevector_tstz_f64,accessorunnest),
+        rollup(timevector_tstz_f64),
         timevector(timestamp with time zone,double precision),
         timevector_combine(internal,internal),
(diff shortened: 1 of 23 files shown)


epgts:
update tests still failing with "ERROR: function timevector(timestamp with time zone, double precision) does not exist", which I can reproduce locally.

extension/src/stabilization_info.rs
```diff
@@ -11,6 +11,27 @@
...
+        timevector_deserialize(bytea,internal),
+        timevector_final(internal),
+        timevector_in(cstring),
+        timevector_out(timevector),
+        timevector_serialize(internal),
+        timevector_trans(internal,timestamp with time zone,double precision),
+        unnest(timevector),
+        unnest(),
+    }
+    "1.8.0" => {
```
epgts, line 31: Do we need these empty blocks?

epgts (commented):
Looks good.

Also looks like an opportune time to rustfmt all the mod time_vector files :)

extension/src/stabilization_info.rs
WireBaron: Fair question, I had thought it better to be explicit that nothing was stabilized in a particular version.  However, I'm not particular stuck on this if you feel it would be better to exclude them.

epgts (approved):

WireBaron:
Ugh, rushed through this and failed to notice that my last push where I formatted time_vector.rs and squashed my commits failed.

Issue #466, opened 2022-07-11
Normalize a monthly value to a fixed-length month


jerryxwu:
**Is your feature request related to a problem? Please describe.**
When comparing month to month values such as sales, it's often more meaningful if the values are normalized to fixed-length month (like 30 days).

**Describe the solution you'd like**
Create a new function with the following interface. The function will determine how many days there are for the month given as `time` and normalize the value appropriately. As an example, `monthly_normalized(812, '2020-02-01'::timestamptz, 30)` will return  840 because the year 2020 is a leap year so Feb 2020 has 29 days.
```SQL
monthly_normalized(
   value: DOUBLE PRECISION // required
   time: TIMESTAMPTZ // required
   n_days = 30: SMALLINT // optional and default to 30
) RETURN DOUBLE PRECISION
```

For the case of normalizing the monthly sales, one could have a query looking like this:
```SQL
WITH t AS (
  SELECT date_trunc('month', time) AS month, 
    sum(sale) AS monthly_sale
  FROM sales
  GROUP BY month
)
SELECT 
  month, 
  monthly_sale, 
  monthly_normalized(month, monthly_sale) AS normalized_monthly_sale 
FROM t
```

**Describe alternatives you've considered**
Without such as function, one could write the query above as the following, which is a bit cumbersome and harder to read.
```SQL
WITH t AS (
  SELECT date_trunc('month', time) AS month, 
    sum(sale) AS monthly_sale
  FROM sales
  GROUP BY month
)
SELECT month,
  monthly_sale, 
  monthly_sale / EXTRACT('day' FROM month + '1 month'::interval - month) * 30 AS normalized_monthly_sale 
FROM t
```

**Additional context**
The function could initially be implemented in PL/pgSQL. A rust implementation might be more efficient.

PR #490, opened 2022-08-08
Monthly Normalization

thatzopoulos:
Issue : https://github.com/timescale/timescaledb-toolkit/issues/466

Added two new sql functions :
- days_in_month
- month_normalize

### DIFF
--- a/extension/src/utilities.rs
+++ b/extension/src/utilities.rs
@@ -1,11 +1,14 @@
-use pgx::*;
 use crate::raw::TimestampTz;
+use pgx::*;
 
-#[pg_extern(name="generate_periodic_normal_series", schema = "toolkit_experimental")]
+#[pg_extern(
+    name = "generate_periodic_normal_series",
+    schema = "toolkit_experimental"
+)]
 pub fn default_generate_periodic_normal_series(
     series_start: crate::raw::TimestampTz,
     rng_seed: Option<i64>,
-) -> impl std::iter::Iterator<Item = (name!(time,TimestampTz),name!(value,f64))> + 'static {
+) -> impl std::iter::Iterator<Item = (name!(time, TimestampTz), name!(value, f64))> + 'static {
     generate_periodic_normal_series(series_start, None, None, None, None, None, None, rng_seed)
 }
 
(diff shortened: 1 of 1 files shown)


thatzopoulos:
All changes in `utilities.rs` before line 88 are from `cargo fmt` and can be removed if necessary

extension/src/utilities.rs
```diff
@@ -53,26 +61,47 @@ pub fn generate_periodic_normal_series(
...
+create or replace function toolkit_experimental.days_in_month(date timestamptz) RETURNS int AS $$\n\
+SELECT CAST(EXTRACT('day' FROM ($1 + interval '1 month' - $1)) as INTEGER)\n\
+$$ LANGUAGE SQL STRICT IMMUTABLE PARALLEL SAFE;\n\
+",
+    name = "days_in_month",
+);
+
+// Normalizes metric based on reference date and days
+extension_sql!("\n\
+create or replace function toolkit_experimental.month_normalize(metric float8, reference_date timestamptz, days float8 DEFAULT 365.25/12) RETURNS float8 AS $$\n\
```
davidkohn88, line 100: @JLockerman Do we need to do something for these in terms dealing with the potential security vulnerability of using CREATE OR REPLACE?

extension/src/utilities.rs
```diff
@@ -110,4 +149,28 @@ mod tests {
...
+        });
+    }
+    #[pg_test]
+    fn test_monthly_normalize() {
+        Spi::execute(|client| {
+            let test_val = client.select("SELECT toolkit_experimental.month_normalize(1000,'2021-01-01 00:00:00+03'::timestamptz)",None,None,).first().get_one::<f64>().unwrap();
+            assert_eq!(test_val, 981.8548387096774f64);
+        });
+        Spi::execute(|client| {
+            let test_val = client.select("SELECT toolkit_experimental.month_normalize(1000,'2021-01-01 00:00:00+03'::timestamptz,30.5)",None,None,).first().get_one::<f64>().unwrap();
```
davidkohn88, line 172: I'd want to test with a value like 30 to make sure we don't accidentally do integer arithmetic on this. (Could just add one more test case).

davidkohn88 (commented):
Only big thing is the potential security issue with using create or replace, I wonder how we can handle that and if we need to do something about it for `time_bucket_range` as well.

extension/src/utilities.rs
JLockerman: @davidkohn88 if it's in experimental I think we can do an unconditional `CREATE`; the experimental schema is always dropped on update so any collision is a bug.

davidkohn88: What about when we graduate it?

davidkohn88: But for now that should work...

davidkohn88 (approved):
LGTM

davidkohn88 (approved):

PR #491, opened 2022-08-08
Stabilize accessors.

epgts:


extension/src/stabilization_info.rs
```diff
@@ -181,10 +430,70 @@ crate::types_stabilized_at! {
 
 crate::operators_stabilized_at! {
     STABLE_OPERATORS
+    "1.9.0" => {
+        // TODO hmmmmmmm
```
epgts, line 434: @JLockerman Can you help with this?  I think this is something you warned me about last week, but it's not obvious to me what the macro wants.

Based on the `// TODO JOSH - this may not be right` in stabilization_tests, I think that's not surprising :)

Thanks!

epgts: AH, never mind!  I see Brian has fixed it on br/timevec_stable

WireBaron (approved):
Looks pretty straightforward, good to go.

Issue #445, opened 2022-06-10
Create financial-focused OHLC aggregation


ryanbooz:
**Is your feature request related to a problem? Please describe.**
Many users of TimescaleDB utilize financial time-series data. Many of our examples and tutorials show how to do basic OHLC bucketing, saving at least four columns using MAX/MIN/FIRST/LAST as the functions. It's a repetitive processes and some functions don't work well with re-aggregation - and even when they do (MAX/MIN), most of our documentation/context warns users against using functions to re-aggregate unless they were designed to do so.

**Describe the solution you'd like**
Saving an aggregate that stores the necessary data that can then be accessed at query time to return the appropriate value for financial charts. At a minimum, users would have to be able to easily identify the Open/High/Low/Closing value within a bucket. Allowing re-aggregation over smaller buckets would be powerful, which means that users could now satisfy multiple bucket widths from one CAGG, rather than having to create various granularities. (1-min, 5-min, 10-min, 60-min).

**Describe alternatives you've considered**
Currently, the only solution we really have to offer people is to follow our schema examples for OHLC data using separate aggregate functions and creating multiple CAGGs. The big challenge here, is that developing an application to view multiple "zoom levels" isn't easy without more complicated `UNION` examples that pull from different CAGGs depending on the "zoom level" (overall query length). Being able to using the same SQL query and simply re-aggregate to a different time bucket would be 🔥

**Additional context**
@davidkohn88 suggested I jump-start the conversation by creating the issue 😉. We have so many users that do this kind of aggregation. If we could simplify their storage (one CAGG) and query (only have to change the time bucket interval value) - this could simplify how users query data significantly.

jerryxwu:
@ryanbooz @davidkohn88 One obvious UX design that I can think of looks like this,

Create a 1-minute CAGG from a raw trading data table called “raw_trades”.
```SQL
CREATE MATERIALIZED VIEW ohlcv_1m
WITH (timescaledb.continuous) AS
SELECT time_bucket('1 minute', ts) AS time,
       symbol, 
       ts_ohlcv(ts, price, volume) AS ohlcv
FROM raw_trades
GROUP BY time, symbol
```

Rollup the 1-minute ohlcv aggregates to the desired time interval at query time and get open, high, low, close and volume data through accessor functions ts_open, ts_high, and etc.
```SQL
SELECT time_bucket('1 hour', time) AS hour,
       ts_open(rollup(ohlcv)) AS open,
[...]

PR #496, opened 2022-08-15
Create OHLC aggregate and accompanying accessors

rtwalker:
Begins to address #445. 

Introduces OHLC aggregate called as `toolkit_experimental.ohlc(timestamp, price)`. 

The following accessors are provided for use with the `OpenHighLowClose` aggregate:
* `open`, `high`, `low`, `close`

along with the timestamp-returning versions:
* `open_at`, `high_at`, `low_at`, `close_at`

Additionally, there is a new `toolkit_experimental.rollup()` aggregate that takes `OpenHighLowClose` aggregates as input.

In the event that multiple rows contain the highest or lowest value, only the earliest occurrence is retained. Similarly, should multiple rows contain the opening or closing timestamp, this implementation only updates the open/close points with a new value when the new timestamp is strictly earlier/later than current timestamp.

Additional work on #445 remains to address handling volume (and possibly creating technical indicators).

### DIFF
--- a/extension/src/ohlc.rs
+++ b/extension/src/ohlc.rs
@@ -1,0 +1,644 @@
+use pgx::*;
+
+use crate::{
+    aggregate_utils::in_aggregate_context,
+    flatten,
+    palloc::{Inner, Internal, InternalAsValue, ToInternal},
+    pg_type,
+    raw::bytea,
+    ron_inout_funcs,
+};
+use tspoint::TSPoint;
+
+#[pg_schema]
+pub mod toolkit_experimental {
+    use super::*;
+
+    pg_type! {
+        #[derive(Copy, Debug)]
+        #[repr(C)]
+        struct OpenHighLowClose {
+            open: TSPoint,
+            high: TSPoint,
+            low: TSPoint,
+            close: TSPoint,
+        }
+    }
+
+    impl OpenHighLowClose<'_> {
+        pub fn new() -> Self {
+            unsafe {
+                flatten!(OpenHighLowClose {
+                    open: TSPoint {
+                        ts: i64::MAX,
+                        val: 0.0
+                    },
+                    high: TSPoint {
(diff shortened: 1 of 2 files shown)


extension/src/ohlc.rs
```diff
@@ -0,0 +1,644 @@
...
+        struct OpenHighLowClose {
+            open: TSPoint,
+            high: TSPoint,
+            low: TSPoint,
+            close: TSPoint,
+        }
+    }
+
+    impl OpenHighLowClose<'_> {
+        pub fn new() -> Self {
```
WireBaron, line 29: As discussed, probably better if this takes the first point.  Feel free to add a comment and file an issue to address this later.

extension/src/ohlc.rs
```diff
@@ -0,0 +1,644 @@
...
+    value: Option<OpenHighLowClose<'input>>,
+    fcinfo: pg_sys::FunctionCallInfo,
+) -> Option<Inner<OpenHighLowClose<'input>>> {
+    unsafe {
+        in_aggregate_context(fcinfo, || match (state, value) {
+            (state, None) => state,
+            (None, Some(value)) => {
+                let mut state = OpenHighLowClose::new();
+                state.combine(&value);
+                Some(state.into())
```
WireBaron, line 186: You should be able to just return value or a copy of value here.

WireBaron (approved):
Didn't really have any feedback aside what came up at all hands this morning.

extension/src/ohlc.rs
rtwalker: sure can :)

rtwalker:
Added tests for extreme values in both the timestamp and price columns. Also required `OpenHighLowClose::new()` to take a `TSPoint` and got rid of default values.

WireBaron (approved):
Looks good

PR #498, opened 2022-08-16
Stabilize ASAP

WireBaron:
This PR moves the ASAP implementation out of toolkit_experimental.  It also removes some of the odd behavior where we were trying to normalize the incoming data points before applying the algorithm (users should do this themselves if this is the behavior they desire).  As a result, we should now correctly match the reference implementation exactly, and this change also adds a PR to that effect.

### DIFF
--- a/extension/src/asap.rs
+++ b/extension/src/asap.rs
@@ -12,16 +12,6 @@ use tspoint::TSPoint;
 
 use crate::time_vector::{Timevector_TSTZ_F64, Timevector_TSTZ_F64Data};
 
-// This is included for debug purposes and probably should not leave experimental
-#[pg_extern(schema = "toolkit_experimental", immutable, parallel_safe)]
-pub fn asap_smooth_raw(data: Vec<f64>, resolution: i32) -> Vec<f64> {
-    asap_smooth(&data, resolution as u32)
-}
-
-// hack to allow us to qualify names with "toolkit_experimental"
-// so that pgx generates the correct SQL
-mod toolkit_experimental {}
-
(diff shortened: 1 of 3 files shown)


thatzopoulos (approved):

extension/src/asap.rs
```diff
@@ -124,22 +95,15 @@ fn asap_final_inner(
...
-            // of points is at least twice the resolution.  Otherwise we keep the number of
-            // points, but still normalize them to equal sized buckets.
-            let downsample_interval = if points.len() >= 2 * state.resolution as usize {
-                find_downsample_interval(&points, state.resolution as i64)
-            } else {
-                (points.last().unwrap().ts - points.first().unwrap().ts) / points.len() as i64
-            };
-            let mut normal = downsample_and_gapfill_to_normal_form(&points, downsample_interval);
+
             let start_ts = points.first().unwrap().ts;
```
epgts, line 99: Can this happen from user error alone?  If so, we ought to provide an error message by changing `.unwrap()` to `.expect("points must not be empty")` or something.  (Same question for another unwrap below in asap_on_timevector)

Is it possible for end_ts to be the same as start_ts and does that matter?

epgts (approved):

extension/src/asap.rs
WireBaron: This should be impossible to hit during normal use.  We exit this function right away if called on `None` and won't create a `Some` unless we have at least on value.  Similarly we need a time/value pair to create a timevector.  If the data is only one point we should just return that point, and, now that I just fixed the div0 we were hitting in that case, we do so.

PR #500, opened 2022-08-17
Prepare for the 1.10.0 release cycle

thatzopoulos:


thatzopoulos:
Tried to squash those two commits but I'm not allowed to force push to this branch for some reason:
```remote: error: GH006: Protected branch update failed for refs/heads/prepare-for-1.10.0.
remote: error: Cannot force-push to this protected branch```

epgts:
Without defending that rule, it is hitting you because you don't have a `prefix/` on your branch name.

rtwalker (approved):

PR #507, opened 2022-08-23
Fix APPROX_COUNT_DISTINCT_DEFAULT_SIZE value

jeremyhaberman:
The [Hyperloglog docs](https://docs.timescale.com/api/latest/hyperfunctions/approx_count_distincts/hyperloglog/) state the value for this should be a power of 2 between 16 and 2^18 and that if you're not sure, start with 32768 (2^15).

The value here (32678) seems to have been an honest mistake.


rtwalker:
Nice catch, and thank you for the contribution!

PR #510, opened 2022-08-29
Removed toolkit_experimental from stabilized accessor

thatzopoulos:


extension/src/time_vector/pipeline/delta.rs
```diff
@@ -10,7 +10,7 @@ use super::*;
     schema = "toolkit_experimental"
 )]
 pub fn delta_pipeline_element<'p, 'e>(
-    accessor: toolkit_experimental::AccessorDelta<'p>,
+    accessor: AccessorDelta<'p>,
```
epgts, line 13: image

epgts: I see what's going on.  The root cause is this nonsense of importing things from one module with the base name `toolkit_experimental` into another.  I made a mistake in leaving the pipeline.rs import of AccessorDelta when I should have instead referenced it by absolute path here.  Then Rust's view would have lined up with pgx's view.  Even better would be if there weren't two different views with one of them not checked by a compiler :)

Would you mind making those two changes?

Thanks!

epgts (approved):
Thanks!

PR #513, opened 2022-08-31
Create accessors to first/last data points

### LINKED ISSUE #512: Accessors to retrieve first and last data points from TimeWeightSummary
Creating this issue for tracebility.

A user on [Stackoverflow](https://stackoverflow.com/questions/73475880/is-there-a-way-to-extract-start-and-end-from-a-timeweightsummary-timescaledb) asked to extract the first and the last point value and timestamp from a `time_weight` object.  We think it make sense to add the same accessors to `counter_agg` as well. There's very little risk of exposing the internal structure of the objects as the first and the last data points are needed for the `rollup` functions.

Specifically, the following accessors are discussed:

- `first_val` / `last_val`
- `first_time` / `last_time` (perhaps consider naming them as `first_at` / `last_at` to be consistent with `open_at`, `close_at`, etc.

rtwalker:
Resolves #512 

We can continue bike shedding the names either here or in the linked issue. 

Just wanted to get the PR up.

### DIFF
--- a/extension/src/counter_agg.rs
+++ b/extension/src/counter_agg.rs
@@ -8,6 +8,7 @@ use crate::{
         AccessorExtrapolatedRate, AccessorIdeltaLeft, AccessorIdeltaRight, AccessorIntercept,
         AccessorIrateLeft, AccessorIrateRight, AccessorNumChanges, AccessorNumElements,
         AccessorNumResets, AccessorRate, AccessorSlope, AccessorTimeDelta, AccessorWithBounds,
+        AccessorFirstTime, AccessorFirstVal, AccessorLastTime, AccessorLastVal,
     },
     aggregate_utils::in_aggregate_context,
     flatten,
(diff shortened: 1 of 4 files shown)


extension/src/counter_agg.rs
```diff
@@ -8,6 +8,7 @@ use crate::{
         AccessorExtrapolatedRate, AccessorIdeltaLeft, AccessorIdeltaRight, AccessorIntercept,
         AccessorIrateLeft, AccessorIrateRight, AccessorNumChanges, AccessorNumElements,
         AccessorNumResets, AccessorRate, AccessorSlope, AccessorTimeDelta, AccessorWithBounds,
+        AccessorFirstTime, AccessorFirstVal, AccessorLastTime, AccessorLastVal,
```
epgts, line 11: Let's keep these sorted.

extension/src/stabilization_info.rs
```diff
@@ -11,6 +11,36 @@
...
+        first_time(timeweightsummary),
+        first_val(),
+        first_val(countersummary),
+        first_val(timeweightsummary),
+        last_time(),
+        last_time(countersummary),
+        last_time(timeweightsummary),
+        last_val(),
+        last_val(countersummary),
+        last_val(timeweightsummary),
```
epgts, line 42: It really is interesting to see how much a simple feature costs.

extension/src/time_weighted_average.rs
```diff
@@ -422,6 +488,29 @@ mod tests {
             let stmt = "SELECT average(time_weight('LOCF', ts, val)) FROM test";
             assert!((select_one!(client, stmt, f64) - 10.0).abs() < f64::EPSILON);
 
+            let stmt = "SELECT first_val(time_weight('LOCF', ts, val)) FROM test";
+            assert!((select_one!(client, stmt, f64) - 10.0).abs() < f64::EPSILON);
+            let stmt = "SELECT last_val(time_weight('LOCF', ts, val)) FROM test";
+            assert!((select_one!(client, stmt, f64) - 20.0).abs() < f64::EPSILON);
+
+	    // arrow syntax should be the same
```
epgts, line 496: misaligned comments; maybe you've inserted TAB characters?

epgts (approved):

epgts:
Does it make sense to have this for counter_agg but not gauge_agg?

extension/src/time_weighted_average.rs
rtwalker: Thanks for catching... adding a `(setq indent-tabs-mode nil)` rust-mode-hook as we speak

epgts (approved):

extension/src/counter_agg.rs
epgts, line 11: Oh, can you restore the sorting here?

rtwalker: Was waiting on the gauge agg conversation before I pushed more changes, but they're fixed now :)

rtwalker:
> Does it make sense to have this for counter_agg but not gauge_agg?

Discussed at standup today where it was agreed that it makes sense to wait on #523 to be sorted out before adding these functions to `gauge_agg`.

Issue #220, opened 2021-09-01
TDigest docs still marked as experimental


udesaiitrs:
**Relevant system information:**
 - TimescaleDB Toolkit version: 1.1.0

**Describe the bug**
The docs still mark TDigest as experimental, however I believe the feature is no longer experimental.

**To Reproduce**
Go to https://github.com/timescale/timescaledb-toolkit/blob/1.1.0/docs/tdigest.md

**Expected behavior**
Do not have the "Experimental" superscript in header.

**Actual behavior**
Does have the "Experimental" superscript in header.

Issue #483, opened 2022-07-29
Stabilize current timevector implementation


WireBaron:
## [Timevector](https://github.com/timescale/timescaledb-toolkit/issues/2)

**What evidence do we have the feature is being used**

We have seen users asking about LTTB and ASAP, which output a timevector as a result.

**Why do we feel this feature is ready to be stable**

This feature has been available in the experimental schema for over a year at this point.  It is also gating ASAP and LTTB stability, and we suspect that those features are otherwise ready for stabilization and will need to be to start seeing more adoption.

**Is there any known further work needed on this feature after stabilization**

Timevectors have tremendous potential to grow and expand and become a more user friendly way of expressing SQL queries.  We almost certainly want to support more value types than just float, and will likely want to support multiple values per timestamp.  We also may want to look at increasing the maximum size of timevectors to support 10s or 100s of millions of values.

**Are there any compatibility concerns that may arise during future work on this feature**

The current format will not be able to support multiple types or columns, but it does contain the complete data for a time, value pair of rows (including any null values).  We should be able to create new versions of the format and cleanly upgrade this version at deserialization time.

### Feature History
- Experimental release version: 1.0.0
- Last version modifying on-disk format: 1.8.0
- Target stabilization version: 1.9.0

### Stabilization checklist:
- [ ] Ensure tests exist for all public API
- [ ] Ensure API documentation exists and is accurate
- [ ] Remove `toolkit_experimental` tags and update test usages
- [ ] Add internal schema tags for non-public portions
- [ ] Add arrow operators for accessors if applicable
- [ ] Ensure arrow operators have test coverage
- [ ] If present, ensure `combine` and `rollup` are tested
- [ ] Add serialization tests for on disk format
- [ ] Add upgrade tests
- [ ] Add continuous aggregate test
- [ ] Add feature level documentation

PR #516, opened 2022-09-01
Drop experimental tag from stabilized feature documentation.

epgts:
Closes issue #220.

### DIFF
--- a/docs/timeseries.md
+++ b/docs/timeseries.md
@@ -1,4 +1,4 @@
-# Timevector [<sup><mark>experimental</mark></sup>](/docs/README.md#tag-notes)
+# Timevector
 
 > [Description](#timevector-description)<br>
 > [Timevector Pipelines](#timevector-pipelines)<br>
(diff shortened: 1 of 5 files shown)


docs/timeseries.md
```diff
@@ -1,4 +1,4 @@
-# Timevector [<sup><mark>experimental</mark></sup>](/docs/README.md#tag-notes)
```
rtwalker, line 1: Is this stabilized? https://github.com/timescale/timescaledb-toolkit/issues/483

WireBaron: Timevectors are stabilized, timevector pipelines are not

docs/timeseries_pipeline_elements.md
```diff
@@ -1,4 +1,4 @@
-# Timevector Pipelines [<sup><mark>experimental</mark></sup>](/docs/README.md#tag-notes)
```
WireBaron, line 1: This is a bit different than the pipelined accessors that we stabilized in 1.10.1.  I think this should still be marked experimental.  Hmm, taking a look this doc also looks terribly out of date as it only has three pipeline elements (I think we're closer to dozen at this point).

WireBaron (approved):

PR #521, opened 2022-09-06
Updating Toolkit To Start Using Cargo Fmt

thatzopoulos:
Part 1 of updating toolkit to use cargo fmt. 
Once this is merged, we will need to store the merge commit from this PR in a file in the repo so that we can set up git blame to ignore this commit:

Create file .git-blame-ignore-revs:
```
# Example Commit
864343aaa7d7e72d8e336d8bb8778d4f49846af9
```

```git config blame.ignoreRevsFile .git-blame-ignore-revs```

```
```


thatzopoulos:
CI is currently failing because Clippy believes that a lot of instances of `use pgx::*;` are unused imports. I checked a couple of the files and they seem to be using pgx.

PR #522, opened 2022-09-06
Move update-tester tests to markdown files

thatzopoulos:
Changes to the update-tester:

- Hardcoded tests are now read in from markdown files, similar to sql-doctester
- Update-tester will now set all timezones to UTC before running sql scripts
- Ability to set a min_toolkit_version that tests can run on


sql/original_update_tests.md
```diff
@@ -0,0 +1,48 @@
+# Original Update Tests
+This file is a port of the original hardcoded update-tester tests to markdown, similar to how sql-doctester works.
```
epgts, line 2: I'm not a fan of using markdown, as we aren't using any other features of markdown.  These files aren't intended to host documentation.  Seems confusing.

sql/original_update_tests.md
```diff
@@ -0,0 +1,48 @@
...
+
+These markdown files are parsed in the same way as sql-doctester would with the addition of two fields: `creation` and `validation`. (Names are not final).
+
+Tests marked `creation` will be ran during the object creation step, tests marked validation will be ran during the validation step.
```
epgts, line 6: These two lines are examples of what I mean! :)

It may be convenient for you to place these comments here, but:
- They could as easily have been SQL comments
- The long-term intent is unclear.

sql/original_update_tests.md
```diff
@@ -0,0 +1,48 @@
...
+ 
+
+ 
+## Object Creation Step
+ 
+Currently, `SET TIME ZONE 'UTC';` will cause tests to fail if used in the `validation` step, so it should used in `creation` if needed.
+```sql,creation
+SET TIME ZONE 'UTC';
```
epgts, line 14: Won't we always want this?  I think update-tester should take care of this across the board.

sql/original_update_tests.md
```diff
@@ -0,0 +1,48 @@
...
+
+CREATE TABLE test_data(ts timestamptz, val DOUBLE PRECISION);
```
epgts, line 16: Normally, tests are fully isolated from each other.  But our very goal here is persistence.

Is it feasible to create a separate database for each test?  Or is that too expensive?

If it's too expensive, how can we prevent tests from stepping on each other?  Can we require that each test creates only tables prefixed with the test's name?  So, update-tester would require after running this file that only tables whose names begin with 'original_update_tests' were created.

This might be a bit of work, so we might want to consider this for a follow-up rather than doing it here.

sql/original_update_tests.md
```diff
@@ -0,0 +1,48 @@
+# Original Update Tests
```
epgts, line 1: I don't see anything here about toolkit versions.  It's OK not to implement that here, but I think we at least need an idea of what that might look like before proceeding.

tools/update-tester/src/parser.rs
```diff
@@ -0,0 +1,447 @@
+use std::{collections::HashMap, ffi::OsStr, fs};
```
epgts, line 1: Oof :(

epgts (commented):
I haven't read the Rust yet.

tools/update-tester/src/parser.rs
```diff
@@ -0,0 +1,447 @@
...
+        if !entry.file_type().is_file() {
+            continue;
+        }
+
+        if entry.path().extension() != Some(OsStr::new("md")) {
+            continue;
+        }
+
+        let realpath;
+        let path = if entry.file_type().is_symlink() {
```
epgts, line 50: This is superfluous:  just pass `entry.path()` directly to `read_to_string`.

sql/original_update_tests.md
thatzopoulos: I'm currently writing min_toolkit_version code, I pushed up what I had so far just in case it took longer to implement than planned. It should be up soon.

sql/original_update_tests.md
thatzopoulos: Sql-doctester seems to run fine and I believe that it also creates a separate database for each test file. We can prevent tests from stepping on each other by porting over the transactional/non-transactional functions from sql-doctester which would allow us to rollback any changes after a test runs.
I'm also ok with the table prefix idea.

tools/update-tester/src/parser.rs
thatzopoulos: 👍 Would you like me to also update this line in sql-doctester?

sql/original_update_tests.md
thatzopoulos: I am not sure if we always want it but if we do I can definitely have update-tester handle it

sql/original_update_tests.md
epgts: OK I just wanted to make sure we planned for it.  No rush on pushing that.

It's great to see this progress!

Thanks.

sql/original_update_tests.md
epgts: If we don't, tests will vary as we all have different local timezone settings :)

If we had some test that needed to test something in some other specific timezone, it should be able to override, as i think the last SET wins.  And would that leak?  Maybe update-tester needs to do this before running every script?

thatzopoulos: Changing code to set timezone to UTC before running any scripts.

tools/update-tester/src/testrunner.rs
```diff
@@ -179,99 +227,143 @@ struct TestClient(Client);
...
+    }
+
+    fn create_test_objects_from_file(
+        &mut self,
+        root_config: ConnectionConfig<'_>,
+        current_toolkit_version: String,
+    ) -> Vec<(Test, Result<(), TestError>)> {
+        let all_tests = parser::extract_tests("sql");
+        // Hack to match previous versions of toolkit that don't conform to Semver.
+        let current_toolkit_semver = match current_toolkit_version.as_str() {
```
epgts, line 255: [After live discussion, we think I was confused when I wrote this, because actually we want the tests even for old versions to come FROM THE NEW CODE.  Not only would running old `create-test-objects` be pointless, but running new `create-test-objects` lets us add new tests for old versions.
So ignore what's below for now.
Recording for posterity.]

Is this necessary?  Two questions make me think it may not be:

1. I want to drop all releases older than 1.9 from `upgradeable_from` for 1.12 and all older than 1.10 for 1.13.
2. Will we be able to run 1.11's validate-test-objects against 1.10's create-test?  If not, s/1.10/1.11/ in #1 :)

I don't think I've spoken of my push to drop <1.9 from upgradeable_from so let me get down an early draft here:

We can't `create-test-objects` on any release older than 1.10, which means our current test automation plan won't support testing upgrading from any release older than that without having two automation schemes at twice the price (of both build AND maintain!), which means we can't call "release automation" done until we drop the old versions.

tools/update-tester/src/testrunner.rs
```diff
@@ -362,3 +454,240 @@ fn get_values(query_results: Vec<SimpleQueryMessage>) -> QueryValues {
         })
         .collect()
 }
+
+// Functions below this line are originally from sql-doctester/src/runner.rs
```
epgts, line 458: We should consider isolating duplicated functionality to separate files.  Is that too painful?

tools/update-tester/src/main.rs
```diff
@@ -225,25 +278,57 @@ fn main() {
                 password: validate_test_object_matches.value_of("PASSWORD"),
                 database: validate_test_object_matches.value_of("DB"),
             };
+            let mut num_errors = 0;
+            let stdout = io::stdout();
+            let mut out = stdout.lock();
+            let on_error = |test: parser::Test, error: testrunner::TestError| {
+                if num_errors == 0 {
+                    let _ = writeln!(&mut out, "{}\n", "Tests Failed".bold().red());
```
epgts, line 286: I get this after reread but it's a tad confusing and I don't see the value of this header.  Delete?

epgts (commented):
Most of these (as discussed) are notes for the future.  I think this is basically ready to go.

Thanks!

tools/update-tester/src/main.rs
```diff
@@ -149,6 +150,8 @@ fn main() {
                     .long("database")
                     .takes_value(true)
             )
+
+            .arg(Arg::new("ROOT_DIR").takes_value(true))
```
epgts, line 154: It would be better to require the current directory be the root of the project rather than making the caller pass that directory as argument.

`open(extension/toolkit.control).unwrap()`

tools/update-tester/src/parser.rs
```diff
@@ -0,0 +1,472 @@
...
+            }
+            _ => {}
+        }
+    }
+
+    info
+}
+
+#[cfg(test)]
+mod test {
```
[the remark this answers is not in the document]
thatzopoulos: I've updated it to use `tests`. I thought that we were keeping them in the same file because they are unit tests? https://doc.rust-lang.org/book/ch11-03-test-organization.html#unit-tests
It seems like counter-agg is the only crate in toolkit that uses a `tests.rs` file right now.

tools/update-tester/src/main.rs
thatzopoulos: Should I make that a requirement for all 3 commands that use root_dir?
I could also change that line to           ```.arg(Arg::new("ROOT_DIR").takes_value(true).default_value("."))``` and then we could keep the flexibility of calling it from other dirs and wouldn't have to update any scripts that call it.

sql/original_update_tests.md
```diff
@@ -0,0 +1,41 @@
...
+        time_weight('locf', ts, val) AS twa,
+        uddsketch(100, 0.001, val) as udd,
+        tdigest(100, val) as tdig,
+        stats_agg(val) as stats
+    FROM test_data;
+```
+
+## Validation Step
+
+```sql,validation,min-toolkit-version=1.4.0
```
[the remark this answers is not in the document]
thatzopoulos: It does not have any significance, I have removed it.

tools/update-tester/src/main.rs
```diff
@@ -211,11 +237,38 @@ fn main() {
                 database: create_test_object_matches.value_of("DB"),
             };
 
-            let res = try_create_objects(&connection_config);
+            let mut num_errors = 0;
+            let stdout = io::stdout();
+            let mut out = stdout.lock();
```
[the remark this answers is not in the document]
thatzopoulos: I've updated the release automation plan to include digging into yanking this out

tools/update-tester/src/testrunner.rs
thatzopoulos: I've updated the release automation plan to include next steps for what could be done to yank out all of the duplicated functionality. Even though the functions are the same, they are using slightly differing versions of Test and TestError. We would need to update parts of sql-doctester and update-tester in order to get them working nicely together with this code which shouldn't be hard but might be worth doing in its own PR.

tools/update-tester/src/main.rs
thatzopoulos: Updated release automation plan to include this in future work.

tools/update-tester/src/parser.rs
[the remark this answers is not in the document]
thatzopoulos: Updated release automation plan to include this in future work.

sql/original_update_tests.md
thatzopoulos: Updated release automation plan to include this in future work.

epgts (approved):

epgts (approved):
one last question:  is it all rustfmted? :)

thatzopoulos:
> 

It seems to all be rustfmted for me locally

rtwalker:
> It seems to all be rustfmted for me locally

same!
```sh
~/t/timescaledb-toolkit (th/update-tester-test-reformat)> cargo fmt --check
~/t/timescaledb-toolkit (th/update-tester-test-reformat)> 
```

PR #524, opened 2022-09-06
Updating Toolkit To Start Using Cargo Fmt

thatzopoulos:
Part 1 of updating toolkit to use cargo fmt. 
Once this is merged, we will need to store the merge commit from this PR in a file in the repo so that we can set up git blame to ignore this commit:

Create file .git-blame-ignore-revs:
```
# Example Commit
864343aaa7d7e72d8e336d8bb8778d4f49846af9
```

```git config blame.ignoreRevsFile .git-blame-ignore-revs```

```
```

### CODE REVIEW

File: crates/hyperloglogplusplus/src/registers.rs [hunk 1]
```diff
@@ -251,7 +254,7 @@ mod test {
     fn quick_test(exp: u8, ops: Vec<(usize, u8)>) -> quickcheck::TestResult {
         use quickcheck::TestResult;
         use std::cmp::max;
-        if exp < 4 || exp > 16 {
+        if !(4..=16).contains(&exp) {
```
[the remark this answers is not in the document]
thatzopoulos: This doesn't seem much clearer to me, I am ok with keeping or reverting this specific change

thatzopoulos (commented):
I went through all of the changes and made a comment on any change that was not just a formatting change to make reviewing a bit easier.

crates/asap/src/lib.rs
```diff
@@ -304,34 +317,227 @@ mod tests {
...
...13.8,10.1,7.3,5.7,5.5,5.7,5.1,10.0,11.2,15.3,17.4,16.8,16.3,11.7,6.6,5.8,4.2,5.3,7.4,7.6,11.3,16.2,15.9,15.6,12.9,9.6,5.7,1.2,3.9,3.7,4.1,6.8,10.1,14.0,16.3,14.8,14.1,9.4,8.5,5.5,2.7,3.4,6.6,9.6,13.4,14.4,16.8,15.8,10.7,8.8,4.1,2.8,3.3,4.3,5.6,7.3,12.6,14.4,15.5,16.2,13.8,9.7,8.5,6.9,2.9,2.6,5.8,7.6,11.2,13.4,14.2,14.6,12.7,11.9,6.9,6.8,2.6,1.2,3.2,9.3,9.7,13.8,17.7,18.1,14.2,9.2,7.0,5.4,3.6,0.2,6.2,6.9,11.7,13.1,15.8,13.5,14.3,9.4,6.0,5.7,5.5,5.3,9.2,8.9,10.3,15.2,16.3,15.4,12.5,10.8,6.4,4.5,3.4,4.7,3.7,7.4,11.1,14.1,15.9,15.8,15.1,10.8,6.4,4.7,1.6,4.4,7.3,9.4,12.8,15.2,17.3,17.2,14.9,12.6,7.1,6.0,3.8,4.1,6.4,8.9,12.8,16.1,15.1,15.0,13.1,10.3,7.3,3.9,3.9,6.9,8.2,10.0,11.0,14.4,15.2,15.4,15.2,10.9,6.0,2.2,4.3,4.4,2.8,7.7,10.3,13.7,15.1,14.5,12.6,10.4,5.5,1.8,2.1,0.7,6.0,8.7,10.6,14.9,15.2,14.3,12.9,11.1,8.2,2.6,3.4,4.5,4.3,8.7,13.3,13.8,16.1,15.5,14.1,8.9,7.4,3.6,3.3,3.1,5.2,8.0,11.7,14.7,14.0,14.9,12.3,11.0,4.5,4.7,2.9,5.7,6.5,7.2,11.0,15.4,15.0,14.7,13.8,10.1,5.6,5.5,4.5,5.4,7.0,7.7,10.4,14.0,16.7,15.7,13.5,10.8,5.4,4.2,4.4,1.9,6.3,8.1,9.8,14.8,15.0,15.4,13.9,12.5,6.5,3.0,5.5,1.0,3.3,7.4,11.2,13.9,16.8,16.4,13.9,13.0,5.4,3.3,3.7,2.9,3.7,6.7,13.0,16.4,15.2,16.0,14.4,10.7,7.8,4.3];
```
Smittyvb, line 334: rustfmt expanded the declaration to be 190 lines long here. Ideally this would be in a separate file and included.

rtwalker, line 334: I think this might be the only one I don't like.. I don't think anyone is actually reading this data and a 200 line wall of numbers doesn't seem obviously better than a hilariously long single line.

crates/counter-agg/src/lib.rs
```diff
@@ -276,14 +281,14 @@ impl MetricSummary {
...
+    pub fn prometheus_rate(&self) -> Result<Option<f64>, CounterError> {
+        let delta = self.prometheus_delta()?;
         if delta.is_none() {
             return Ok(None);
         }
         let delta = delta.unwrap();
-        let bounds = self.bounds.unwrap() ; // if we got through delta without error then we have bounds
-        /* bounds stores [L,H), but Prom takes the duration using the inclusive range [L, H-1ms]. So subtract an extra ms from the duration*/
+        let bounds = self.bounds.unwrap(); // if we got through delta without error then we have bounds
+                                           /* bounds stores [L,H), but Prom takes the duration using the inclusive range [L, H-1ms]. So subtract an extra ms from the duration*/
```
rtwalker, line 291: I don't understand why this whitespace space was added? I think this comment refers to the line below; does making this a `//` comment help keep it aligned?

extension/src/aggregate_builder_tests.rs
```diff
@@ -97,12 +88,14 @@ mod tests {
     #[pg_test]
     fn test_anything_in_experimental_and_returns_first() {
         Spi::execute(|client| {
-            let output = client.select(
-                "SELECT toolkit_experimental.anything(val) \
+            let output = client
+                .select(
+                    "SELECT toolkit_experimental.anything(val) \
```
rtwalker, line 93: I don't know if it's worth doing (in this PR or at all) but this is one of SEVERAL places where `rustfmt` makes the SQL statement inside the string have a peculiar alignment.

rtwalker (approved):
GitHub says I've viewed 49/59 files but I think I have to take a break for a little bit. Will pick up again later today.

To go along with a few minor comments I had, I'd like to very timidly propose `max_width = 120`. There's several places where I feel like we get bitten by line length (of course this is a symptom of using formatting tools and it will never be perfect).

Approving because I don't think anything I've seen is worth holding up the PR over

crates/asap/src/lib.rs
thatzopoulos: Undid this change and added a `rustfmt::skip`

thatzopoulos: Undid this change and added a `rustfmt::skip`

crates/counter-agg/src/lib.rs
thatzopoulos: Swapping to `//` didnt' keep it aligned but moving it to the end of the line that its a comment for did

extension/src/aggregate_builder_tests.rs
epgts: I think the answer is to use raw string literals and align manually.  We have some of those already, as well as some use of raw that still doesn't have clean alignment.

I think we should isolate rustfmt changes to one commit, so if Thomas does feel like doing anything about this right now, let's do it before or after this.

But I think it can just wait.

extension/src/lib.rs
```diff
@@ -1,35 +1,37 @@
-#![allow(clippy::modulo_one)]   // flat_serialize! alignment checks hit this for any single byte field (of which all pg_types! have two by default)
-#![allow(clippy::extra_unused_lifetimes)]  // some disagreement between clippy and the rust compiler about when lifetime are and are not needed
-#![allow(clippy::not_unsafe_ptr_arg_deref)]  // every function calling in_aggregate_context should be unsafe
+#![allow(clippy::modulo_one)]
+// flat_serialize! alignment checks hit this for any single byte field (of which all pg_types! have two by default)
```
epgts, line 2: All these comments now apply to the wrong line.

extension/src/stats_agg.rs
```diff
@@ -209,51 +191,50 @@ pub fn stats2d_trans_inner(
...
                 (None, _) => None,
                 (_, None) => None,
-                (Some(y), Some(x)) => Some(XYPair{y, x})
+                (Some(y), Some(x)) => Some(XYPair { y, x }),
             };
             match (state, val) {
-                (None, None) => Some(StatsSummary2D::from_internal(InternalStatsSummary2D::new()).into()), // return an empty one from the trans function because otherwise it breaks in the window context
+                (None, None) => {
+                    Some(StatsSummary2D::from_internal(InternalStatsSummary2D::new()).into())
+                } // return an empty one from the trans function because otherwise it breaks in the window context
```
rtwalker, line 199: nit: could turn this into 
```rust
                    ...
                    // return an empty one from the trans function because otherwise it breaks in the window context
                    Some(StatsSummary2D::from_internal(InternalStatsSummary2D::new()).into())
                 } 
```
if it's not too much trouble

rtwalker (commented):
Ok I've at least "glanced" at everything now. Seems OK to me!

epgts (approved):
My eyes are bleeding now, but let's do it!

extension/src/aggregate_builder_tests.rs
epgts: Along these lines, I suggest NOT squashing this, but keeping the three commits you have now.

extension/src/lib.rs
thatzopoulos: Added to reviewer comments commit.

extension/src/stats_agg.rs
thatzopoulos: Added to reviewer comments commit.

Issue #455, opened 2022-06-23
Integral function


mayku00:
Hi, I asked in the slack about how to get the function integral (like in influxDB)

The answer:
“Time weighted average def does that! Sounds like we should be exposing that! can you create an issue here: https://github.com/timescale/timescaledb-toolkit/issues? I think we have what we need in time_weight but just need to expose it the right way, so probably would be relatively quick to implement, and will just be an accessor, so if you want to, say, create continuous aggregates you can use the time_weight aggregate and when we make this accessor, you'll just be able to use it on that.”

so open is!!

jerryxwu:
@mayku00 thank you for the feature request! I think I have a general idea of what you are asking for. It would be great to learn your specific use case for this function. It can help us design the function better.

mayku00:
Hi! Sure.
I have an hypertable with 3 colums: timestamp, upload_bitrate, download_bitrate. I use the table to save the bitrate that I get from a modem every x seconds.

I use the table as the historical register for the bitrate, but also I need to calculate the volume consumed por a period of time.

As I have a line/curve of values ordered by the timestamp, I calculate the volume as the are under this line/curve:

On Influx I was using: https://docs.influxdata.com/flux/v0.x/stdlib/universe/integral/

I asked for a similar function

KR

davidkohn88:
Thanks! Leaving as much context here is awesome because the links to Slack eventually break because we don't have full history there! So feel free to edit or add more if you think of anything. But I think `time_weight('Linear', time, value) -> integral(unit, start_time, end_time)` would be a good start for what you're looking for here where unit is the unit of time to use to calculate the integral, all of those could have defaults, ('1s', NULL, NULL), where null values would just use the starting and ending times seen in the data set.

jerryxwu:
Having integral built on top of `time_weight` definitely makes sense. A few ideas and comments:
1. Users wanting to compute integral are more likely familiar with numeric integration lingo, so we should document 'Linear' means [trapezoidal method](https://sites.und.edu/timothy.prescott/apex/web/apex.Ch8.S7.html). Even better, allowing `time_weight` to accept `trapezoidal` as an alternative to `Linear`.
2. It's unlikely that `integral` can accept `start_time` and `end_time`, only `unit` makes sense
3. `interpolated_integral(...)` analogous to `interpolated_average` should also be implemented

jerryxwu:
[...]

PR #526, opened 2022-09-07
Add integral function for time_weight

Smittyvb:
Implements functionality requested in #455:
- Adds an experimental `integral(tws[, unit]) -> float8` function (unit defaults to `'second'`)
- Adds an experimental `interpolated_integral(tws, start, interval, prev, next[, unit]) -> float8` function (unit defaults to `'second'`)
- Adds experimental arrow functions for `integral`/`interpolated_integral`
- Makes `trapezoidal` an alias for `linear` in the `time_weight` function

The `unit` parameter to `integral`/`interpolated_integral` is a string that specifies what time unit to use for the returned `f64`. It can be any fixed-duration unit that PostgreSQL allows in a `interval`. An alternative would be to have an `interval` be passed as the unit instead, but `interval`s can have variable length units (days and months) which might not be wanted. Also, `integral(..., 'hour')` is clearer than `integral(..., '1 hour'::interval)`.

### DIFF
--- a/extension/src/accessors.rs
+++ b/extension/src/accessors.rs
@@ -518,3 +518,31 @@ pub fn accessor_unnest() -> AccessorUnnest<'static> {
         }
     }
 }
+
+#[pg_schema]
+pub mod toolkit_experimental {
+    use super::*;
+
+    pg_type! {
+        #[derive(Debug)]
+        struct AccessorIntegral<'input> {
+            len: u32,
+            bytes: [u8; self.len],
+        }
+    }
+
+    // FIXME string IO
+    ron_inout_funcs!(AccessorIntegral);
+
+    #[pg_extern(immutable, parallel_safe, name = "integral")]
+    pub fn accessor_integral(unit: default!(&str, "'second'")) -> AccessorIntegral<'static> {
+        unsafe {
+            flatten! {
+                AccessorIntegral {
+                    len: unit.len().try_into().unwrap(),
+                    bytes: unit.as_bytes().into(),
+                }
+            }
+        }
+    }
+}
--- a/extension/src/time_weighted_average.rs
+++ b/extension/src/time_weighted_average.rs
@@ -602,6 +602,10 @@ mod tests {
             FROM test";
             // arrow syntax should be the same
             assert!((select_one!(client, stmt, f64) - 25500000000.00).abs() < f64::EPSILON);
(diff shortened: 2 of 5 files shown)


extension/src/time_weighted_average.rs
```diff
@@ -401,6 +402,25 @@ pub fn arrow_time_weighted_average_average(
...
+        accessor: AccessorIntegral,
+    ) -> Option<f64> {
+        time_weighted_average_integral(
+            tws,
+            String::from_utf8_lossy(accessor.bytes.as_slice()).to_string(),
+        )
+    }
+}
+
+pub use toolkit_experimental::*;
```
rtwalker, line 422: I *think* that we've typically said that arrow functions are considered experimental as long as one of its arguments is experimental, i.e. you might not need the surrounding `mod toolkit_experimental {}`

extension/src/time_weighted_average.rs
```diff
@@ -401,6 +402,25 @@ pub fn arrow_time_weighted_average_average(
...
 
+mod toolkit_experimental {
+    use super::*;
+    use crate::accessors::toolkit_experimental::*;
+
+    #[pg_operator(immutable, parallel_safe)]
+    #[opname(->)]
+    pub fn arrow_time_weighted_average_integral(
+        tws: Option<TimeWeightSummary>,
+        accessor: AccessorIntegral,
```
rtwalker, line 413: I think I'd prefer to see `accessor: toolkit_experimental::AcessorIntegral`

extension/src/accessors.rs
```diff
@@ -518,3 +518,32 @@ pub fn accessor_unnest() -> AccessorUnnest<'static> {
...
+            flatten! {
+                AccessorIntegral {
+                    len: unit.len().try_into().unwrap(),
+                    bytes: unit.as_bytes().into(),
+                }
+            }
+        }
+    }
+}
+pub use toolkit_experimental::*;
```
rtwalker, line 549: Do we need this line? I honestly don't know for sure, but I don't think it was there when we'd previously had experimental accessors.

rtwalker (approved):
Thanks for the changes! Looks good to me though I'd wait for someone else to weigh in :)

davidkohn88 (approved):
LGTM!

WireBaron (approved):

PR #532, opened 2022-09-12
Adding git-blame-ignore-revs file

thatzopoulos:
Part 2 of updating toolkit to use `cargo fmt` by uploading `.git-blame-ignore-revs` with the `cargo fmt` commits in the file.

Part 1 can be found here: https://github.com/timescale/timescaledb-toolkit/pull/524

```
git config blame.ignoreRevsFile .git-blame-ignore-revs
```
```
```

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,3 @@
+# Merge and parent commit for cargo fmt changes
+b7433344f90b142094e73e84c332385498db9335
+8b50127c9e4bad1696a68a800ce1ef019cf6fc3c


thatzopoulos:
I am not totally sure if I actually need both commits in this file, its possible just one of the two is enough.

Smittyvb:
It looks like this successfully altered GitHub's blame view: [before](https://github.com/timescale/timescaledb-toolkit/blame/main/crates/encodings/src/lib.rs#L49) and [after](https://github.com/timescale/timescaledb-toolkit/blame/th/add-git-blame-ignore-revs-file/crates/encodings/src/lib.rs#L49). There are still a few lines attributed to ignored revisions, I think because there is not previous commit to attribute them to.

rtwalker:
> There are still a few lines attributed to ignored revisions, I think because there is not previous commit to attribute them to.

Nice catch! I hadn't considered this beforehand, but I suppose it makes sense. Nevertheless, I feel like this accomplishes our goal of the wholesale `rustfmt` not "ruining" the blaming

rtwalker (approved):

Smittyvb (approved):

WireBaron (approved):

PR #533, opened 2022-09-13
Verify `cargo fmt` in CI: part 1

Smittyvb:
Updates the Docker image so that later the CI will be able to verify everything is formatted with `rustfmt`.


Smittyvb:
The "Build CI Image" failure can be ignored, it always tries to login to Docker Hub but fails because some PRs don't have access to secrets.

rtwalker (approved):

epgts:
> The "Build CI Image" failure can be ignored, it always tries to login to Docker Hub but fails because some PRs don't have access to secrets. (and bors doesn't treat it as a mandatory check)

Are we sure it's not mandatory?

epgts (approved):

PR #535, opened 2022-09-15
Make tests pass in Canadian locales

Smittyvb:
Currently various date-related tests fail on my machine that's in the Canadian English (en_CA) locale. Postgres date parsing is locale-dependent, so this PR makes the dates parsed in tests use a format that is parsed the same in (hopefully) all locales.

It would be better to set the locale of the database to `en_US` and write all dates US-style, but SETting the locale parameter doesn't seem to have an impact on date parsing for me. Ideally pgx should set the database locale when creating a test database to a consistent value (since some locale-related values [can't be changed after database creation](https://www.postgresql.org/docs/current/locale.html#id-1.6.11.3.6) (not sure if date parsing is one of them)).


epgts (approved):
Thanks!

PR #536, opened 2022-09-15
Document requirement to use same compiler for cargo-pgx and Toolkit

Smittyvb:
The same compiler must be used to compile cargo-pgx and Toolkit, or bad things (undefined behaviour) might happen. See https://github.com/tcdi/pgx/pull/687 for why this is the case.

This PR updates the installation documentation in the README to indicate this. I split the command in two since you only need to reinstall cargo-pgx, there's no need to re-run `cargo pgx init`.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -49,10 +49,20 @@ For Ubuntu you can follow the [postgres install instructions](https://www.postgr
 ```bash
 sudo apt-get install make gcc pkg-config clang postgresql-server-dev-14 libssl-dev
 ```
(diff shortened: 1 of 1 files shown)


rtwalker:
> I split the command in two since you only need to reinstall cargo-pgx, there's no need to re-run `cargo pgx init`.

Considering this, do you want to make them two separate code blocks with the blurb about rust and cargo-pgx in between? Could also add something explaining that "there's no need to re-run cargo pgx init" after the second block for good measure

rtwalker (approved):
Thanks!

PR #537, opened 2022-09-15
Verify `cargo fmt` in CI

epgts:


epgts:
The "Verify formatting" step takes TWO seconds even with adding the component.  I've forgotten how to count that low.

rtwalker (approved):

Smittyvb (approved):
I was going to suggest just creating a replacement bot user, but that's not worth it to save 2 seconds.

.github/workflows/clippy_rustfmt.yml
```diff
@@ -13,7 +13,7 @@ on:
 
 jobs:
   clippy:
-    name: Clippy Test
+    name: Clippy/rustfmt Test
```
rtwalker, line 16: maybe just `name: Test`?

Screen Shot 2022-09-15 at 4 56 41 PM

bors.toml
```diff
@@ -1,5 +1,5 @@
 status = [
-    "Clippy Test",
+    "Clippy/rustfmt Test",
```
Smittyvb, line 2: bors is waiting forever for this pull request since it checks if CI passes based on this file, which now uses an outdated name for the job.

epgts: > bors is waiting forever for this pull request since it checks if CI passes based on this file, which now uses an outdated name for the job.

Brilliant!

epgts: Hmmm bors.toml only lists basename and just "Test" would not be accurate, so I discarded that commit.  Oh well!  Maybe there's something smarter we can do some other time.

Thanks.

epgts: Probably we should move clippy+rustfmt into ci.yml - I'm not sure there's any reason to separate them.  Again:  later.

epgts:
> I was going to suggest just creating a replacement bot user, but that's not worth it to save 2 seconds.

Once Josh is back, we'll fix the one we have and undo this hack.  Adding components in two places is complexity we don't want, and we do need to be able to update the image.

Thanks!

Issue #419, opened 2022-05-12
Release automation phase 1


jerryxwu:
Currently, our release procedure produces a version tag. Ideally, we want all the final packages (docker etc) built and tested. Since we don't currently own the docker images (we may not own them in the future either), as a first step, we want to achieve the following:

1. Build the deb and rpm packages
2. Run all the tests that we have, if we don't have any tests yet, define a small smoke test suite for final verification
3. Publish the packages in the right places

epgts:
Currently I test binaries manually by following our official installation instructions on CentOS 7 (well, Rocky 8.6 this time) and Ubuntu 20.04 and running a manual smoke test, usually by copying something out of a doc test or pgx test.

For automated testing of binaries, I left a comment at line 21 here: https://github.com/timescale/release-build-scripts/blob/90da47dc17ba5bcf677e6ecfaa85690a83ec9fcb/debian/docker-build-toolkit.sh#L21

What I left out of that comment is moving the post-install run to after we build the deb and install it with `dpkg -i`. After test passes, remove it again with `dpkg -r`.

Currently we:

```
  cargo run --manifest-path tools/post-install/Cargo.toml -- --dir target/release/timescaledb_toolkit-pg${PG_MAJOR}
```

I mentioned this morning that `--dir /usr` ought to work but doesn't, but actually I noticed that's not how we do it from `tools/build` anyway!

So the real fix might be as simple as changing that to

```
  cargo run --manifest-path tools/post-install/Cargo.toml /bin/pg_config
```

Possibly we may need to execute other commands from https://docs.timescale.com/timescaledb/latest/how-to-guides/hyperfunctions/install-toolkit/#install-toolkit-on-ubuntu-and-other-debian-based-systems in order for that to work (in particular, possibly `/usr/share/postgresql-common/pgdg/apt.postgresql.org.sh` and `systemctl restart postgresql`).

PR #540, opened 2022-09-19
consider: Stop supporting upgrade from 1.4 and 1.5 .

epgts:
We are automating testing upgrades of binaries, but these, our two oldest versions, are hard to automate because:

- We released no binaries for 1.4
- The deb for 1.5 was labeled 1.5.0

We could special case these, but requiring two upgrade steps (1st to 1.5.1 and 2nd to latest) is not unreasonable for users who are eight months behind.

For issue #419


WireBaron (approved):

epgts:
We'll include this in 1.15.0 .

WireBaron:
This PR is being supplanted by #743

PR #542, opened 2022-09-21
Make unused version number 0.0.0-unused

Smittyvb:
We don't use the version number in the extension's `Cargo.toml` anywhere, replace it with `0.0.0-unused`. pgx supports replacing `@CARGO_VERSION@` in control files but we don't use that in our control file.


epgts (commented):
> pgx supports replacing @CARGO_VERSION@ in control files but we don't use that in our control file.

Interesting.  Should we?  Would that allow us to have only one place to update the version number and that places is extension/Cargo.toml ?  If so, I think I like that.

WireBaron:
> > pgx supports replacing @CARGO_VERSION@ in control files but we don't use that in our control file.
> 
> Interesting. Should we? Would that allow us to have only one place to update the version number and that places is extension/Cargo.toml ? If so, I think I like that.

The problem is that we still have to update the control file to update the upgrade version, so I'm not sure it buys us that much.  Is there anywhere else we can make use of this `@CARGO_VERSION@`?  Let's discuss tomorrow.

Smittyvb:
I updated the PR to use `Cargo.toml` as the source for the version number, and updated update-tester to use the new source.

epgts (approved):

PR #545, opened 2022-09-27
Make interpolation errors more user friendly

Smittyvb:
Currently the error message for invalid interpolation bounds doesn't make it clear what the issue is, and doesn't say what the expected value is. This PR makes the bound checking assertions more clear

### DIFF
--- a/extension/src/time_weighted_average.rs
+++ b/extension/src/time_weighted_average.rs
@@ -51,7 +51,19 @@ impl<'input> TimeWeightSummary<'input> {
         prev: Option<TimeWeightSummary>,
         next: Option<TimeWeightSummary>,
     ) -> TimeWeightSummary<'static> {
-        assert!(interval_start <= self.first.ts && interval_start + interval_len > self.last.ts);
+        assert!(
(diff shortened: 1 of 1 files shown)


extension/src/time_weighted_average.rs
```diff
@@ -51,7 +51,19 @@ impl<'input> TimeWeightSummary<'input> {
         prev: Option<TimeWeightSummary>,
         next: Option<TimeWeightSummary>,
     ) -> TimeWeightSummary<'static> {
-        assert!(interval_start <= self.first.ts && interval_start + interval_len > self.last.ts);
+        assert!(
+            interval_start <= self.first.ts,
+            "Interval start ({}) must be before first timestamp ({})",
```
rtwalker, line 56: what do you think about 
```suggestion
            "Interval start ({}) must be at or before first timestamp ({})",
```
since we're doing a `<=` comparison?

rtwalker (approved):
LGTM! Might be worth dropping the github generated commit since it seems to be playing weirdly with our CI.

Smittyvb:
bors retry

Issue #382, opened 2022-04-04
Create arm64 packages in the repository


wtfuzz:
**Is your feature request related to a problem? Please describe.**
I'm using Timescale on `aarch64` AWS EC2 nodes, and the official repository only has `x86_64`  binary packages.

**Describe the solution you'd like**
It would be convenient to be able to install a binary package of TimescaleDB toolkit on ARM architectures.

**Describe alternatives you've considered**
Building the toolkit from source on the target architecture as a workaround

jerryxwu:
@wtfuzz thank you for the request. ARM64 support is something that we are looking closely at. I have couple questions for you. 
1. What Linux distro(s) are you using specifically? 
2. What kind(s) of binary packages are most interesting/useful to you?

wtfuzz:
> * What Linux distro(s) are you using specifically?

On our ARM platforms, I use:

* Debian Bullseye
* Ubuntu 20.04 LTS

> * What kind(s) of binary packages are most interesting/useful to you?

* .deb package of `timescaledb-toolkit-postgresql-14` for the `arm64` architecture

jfjoly:
@wtfuzz would a docker image for arm64 work for you?

epgts:
see also #212

jerryxwu:
@wtfuzz we've just released arm64 deb packages for Ubuntu (20.04 & 22.04) and Debian (10 & 11). Check them out. We'd love to hear any feedback you may have.

PR #546, opened 2022-09-28
Update README supported platforms section

rtwalker:
Figure we should remove this line about "in development" now that #382 is closed.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -30,10 +30,9 @@ All versions of the extension contain experimental features in the `toolkit_expe
 The engineering team regularly tests the extension on the following platforms:
 
 - x86_64-unknown-linux-gnu (Ubuntu Linux 20.04) (tested prior to every merge)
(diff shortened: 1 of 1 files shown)


Readme.md
```diff
@@ -33,8 +33,6 @@ The engineering team regularly tests the extension on the following platforms:
 - x86_64-apple-darwin (MacOS 12) (tested frequently on eng workstation)
 - aarch64-apple-darwin (MacOS 12) (tested frequently on eng workstation)
 
-aarch64-unknown-linux-gnu support is [in development](https://github.com/timescale/timescaledb-toolkit/issues/382).
-
```
epgts, line 37: Did you mean to remove it entirely?  It should be folded into the above list with the [in development] part removed.  Thanks!

epgts: I'd say list it second, something like:
- aarch64-unknown-linux-gnu (Ubuntu Linux 20.04) (tested at release time)

PR #549, opened 2022-09-29
Try using org-wide Docker Hub secrets

Smittyvb:
Not sure if this will work, but maybe `DOCKERHUB_*` are org-wide secrets and so using those will let us push a new CI image to Docker Hub? There's no way to find out without creating a PR.


Smittyvb:
Apparently these are not org-wide secrets.

PR #552, opened 2022-10-03
Binary update testing in CI

Smittyvb:
Adds support for binary update testing in GitHub Actions. #554 and #556 need to be merged first.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -167,8 +167,11 @@ jobs:
         key: ${{ runner.os }}-test-old-versions-1.12.0
         restore-keys: ${{ runner.os }}-test-old-versions-1.11.0
 
-    - name: Run Update Tests
-      run: su postgres -c 'sh tools/build -pg$PGVERSION -cargo-pgx /home/postgres/pgx/0.4/bin/cargo-pgx -cargo-pgx-old /home/postgres/pgx/0.2/bin/cargo-pgx test-updates 2>&1'
+    - name: Run binary update tests
+      # TODO We've always tested only one postgres version here.  We test them all at release
+      #  time anyway.  If we want to test them here, too, let's make it less slow first.
+      run: |
+        su postgres -c 'OS_NAME=debian OS_VERSION=11 tools/testbin -version no -bindir / -dbroot /tmp/db -pgport 28800 -pgversions 14 ci 2>&1'
 
     - name: Check tools/testbin
       run: sh tools/testbin check-control


.github/workflows/ci.yml
```diff
@@ -166,9 +166,13 @@ jobs:
         path: old-versions
         key: ${{ runner.os }}-test-old-versions-1.12.0
         restore-keys: ${{ runner.os }}-test-old-versions-1.11.0
+    
+    - name: Make postgres sudoer
+      run: |
+        echo 'postgres ALL=(ALL) NOPASSWD: ALL' >> /etc/sudoers
```
epgts, line 172: Ah!  Let me put that into the image.

.github/workflows/ci.yml
```diff
@@ -166,9 +166,13 @@ jobs:
...
 
-    - name: Run Update Tests
-      run: su postgres -c 'sh tools/build -pg$PGVERSION -cargo-pgx /home/postgres/pgx/0.4/bin/cargo-pgx -cargo-pgx-old /home/postgres/pgx/0.2/bin/cargo-pgx test-updates 2>&1'
+    - name: Run binary update tests
+      run: OS_NAME=ubuntu OS_VERSION=20.04 su postgres -c 'sh tools/testbin -bindir .. -version 1.11.0 -dbroot /tmp/db -pgport 28814 deb'
```
epgts, line 175: As I learned this afternoon, we actually run on Debian 10!

My near-term goal after finishing the current batch of release automation work is going to be extracting the rest of the Docker image building logic out of the packaging scripts and into the image-building script I pushed up today and (sadly) moving that script into the private repository so we can build arm64 images too, and then run both our CI here on that image and ALSO build packages on the same image.

Currently the packaging scripts build and discard a Docker image every time! :(

epgts (commented):
Thanks for starting this!

.github/workflows/ci.yml
epgts: I'm not sure testbin will work as is.  I think we need a third run mode (in addition to the existing 'deb' and 'rpm').

epgts: I think this is what we need here:  https://github.com/timescale/timescaledb-toolkit/pull/556/files#diff-018192a8c8e7dea0df3dbbfdc766bfad7a3017167956b9714f8653b5f7644f72R111

Smittyvb:
Currently this is blocked on #554 and #556 merged so that we can push a new CI Docker image with the changes needed for this PR.

tools/testbin
```diff
@@ -135,7 +135,7 @@ deb_start_test() {
     for PG_VERSION in $PG_VERSIONS; do
         select_pg $PG_VERSION
         deb=timescaledb-toolkit-postgresql-${PG_VERSION}=${EPOCH}${FROM_VERSION}~${OS_NAME}${OS_VERSION}
-        $nop sudo apt-get -qq install $deb
+        [ $1 = deb ] && $nop sudo apt-get -qq install $deb
```
epgts, line 138: apt-get install is idempotent so there should be no reason to special-case this; did you hit a problem?

.github/workflows/ci.yml
```diff
@@ -167,6 +167,15 @@ jobs:
         key: ${{ runner.os }}-test-old-versions-1.12.0
         restore-keys: ${{ runner.os }}-test-old-versions-1.11.0
 
+    - name: Install Toolkit
+      run: |
+        (
+          cd extension
+          cargo pgx install -c /usr/lib/postgresql/12/bin/pg_config
```
epgts, line 174: That's what https://github.com/timescale/timescaledb-toolkit/commit/1ffa4e9d02ed9dd1011f4200224ab4d4e535a8c8#diff-018192a8c8e7dea0df3dbbfdc766bfad7a3017167956b9714f8653b5f7644f72R167 is supposed to do - is it not working?

I did admit that was only lightly tested :)

We should fix it rather than trying to work around it, though.

tools/testbin
Smittyvb: In [this run](https://github.com/timescale/timescaledb-toolkit/actions/runs/3192423093/jobs/5209880700) (before I changed that line) I got an error:
```
+ sudo apt-get -qq install timescaledb-toolkit-postgresql-12=1.5.1~debian10
E: Version '1.5.1~debian10' for 'timescaledb-toolkit-postgresql-12' was not found
```

.github/workflows/ci.yml
Smittyvb: I didn't realize that `testbin` was supposed build Toolkit. I added that after seeing [this failed run](https://github.com/timescale/timescaledb-toolkit/actions/runs/3192484059/jobs/5210012286): it had an error since it couldn't find a control file for Toolkit. I think the problem is that `test_ci` calls `deb_start_test` which calls `start_test` which creates test objects before Toolkit is installed.

tools/testbin
epgts: That's because a few days ago when I wrote "we actually run on Debian 10!" I had mixed up bullseye and buster for the 50th time :)

We're actually on Debian 11.

.github/workflows/ci.yml
epgts: No, you made the line that installs the old version not run in 'ci' mode; that's why you found no toolkit installed in `start_test`.

.github/workflows/ci.yml
epgts: This should work:

```
su postgres -c 'OS_NAME=debian OS_VERSION=11 tools/testbin -version no -bindir / -dbroot /tmp/db -pgport 28800 ci 2>&1'
```

Although the chown bug-fix in PR #562 will need to go in first.

epgts: Oops, I forgot to suggest using -pgversions - this is what we want:

```
# TODO We've always tested only one postgres version here.  We test them all at release
#  time anyway.  If we want to test them here, too, let's make it less slow first.
su postgres -c 'OS_NAME=debian OS_VERSION=11 tools/testbin -pgversions 14 -version no -bindir / -dbroot /tmp/db -pgport 28800 ci 2>&1'
```

epgts (approved):

PR #553, opened 2022-10-03
Make tools/build default to pgx's configure pg_config locations.

epgts:
This is a much better default, such that the need for the -pgconfig override is reduced if not completely eliminated.

In particular, this unblocks the ci image improvements to use binary installations of postgresql rather than letting pgx build it (as the pgconfig path is different in that case).

### DIFF
--- a/tools/build
+++ b/tools/build
@@ -26,14 +26,7 @@ require_pg_version() {
 find_pg_config() {
     if [ -z "$pg_config" ]; then
         require_pg_version
-        full_version=$(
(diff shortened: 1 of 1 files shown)


epgts:
Is it just me, or does the bors check always get stuck now?

Smittyvb:
@epgts bors is currently a required check but doesn't report anything until it tries to merge the PR. I think it should be removed from the list of required checks in the branch protection rules.

Smittyvb (approved):

PR #554, opened 2022-10-03
Install postgresql and timescaledb from packages.

epgts:
This is much less wasteful of both maintainer and compute resources.

Extract sh script embedded into Dockerfile into a file and tidy up.
It's much easier to edit AND test when it's a standalone file!

Also add a README from Mats.


epgts:
PR #553 needs to go first, as this changes where pg_config lives and the build script didn't know how to find it there.

Smittyvb (approved):

PR #556, opened 2022-10-04
Add 'ci' mode to testbin .

epgts:
This is mostly the same as 'deb' but for use in our continuous
integration checks, where we are installing from source rather than
testing a new binary package.

### DIFF
--- a/tools/testbin
+++ b/tools/testbin
@@ -124,21 +124,21 @@ deb_init() {
 # Requires:
 # - FROM_VERSION
 deb_start_test() {
-        # We released 1.10.0-dev by accident.  We have to support upgrades
-        # from it (and we tested that at the time), but we pulled the deb, so
-        # we can't test it here.
-        [ $FROM_VERSION = 1.10.0-dev ] && return 1
-        cmp_version=`cmp_version $FROM_VERSION`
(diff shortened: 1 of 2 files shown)


tools/testbin
```diff
@@ -111,29 +111,40 @@ finish_test() {
...
+
         [ $cmp_version -ge $MIN_DEB_EPOCH ] && EPOCH=1:
         for PG_VERSION in $PG_VERSIONS; do
             select_pg $PG_VERSION
-            deb=timescaledb-toolkit-postgresql-${PG_VERSION}=${EPOCH}${from_version}~${OS_NAME}${OS_VERSION}
+            deb=timescaledb-toolkit-postgresql-${PG_VERSION}=${EPOCH}${FROM_VERSION}~${OS_NAME}${OS_VERSION}
             $nop sudo apt-get -qq install $deb
 
             PATH=/usr/lib/postgresql/$PG_VERSION/bin:$PATH start_test
         done
```
Smittyvb, line 141: This section is indented twice; it should be indented once.

Smittyvb (approved):

tools/testbin
epgts: I left that as is for clear diff on first review, but then forgot!  Fixed now.  Thanks.

PR #557, opened 2022-10-04
Remove testrunner.

epgts:
testrunner ran all our pgx tests by finding all functions installed into
a specified database in a namespace 'tests'.

No normal database would have our tests installed this way.  In fact,
the only obvious way is to run our extension tests.  But running our
extension tests after running our extension tests doesn't make sense.

I can imagine how this might have been useful once, but it seems to have
been forgotten and we no longer need it.


epgts:
Obviously not urgent - can wait for Brian's return :)

WireBaron (approved):

PR #558, opened 2022-10-04
Adding Time Vector tests to update tester

thatzopoulos:
Added Update Tester tests for time vector
Removed `break` from the match statement in the `validate_output` fn so that we can have multiple sql statements in a validation test.

### DIFF
--- a/tools/update-tester/src/testrunner.rs
+++ b/tools/update-tester/src/testrunner.rs
@@ -148,6 +148,7 @@ pub fn update_to_and_validate_new_toolkit_version<OnErr: FnMut(Test, TestError)>
     let mut test_client = connect_to(&test_config);
 
     test_client.set_timezone_utc();
+
     // get the currently installed version before updating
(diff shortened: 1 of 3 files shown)


tools/update-tester/src/testrunner.rs
```diff
@@ -339,6 +338,9 @@ impl TestClient {
             .into_iter()
             .flat_map(|tests| {
                 let mut client = connect_to(&root_config).0;
+                client
```
thatzopoulos, line 341: Moving the code that sets the timezone to here instead of line 150 seems to have fixed the issue on my machine. It could also have been moved to run_test(), but then it would be called on every sql query and that seemed excessive.

Smittyvb (approved):
At some point we should run `SET timezone to 'UTC'` automatically so that we don't need to keep adding it to every test. This looks good to me for now though!

epgts (approved):

PR #547, opened 2022-09-28
Upgrade to pgx 0.5.0

Smittyvb:
Upgrades to pgx `0.5.0-beta.1` (0.5.0 will hopefully be released in 0-7 days). I originally tested on another branch with a lot of code commented out; this branch cherry picks the commits that fix errors but not the ones that comment code out.

To do:
- [x] Fix the three failing tests in `datum_utils.rs` (https://github.com/tcdi/pgx/pull/740)
- [x] Remove workaround for lack of `Copy` on some types
- [x] Make update tester work in CI (#552)
- [x] Change pgx version from 0.5.0-beta.1 to 0.5.0
- [x] Publish new CI docker image with pgx 0.5.0 right before merging (https://github.com/timescale/timescaledb-toolkit/pull/571)

Future work:
- Replace `use pgx::*` with `use pgx::prelude::*`
- Use native pgx types instead of `crate::raw`
- Use `#[pg_aggregate]` instead of our own aggregate builder macro

PR #559, opened 2022-10-04
Basic test for space saving aggregate serialization

Smittyvb:
I noticed while working on #547 that we don't have any tests for `SpaceSavingAggregate`s. This PR adds a simple test to verify that the `raw_freq_agg` function will at least run without failing.

This test currently fails on #547, which is good since that PR causes a bug that leads to `SpaceSavingAggregate` not working at all.


epgts (approved):
Something is certainly better than nothing :)

PR #560, opened 2022-10-05
chown all postgres dirs, not only for postgres 14

epgts:


Smittyvb (approved):

epgts:
None of the checks will pass, so it'll be interesting to see if bors is willing to merge.  If it's not, I know another way out.  It's also the way to allow us to test with a candidate build image, which last time we discussed we didn't know how to do, but now I do.

PR #565, opened 2022-10-06
Add github action to automate the release process.

epgts:
This ties together all the work we've done so far, giving us that push-button release process.

### DIFF
--- a/tools/release
+++ b/tools/release
@@ -154,9 +154,6 @@ branch="$BRANCH_BASENAME"$VERSION
 $nop git checkout -b $branch $COMMIT
 
 # Set default_version in control file.
-# TODO delete if we're not going this direction.
-# $nop sed --in-place -e "/^default_version = / { s/'[^']*'\$/'$VERSION'/ }" $CONTROL
-# assert_dirty || die "failed to set default_version = '$VERSION' in $CONTROL"
 
 # 2. Set toolkit version.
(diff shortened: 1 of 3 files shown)


tools/release
```diff
@@ -0,0 +1,210 @@
...
+    arg=$1
+    shift
+    case "$arg" in
+        -n)
+            dry_run_flag=--dry-run
+            nop=:
+            ;;
+
+        -y)
+            # Ignored - gives us something to pass in the Github Action where default is no-op.
```
epgts, line 130: This is confusing and gross, but I just don't have a better idea!  It gets the job done...

tools/release
```diff
@@ -0,0 +1,210 @@
...
+# 7a. Set toolkit version to released version with '-dev' appended.
+cargo set-version $dry_run_flag -p timescaledb_toolkit $VERSION-dev
+assert_dirty || die "failed to set toolkit version to $VERSION-dev in $TOML"
+start_commit $TOML
+# Update cargo.lock - this form of cargo update doesn't update dependency versions.
+$nop cargo update -p timescaledb_toolkit
+assert_dirty || die "failed to set toolkit version to $VERSION-dev in Cargo.lock"
+commit Cargo.lock
+
+# 7b. Bump github action cache keys.
```
Smittyvb, line 205: This can be removed now that there isn't anything to do in this stage.

PR #568, opened 2022-10-07
Allow `approx_count` accessor to take NULL inputs

### LINKED ISSUE #567: `approx_count` of `count_min_sketch` fails when called via `time_bucket_gapfill` with empty buckets
 - OS: MacOS 12.6
 - PostgreSQL version (output of `SELECT version();`): PostgreSQL 14.5 (Ubuntu 14.5-1.pgdg22.04+1) on x86_64-pc-linux-gnu, compiled by gcc (Ubuntu 11.2.0-19ubuntu1) 11.2.0, 64-bit
 - TimescaleDB Toolkit version (output of `\dx timescaledb_toolkit` in `psql`): 1.11.0
 - Installation method: Docker Desktop for MacOS, version 4.12.0 (85629), running `timescale/timescaledb-ha` image ID `69e412b09e24`

**Describe the bug**
Query for `approx_count` of `count_min_sketch` fails when called via `time_bucket_gapfill` and there are completely empty buckets somewhere in the queried range.

**To Reproduce**
1. Create example table:
```sql
CREATE TABLE IF NOT EXISTS foo (
    PRIMARY KEY (id, time),
	
    id    VARCHAR(10) NOT NULL,
    time  TIMESTAMPTZ NOT NULL DEFAULT now(),
    name  VARCHAR(10) NOT NULL DEFAULT ''
);
SELECT create_hypertable('foo', 'time', chunk_time_interval => INTERVAL '1 day');
SELECT add_retention_policy('foo', INTERVAL '12 months');
```
2. Insert some sample data:
```sql
INSERT INTO foo (id, time, name)
VALUES
('id1', now() - INTERVAL '1 minute', 'bar'),
('id1', now() - INTERVAL '2 minutes', 'bar'),
('id1', now() - INTERVAL '3 minutes', 'baz'),
('id1', now() - INTERVAL '4 minutes', 'bar');
```
3. Run query:
```sql
SELECT time_bucket_gapfill('1 day', time) AS bucket,
       toolkit_experimental.approx_count('bar', toolkit_experimental.count_min_sketch(name, 0.01, 0.01)) AS count
FROM foo
WHERE time > now() - INTERVAL '1 week'
AND time < now()
GROUP BY bucket
ORDER BY bucket DESC
LIMIT 7;
```

**Expected behavior**
Query output should be a set of rows corresponding to the past 7 days, with a non-null value for the most recent bucket and null for the rest.
```
         bucket         | count 
------------------------+-------
 2022-10-07 00:00:00+00 |     3
 2022-10-05 00:00:00+00 |     
 2022-10-04 00:00:00+00 |     
 2022-10-03 00:00:00+00 |     
 2022-10-02 00:00:00+00 |     
 2022-10-01 00:00:00+00 |     
 2022-09-30 00:00:00+00 |     
(7 rows)
```

**Actual behavior**
Query results in an error:
```
ERROR:  called `Option::unwrap()` on a `None` value
CONTEXT:  extension/src/countminsketch.rs:131:28
```

**Screenshots**
N/A.

**Additional context**
Only seems to be an issue when `time_bucket_gapfill` runs over a bucket that is completely empty. As long as all buckets in the range have at least *some* data the query succeeds.

For what it's worth, `time_bucket` does not result in the same issue but obviously does not return a contiguous set of query results for the requested range. With `time_bucket` instead:
```sql
SELECT time_bucket('1 day', time) AS bucket,
       toolkit_experimental.approx_count('bar', toolkit_experimental.count_min_sketch(name, 0.01, 0.01)) AS count
FROM foo
WHERE time > now() - INTERVAL '1 week'
AND time < now()
GROUP BY bucket
ORDER BY bucket DESC
LIMIT 7;
```
The output looks good given the expected behaviour of `time_bucket`:
```
         bucket         | count 
------------------------+-------
 2022-10-07 00:00:00+00 |     3
(1 row)
```

I can work around this for now with something like:
```sql
SELECT time_bucket_gapfill('1 day', time) AS bucket,
       locf(toolkit_experimental.approx_count('bar', toolkit_experimental.count_min_sketch(name, 0.01, 0.01))) AS count,
       toolkit_experimental.count_min_sketch(name, 0.01, 0.01) AS cms
FROM foo
WHERE time > now() - INTERVAL '1 week'
AND time < now()
GROUP BY bucket
ORDER BY bucket DESC
LIMIT 7;
```
The interpolation using `locf` produces a value where the empty bucket would trip the query up, and the additional return of the function allows distinguishing between "real" and interpolated data (which I can then drop) but this is not ideal.

Perhaps there's a nicer workaround, but it feels like the function behaviour is off in the first place so I'm checking if there's a proper fix possible instead.

rtwalker:
Fixes #567

The `approx_count` function previously unwrapped it's aggregate argument without handling the NULL case. With this change, it will instead return NULL when given a NULL for the CountMinSketch.

### DIFF
--- a/extension/src/countminsketch.rs
+++ b/extension/src/countminsketch.rs
@@ -127,9 +127,8 @@ impl toolkit_experimental::count_min_sketch {
 }
 
 #[pg_extern(immutable, parallel_safe, schema = "toolkit_experimental")]
-pub fn approx_count(item: String, aggregate: Option<CountMinSketch>) -> i64 {
-    let sketch = aggregate.unwrap();
-    CountMinSketch::to_internal_countminsketch(&sketch).estimate(item)
+pub fn approx_count(item: String, aggregate: Option<CountMinSketch>) -> Option<i64> {
+    aggregate.map(|sketch| CountMinSketch::to_internal_countminsketch(&sketch).estimate(item))
 }
 
 #[cfg(any(test, feature = "pg_test"))]
@@ -258,6 +257,21 @@ mod tests {
             let output = client
                 .select(
                     "SELECT toolkit_experimental.count_min_sketch(NULL::TEXT, 0.1, 0.1)::TEXT",
+                    None,
+                    None,
+                )
+                .first()
+                .get_one::<String>();
+            assert_eq!(output, None)
+        })
+    }
+
+    #[pg_test]
+    fn test_approx_count_null_input_yields_null_output() {
+        Spi::execute(|client| {
+            let output = client
+                .select(
+                    "SELECT toolkit_experimental.approx_count('1'::text, NULL::toolkit_experimental.countminsketch)",
                     None,
                     None,
                 )


Smittyvb (approved):

epgts (approved):
worth a 1.11.1 ?

rtwalker:
> worth a 1.11.1 ?

up to you :)

PR #569, opened 2022-10-07
Remove incorrect comment in Dockerfile

Smittyvb:
bullseye is Debian 11; buster is Debian 10. We're on the latest Debian release. See https://github.com/timescale/timescaledb-toolkit/pull/552#discussion_r989113244.


epgts (approved):
Thanks!

PR #570, opened 2022-10-07
Use toml_edit to read extension version from Cargo.toml .

epgts:
Reusing control-file parser led us to require single-quotes which aren't the general style and aren't what cargo-edit uses.

### DIFF
--- a/tools/update-tester/src/main.rs
+++ b/tools/update-tester/src/main.rs
@@ -12,7 +12,7 @@ use colored::Colorize;
 
 use xshell::{read_file, Cmd};
 
-use control_file_reader::{get_current_version, get_upgradeable_from};
+use control_file_reader::get_upgradeable_from;
 use postgres_connection_configuration::ConnectionConfig;
 
 // macro for literate path joins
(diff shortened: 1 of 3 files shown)


tools/update-tester/src/main.rs
```diff
@@ -12,7 +12,7 @@ use colored::Colorize;
 
 use xshell::{read_file, Cmd};
 
-use control_file_reader::{get_current_version, get_upgradeable_from};
+use control_file_reader::get_upgradeable_from;
```
Smittyvb, line 15: I updated `control_file_reader::get_current_version` in [my original PR](https://github.com/timescale/timescaledb-toolkit/pull/542/files#diff-afeba27bb9d030df7815c311e6d6176f9ebf81748922fb27282e53a4e3a4afc7R8) as well. `get_current_version` is used in some other places (e.g. the post-install script), so it would probably be best to to update the version-getting logic there.

epgts: I looked at that one.  It reads the control file.  It was accidental that we could get it to extract version numbers out of two different file formats, and even then only by making one of them odd :)

The only remaining use in post-install is actually on the generated control file (with actual version substituted in), not Cargo.toml .

epgts: Although, I will change the parameter name back to control_file .

PR #571, opened 2022-10-07
Update Docker image for pgx 0.5.0

Smittyvb:
pgx 0.5.0 just released! This PR updates the CI to have pgx 0.5.0 and adds pgx 0.5.0 to the `$PATH` instead of pgx 0.4.5.

This PR should be merged right before merging #547 so that the CI can pass on that PR before merging it.

### DIFF
--- a/docker/ci/setup.sh
+++ b/docker/ci/setup.sh
@@ -60,10 +60,7 @@ EOF
 
         # Install cargo pgx
         # Keep synchronized with `cargo install --version N.N.N cargo-pgx` in Readme.md and Cargo.toml
-        cargo install cargo-pgx --version =0.2.4 --root ~/pgx/0.2
-        cargo install cargo-pgx --version =0.4.5 --root ~/pgx/0.4
-
-        PATH=$HOME/pgx/0.4/bin:$PATH
+        cargo install cargo-pgx --version =0.5.0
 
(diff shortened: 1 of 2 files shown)


docker/ci/setup.sh
```diff
@@ -62,6 +62,7 @@ EOF
         # Keep synchronized with `cargo install --version N.N.N cargo-pgx` in Readme.md and Cargo.toml
         cargo install cargo-pgx --version =0.2.4 --root ~/pgx/0.2
         cargo install cargo-pgx --version =0.4.5 --root ~/pgx/0.4
+        cargo install cargo-pgx --version =0.5.0 --root ~/pgx/0.5
```
epgts, line 65: We're no longer using 2 and 4, right?

Let's drop the other two and also drop the `--root`.

Or is something stopping us from doing that?

epgts (changes requested):
Thanks!

docker/ci/setup.sh
Smittyvb: It looks like we don't use 0.2/0.4 anywhere, removed.

epgts (approved):

Smittyvb:
The CI passes on #571 with this PR's Docker image: https://github.com/timescale/timescaledb-toolkit/actions/runs/3266892912/jobs/5371291553

bors r+

PR #572, opened 2022-10-11
Drop now-unused tools/build test-updates command.

epgts:
And stop caching its now nonexistent cache dir.

And leave TODO about dropping the now-unused full-update-test-source.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -164,13 +164,6 @@ jobs:
         key: ${{ runner.os }}-test-updates-target-1.12.0
         restore-keys: ${{ runner.os }}-test-crates-target-1.11.0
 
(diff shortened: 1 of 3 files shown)


Smittyvb (approved):

Smittyvb:
Just to note: this would prevent us from running update tests on macOS (really any platform that can't install debs or rpms).

epgts:
> Just to note: this would prevent us from running update tests on macOS (or any platform that can't install debs or rpms).

Ah, Mac!  Hrm.  Hrrrrm.

Well, deleting this now is certainly premature, so I'll close this pull request.

You Mac folks will need to figure out whether you plan to install three copies of pgx, or if we should prioritize @rtwalker's brew work and add support for that to testbin...

@WireBaron thoughts?

PR #573, opened 2022-10-12
Add .dockerignore and update .gitignore .

epgts:
Without .dockerignore, 'docker build' copies muchos gigabytes into docker-land for no good reason.

.gitignore wasn't ignoring 'old-versions' (from tools/update-tester) and was ignoring .sql files we no longer litter the working directory with (maybe old versions of pgx did?).

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,11 @@
+**/*.iml
+**/*.o
+**/.DS_Store
+.editorconfig
+.idea
+.vscode
+.vsls.json
+.git
+old-versions
+target
+target-analyzer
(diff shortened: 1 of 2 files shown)


epgts:
If this looks good, I'll squash them together and un-draft.  If this looks wrong, please have a look at the TODOs in commit 243ea4f and let me know what you think.

Thanks!

WireBaron (approved):
This looks good.  The sql ignores were indeed needed for an older version of pgx (actually a little sad they don't have this anymore as it was a convenient way to see your generated sql).

epgts:
"   0: `pgx-0.4.5` shouldn't be used with `cargo-pgx-0.5.0`, please use `pgx = "~0.5.0"` in your `Cargo.toml`."

hehe

I'll let

PR #576, opened 2022-10-13
Make Actions cache keys update automatically

Smittyvb:
Based on [how pgx does caching](https://github.com/tcdi/pgx/blob/b92c60974c203e07e05cf1320581084be20a8e8a/.github/workflows/tests.yml#L103), compute GitHub Actions cache keys based on the hash of lockfile and workflow file. Now we don't need to bump the versions manually every release.

Note that `restore-keys` will look for the most recent matching cache with the same prefix if there are no exact matches, so a cache will be available even on commits that update the lockfile/workflow file.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,15 +63,15 @@ jobs:
         path: |
           /usr/local/cargo/registry
           /usr/local/cargo/git
-        key: ${{ runner.os }}-test-pg${{ matrix.pgversion }}-cargo-1.12.0
-        restore-keys: ${{ runner.os }}-test-pg${{ matrix.pgversion }}-cargo-1.11.0
+        key: ${{ runner.os }}-test-pg${{ matrix.pgversion }}-cargo-${{ hashFiles('Cargo.lock', '.github/workflows/ci.yml') }}
+        restore-keys: ${{ runner.os }}-test-pg${{ matrix.pgversion }}-cargo-
 
     - name: Cache cargo target dir
       uses: actions/cache@v2
       with:
         path: target
-        key: ${{ runner.os }}-test-pg${{ matrix.pgversion }}-target-1.12.0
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -63,15 +63,15 @@ jobs:
         path: |
           /usr/local/cargo/registry
           /usr/local/cargo/git
-        key: ${{ runner.os }}-test-pg${{ matrix.pgversion }}-cargo-1.12.0
-        restore-keys: ${{ runner.os }}-test-pg${{ matrix.pgversion }}-cargo-1.11.0
+        key: ${{ runner.os }}-test-pg${{ matrix.pgversion }}-cargo-${{ hashFiles('Cargo.lock', '.github/workflows/ci.yml') }}
```
epgts, line 66: How about hashing extension/Cargo.toml instead?  That would keep the cache freshening at the same frequency we currently intend (release-based), since that's where we change the version number.

Our Cargo.lock changes far less often, so the cache would fall farther behind if we use it.  Same for ci.yml .

I'm going to propose getting rid of these entirely after I finish with the unified build container image.

epgts (changes requested):
Thanks!

.github/workflows/ci.yml
Smittyvb: All the versions in `extension/Cargo.lock` are also in the lockfile. Since we use the Toolkit version as the Cargo version of the extension the lockfile will change whenever we change the version:
https://github.com/timescale/timescaledb-toolkit/blob/d5eeb7b90b4e01bb795d0855423f6854cbeb2f17/Cargo.lock#L2312-L2314

So by using the lockfile the cache key will change when we:
- update dependency versions
- update the Toolkit version
- update the CI workflow

epgts: Right, of course it is!  Thanks! :)

epgts (approved):

PR #577, opened 2022-10-13
Removed ROOT_DIR arg

thatzopoulos:
Removed ROOT_DIR as an arg from all subcommands and set its value to always be "."

### DIFF
--- a/tools/build
+++ b/tools/build
@@ -133,7 +133,6 @@ while [ $# -gt 0 ]; do
                  -h localhost \
                  -p $pg_port \
                  --cache old-versions \
-                 . \
                  "$pg_config" \
                  "$cargo_pgx" \
                  "$cargo_pgx_old"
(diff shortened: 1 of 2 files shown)


rtwalker (approved):

Smittyvb (approved):

epgts (approved):
Can we just drop ROOT_DIR entirely?

Thanks!

thatzopoulos:
> Can we just drop ROOT_DIR entirely?
> 
> Thanks!

Removed root_dir arg from the CLI and updated tools/build to not use it anymore

epgts:
> Would you like me to remove it from the internal functions that currently use it as well or is it enough to set it to "." in main.rs?

I think it's fine either way.

Don't forget to squash away the intermediate commit and update the PR description :)

Thanks!

PR #583, opened 2022-10-17
Update Docker image to support building binary packages.

epgts:
Our current package-building script builds fresh Docker images for all platforms, builds the packages, and then discards the Docker images. So slow and wasteful!

It's also all duplicated with the container-building script here.

Now all this logic lives in just one place.

We do have to keep the packaging in the private repository in order to access our hosted arm64 runner.

Next steps:
- Use the new image
- Move the package-build script here
- Automated release action

### DIFF
--- a/.github/workflows/ci_image_build.yml
+++ b/.github/workflows/ci_image_build.yml
@@ -7,56 +7,31 @@ on:
       - '.github/workflows/ci_image_build.yml'
   workflow_dispatch:
     inputs:
-      tags:
-        description: 'Push image to DockerHub with these tags (change testing to latest to enable)'
+      tag-base:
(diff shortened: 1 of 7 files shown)


epgts:
Plan of attack:
1. #573
2. #582
3. This
4. https://github.com/timescale/release-build-scripts/pull/73
5. #584

Successful action runs:
- building all the containers: https://github.com/timescale/release-build-scripts/actions/runs/3254493687/jobs/5342809153
- building the packages:  https://github.com/timescale/release-build-scripts/actions/runs/3266558375

WireBaron (approved):
Looks good over all.  There does seem to be a fair bit of overlap with other PRs you have out, but I assume you're managing all that.

Issue #586, opened 2022-10-18
use pgx::prelude::* instead of pgx::*


syvb:
pgx 0.5.0 [adds a prelude](https://github.com/tcdi/pgx/pull/699). Instead of importing `pgx::*` everywhere, we can now do `use pgx::prelude::*` instead and not import everything (importing everything pollutes the scope of the importer and makes it less clear where an item came from).

We should update Toolkit to use the prelude everywhere and remove all glob imports of `pgx::*`.

PR #590, opened 2022-10-21
Adding min_n and max_n aggregates

### LINKED ISSUE #511: maxn_n / min_n
A user on [Stackoverflow](https://stackoverflow.com/questions/73414332/there-is-any-equivalent-for-top-and-bottom-in-sql-that-available-in-influxdb) asked for equivalent functions to `TOP` and `BOTTOM` when they are porting queries from InfluxDB to TimescaleDB. `TOP` and `BOTTOM` are different from `MAX` and `MIN`. They allow multiple values to be returned.

One way to design this is to follow our established two-step aggregate pattern. We can have `maxn_agg` / `minn_agg` as the aggregates. They will support `rollup` and have accessors such as: 

- max (for both maxn_agg & minn_agg)
- min (for both maxn_agg & minn_agg)
- ...?

We will likely need an `unnest` function as well to allow values to be accessed.

WireBaron:
This adds new aggregates `min_n` and `max_n` for getting the n largest or smallest values from a column.  It will work with `integer`, `float`, and `timestamptz` values.  These functions will return an aggregate object which can be combined with other such objects via `rollup`.  The data can be extracted from the aggregate via `into_array` or `into_values` methods, which return either an array of the values, or a table containing them.

It further adds `min_n_by` and `max_n_by` functions that take one of the above types plus an associated piece of data (takes this as an `AnyElement`, so can be any type).  This will behave the same as the `min_n`/`max_n` above, but will also return the associated data for the smallest or largest elements.  `into_array` is not implemented for these aggregates, as it's not clear what that array would look like, and `into_values` will require a value of the appropriate type as an input to allow postgres to determine the function output (suggested approach is to just cast a `NULL` to the type of the associated data).

Fixes #511

### DIFF
--- a/extension/src/nmost.rs
+++ b/extension/src/nmost.rs
@@ -1,0 +1,264 @@
+use pgx::*;
+
+use serde::{Deserialize, Serialize};
+
+use crate::{
+    aggregate_utils::in_aggregate_context,
+    datum_utils::{deep_copy_datum, free_datum, DatumStore},
+    palloc::{Inner, Internal, InternalAsValue},
+};
+
+use std::collections::BinaryHeap;
+
+mod max_float;
+mod max_int;
+mod max_time;
+mod min_float;
+mod min_int;
+mod min_time;
+
+mod max_by_float;
+mod max_by_int;
+mod max_by_time;
+mod min_by_float;
+mod min_by_int;
+mod min_by_time;
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct NMostTransState<T: Ord> {
+    capacity: usize,
+    heap: BinaryHeap<T>,
+}
+
+impl<T: Ord> NMostTransState<T> {
+    fn new(capacity: usize, first_val: T) -> NMostTransState<T> {
+        let mut new_heap = NMostTransState {
+            capacity,
+            heap: BinaryHeap::with_capacity(capacity),
+        };
+
+        new_heap.new_entry(first_val);
+
+        new_heap
+    }
+
+    fn new_entry(&mut self, new_val: T) {
+        // If at capacity see if we need to replace something
+        if self.heap.len() == self.capacity {
+            if !self.belongs_in_heap(&new_val) {
+                return;
+            }
+
+            self.heap.pop();
+        }
+
+        self.heap.push(new_val)
+    }
+
+    fn belongs_in_heap(&self, val: &T) -> bool {
+        // Note that this will actually be '>' if T is a Reverse<...> type
+        val < self.heap.peek().unwrap()
+    }
+}
+
+impl<T: Ord + Copy> From<(&[T], usize)> for NMostTransState<T> {
+    fn from(input: (&[T], usize)) -> Self {
+        let (vals, capacity) = input;
+        let mut state = Self::new(capacity, vals[0]);
+        for val in vals[1..].iter() {
+            state.new_entry(*val);
+        }
+        state
+    }
+}
+
+fn nmost_trans_function<T: Ord>(
+    state: Option<Inner<NMostTransState<T>>>,
+    val: T,
+    capacity: usize,
+    fcinfo: pg_sys::FunctionCallInfo,
+) -> Option<Inner<NMostTransState<T>>> {
+    unsafe {
+        in_aggregate_context(fcinfo, || {
+            if state.is_none() {
+                return Internal::new(NMostTransState::<T>::new(capacity, val)).to_inner();
+            }
+
+            let mut state = state.unwrap();
+            state.new_entry(val);
+            Some(state)
+        })
+    }
+}
+
+fn nmost_rollup_trans_function<T: Ord + Copy>(
+    state: Option<Inner<NMostTransState<T>>>,
+    sorted_vals: &[T],
+    capacity: usize,
+    fcinfo: pg_sys::FunctionCallInfo,
+) -> Option<Inner<NMostTransState<T>>> {
+    unsafe {
+        in_aggregate_context(fcinfo, || {
+            if let Some(mut state) = state {
+                for val in sorted_vals {
+                    // The values are sorted, so as soon as we find one that shouldn't be added, we're done
+                    if !state.belongs_in_heap(val) {
+                        return Some(state);
+                    }
+                    state.new_entry(*val);
+                }
+
+                Some(state)
+            } else {
+                Internal::new::<NMostTransState<T>>((sorted_vals, capacity).into()).to_inner()
+            }
+        })
+    }
+}
+
+fn nmost_trans_combine<T: Clone + Ord + Copy>(
+    first: Option<Inner<NMostTransState<T>>>,
+    second: Option<Inner<NMostTransState<T>>>,
+) -> Option<Inner<NMostTransState<T>>> {
+    match (first, second) {
+        (None, None) => None,
+        (None, Some(only)) | (Some(only), None) => unsafe {
+            Internal::new(only.clone()).to_inner()
+        },
+        (Some(a), Some(b)) => {
+            let mut a = a.clone();
+            // This could be made more efficient by iterating in the appropriate order with an early exit, but would requiring ordering the other heap
+            for entry in b.heap.iter() {
+                a.new_entry(*entry);
+            }
+            unsafe { Internal::new(a).to_inner() }
+        }
+    }
+}
+
+// TODO: serialize and deserialize will need to be implemented with Datum handling code
+#[derive(Clone, Debug)]
+pub struct NMostByTransState<T: Ord> {
+    values: NMostTransState<(T, usize)>,
+    data: Vec<Datum>,
+    oid: pg_sys::Oid,
+}
+
+impl<T: Clone + Ord> NMostByTransState<T> {
+    fn new(capacity: usize, first_val: T, first_element: pgx::AnyElement) -> NMostByTransState<T> {
+        // first entry will always have index 0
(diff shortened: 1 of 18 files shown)


extension/src/minn_maxn.rs
```diff
@@ -0,0 +1,1177 @@
...
+};
+
+use ordered_float::NotNan;
+
+use std::cmp::Reverse;
+use std::collections::BinaryHeap;
+use std::marker::PhantomData;
+
+#[derive(Clone, Serialize, Deserialize)]
+pub struct NMostTransState<T: Ord> {
```
Smittyvb, line 21: This struct should have `Debug` derived as well to make debugging easier

extension/src/minn_maxn.rs
```diff
@@ -0,0 +1,1177 @@
...
+    capacity: usize,
+    heap: BinaryHeap<T>,
+    phantom: PhantomData<T>,
```
Smittyvb, line 24: The `PhantomData` is unnecessary, since the type parameter is used by the `BinaryHeap<T>` (which indirectly [uses a `PhantomData` internally](https://github.com/rust-lang/rust/blob/4ca19e09d302a4cbde14f9cb1bc109179dc824cd/library/core/src/ptr/unique.rs#L42)).

extension/src/minn_maxn.rs
```diff
@@ -0,0 +1,1177 @@
+use pgx::{iter::SetOfIterator, *};
```
Smittyvb, line 1: Ideally we would [use the pgx prelude](https://github.com/timescale/timescaledb-toolkit/issues/586) here.

extension/src/minn_maxn.rs
```diff
@@ -0,0 +1,1177 @@
...
+}
+
+fn min_max_trans_function<T>(
+    state: Option<Inner<NMostTransState<T>>>,
+    val: T,
+    capacity: usize,
+    fcinfo: pg_sys::FunctionCallInfo,
+) -> Option<Inner<NMostTransState<T>>>
+where
+    T: Ord,
```
Smittyvb, line 76: nit: this looks clearer without a `where` clause
```suggestion
fn min_max_trans_function<T: Ord>(
    state: Option<Inner<NMostTransState<T>>>,
    val: T,
    capacity: usize,
    fcinfo: pg_sys::FunctionCallInfo,
) -> Option<Inner<NMostTransState<T>>>
```

extension/src/minn_maxn.rs
```diff
@@ -0,0 +1,1177 @@
...
+    }
+}
+
+fn min_max_rollup_trans_function<T>(
+    state: Option<Inner<NMostTransState<T>>>,
+    sorted_vals: &[T],
+    fcinfo: pg_sys::FunctionCallInfo,
+) -> Option<Inner<NMostTransState<T>>>
+where
+    T: Ord + Copy,
```
Smittyvb, line 97: as above
```suggestion
fn min_max_rollup_trans_function<T: Ord + Copy>(
    state: Option<Inner<NMostTransState<T>>>,
    sorted_vals: &[T],
    fcinfo: pg_sys::FunctionCallInfo,
) -> Option<Inner<NMostTransState<T>>>
```

extension/src/minn_maxn.rs
WireBaron: This is a little complicated, as some of the code generated by our macros uses pgx types that aren't a part of the prelude (notably `InOutFuncs`, `StringInfo`, and `rust_regtypein`).  We should still make this change, but I think it should be done in a more comprehensive PR that fixes up the macros (hopefully to remove any dependencies on external `use` statements).

extension/src/minn_maxn.rs
WireBaron: Good catch.

extension/src/minn_maxn.rs
WireBaron: Another good catch.

extension/src/minn_maxn.rs
WireBaron: Fixed

extension/src/minn_maxn.rs
WireBaron: Fixed

extension/src/minn_maxn.rs
```diff
@@ -0,0 +1,1159 @@
...
+}
+
+// Note that min_n queries store a max heap of the lowest values, which is the normal behavior of Rust's BinaryHeap
+// Max_n queries on the other hand, store a min heap of the highest values, which requires a Reverse<'type'> as the type of BinaryHeap.
+type IntMinTransType = NMostTransState<i64>;
+type IntMaxTransType = NMostTransState<Reverse<i64>>;
+type FloatMinTransType = NMostTransState<NotNan<f64>>;
+type FloatMaxTransType = NMostTransState<Reverse<NotNan<f64>>>;
+type TimeMinTransType = NMostTransState<pg_sys::TimestampTz>;
+type TimeMaxTransType = NMostTransState<Reverse<pg_sys::TimestampTz>>;
```
epgts, line 136: What do you think about keeping this file for the common type and functions and creating one child module file per type?  I think it would not only make this file and each individual implementation easier to read, but if kept in the same order, it makes it easy to diff the implementations to spot interesting differences.

extension/src/minn_maxn.rs
```diff
@@ -0,0 +1,1159 @@
...
+    ron_inout_funcs,
+};
+
+use ordered_float::NotNan;
+
+use std::cmp::Reverse;
+use std::collections::BinaryHeap;
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct NMostTransState<T: Ord> {
```
epgts, line 20: The module is named "minn_maxn", functions are named like "min_max_trans_function", but this is named "NMost".  Can we rename this to match?  I think I like "NMost" better so maybe rename the module and functions?

extension/src/minn_maxn.rs
```diff
@@ -0,0 +1,1159 @@
...
+            }
+
+            let mut state = state.unwrap();
+            state.new_entry(val);
+            Some(state)
+        })
+    }
+}
+
+fn min_max_rollup_trans_function<T: Ord + Copy>(
```
epgts, line 85: Are you going to add rollup tests?

extension/src/minn_maxn.rs
WireBaron: Can't say I like it, but I think I dislike it less than any of the alternatives.

extension/src/minn_maxn.rs
WireBaron: I went with `nmost` for the generic functions, the specialized types and functions all follow `min_int` or `min_by_int`.

extension/src/minn_maxn.rs
WireBaron: Added to all unit tests.

extension/src/nmost/max_by_float.rs
```diff
@@ -0,0 +1,220 @@
...
+                    })
+                    .0,
+                    data,
+                })
+            }
+        }
+    }
+}
+
+#[pg_extern(schema = "toolkit_experimental", immutable, parallel_safe)]
```
epgts, line 52: Stylistic question:  why do we sometimes do this rather than just move all these functions into the toolkit_experimental module?  It's always confused me that we use the two different ways to put something into a schema.

extension/src/nmost.rs
```diff
@@ -0,0 +1,264 @@
...
+        {
+            // Full and value belongs in the heap (using len() for this check just keeps us from
+            // succeeding if we tie the max heap element)
+
+            let (_, index_to_replace) = *self
+                .values
+                .heap
+                .peek()
+                .expect("Can't be empty in this case");
+            let _old_datum = std::mem::replace(&mut self.data[index_to_replace], unsafe {
```
epgts, line 177: Why use std::mem::replace when we don't want the previous value?

extension/src/nmost.rs
```diff
@@ -0,0 +1,264 @@
...
+                deep_copy_datum(new_element.datum(), new_element.oid())
+            });
+            // TODO: should we cleaning up 'old_datum' here?  There don't seem to be any convenience functions for this...
```
epgts, line 180: Is this a potential leak?  I don't see where Datum implements Drop...

epgts (approved):
I thought of some more questions but I'm not trying to block merging with them :)

thatzopoulos (approved):

Smittyvb (approved):

extension/src/nmost/max_by_float.rs
WireBaron: You know, I've actually just sort of always followed the existing code and not really thought about it.  I had thought that the schema name was needed for functions to make pgx put them in the right space, but it seems to do the right thing if we just put it in the module as well.  Maybe this wasn't as reliable in a previous version of pgx?

extension/src/nmost.rs
WireBaron: The replace was so that we could free the old datum, but I had just left that as a todo.

extension/src/nmost.rs
WireBaron: Well, everything here is done in the aggregate context, so the memory will be reclaimed with the aggregate finishes, which should be fine in almost every case.  However, I couldn't quite convince myself that we weren't at risk of running out of memory if we ran this on a very large dataset, sorted such that every new element replaces an old value.  So I added a free_datum function that I'm now using here.

PR #589, opened 2022-10-20
Update pgx to 0.5.4.

epgts:
This has various fixes for RHEL derivatives (centos 7 and Rocky 8).

PR #592, opened 2022-10-24
Add build script option to install in release mode

Smittyvb:
Being able to install Toolkit with optimizations enabled is useful for benchmarking purposes. This PR adds an option to build in release mode to the build script.

### DIFF
--- a/tools/build
+++ b/tools/build
@@ -16,7 +16,7 @@ die() {
 }
 
 usage() {
-    die 'build [ -n -pg1[234] ] ( test-crates | test-extension | install | test-doc | test-updates | clippy)'
+    die 'build [ -n -pg1[234] -profile release ] ( test-crates | test-extension | install | test-doc | test-updates | clippy)'
 }
 
 require_pg_version() {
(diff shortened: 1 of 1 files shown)


epgts (commented):
Can you add $mode to all the other cargo command too?  At least those that support it.

Also, what do you think about using `-profile [whatever]` instead?  `--release` is but a strictly less flexible synonym for `--profile release`.

Thanks!

Smittyvb:
> Can you add $mode to all the other cargo command too? At least those that support it.

The only other commands are for running tests – turning on optimizations for tests is usually a bad idea (since in release mode some overflow checks are disabled and debug asserts are also removed). It should probably give an error or at least a warning if one tries to run the tests in release mode.

Smittyvb:
The update test CI failed because the Docker image hasn't been updated with the latest cargo-pgx (which adds `--profile` to `cargo pgx install`). I'll push a new Docker image with the changes from #589.

Smittyvb:
Yep, CI passes now after pushing the updated Docker image.

epgts (approved):

PR #596, opened 2022-10-27
Introduce Candlestick Aggregate

rtwalker:
Use new Candlestick struct to handle volume data.

Introduces two new aggregates
 * `toolkit_experimental.candlestick(ts, price,  volume)` 
 * `toolkit_experimental.candlestick(ts, open, high, low, close,  volume)` 

Happy to consider other names: e.g. `ohlcv`, `candlestick_agg`, `finance_agg`, etc.

Both new aggregates support the following accessors (in addition to being re-aggregated via `rollup`):
 * `open`
 * `high`
 * `low`
 * `close`
 * `open_time`
 * `high_time`
 * `low_time`
 * `close_time`
 * `volume`
 * `vwap` (Volume Weighted Average Price -- hard coded to use closing price for now) 

Original `toolkit_experimental.ohlc(ts, price)` aggregate should still work the same.


extension/src/ohlc.rs
```diff
@@ -14,65 +15,164 @@ use tspoint::TSPoint;
 pub mod toolkit_experimental {
     use super::*;
 
+    flat_serialize_macro::flat_serialize! {
```
rtwalker, line 18: Would appreciate some review on this specifically. Was kinda winging it with getting `flat_serialize!`ing an `enum` to work, and this arrangement was the first one the compiler didn't complain about.

jerryxwu:
I kind of like `ohlcv_agg` and `candlestick_agg` as they follow our previously established naming convention.

rtwalker:
> I kind of like `ohlcv_agg` and `candlestick_agg` as they follow our previously established naming convention.

Do you mean, e.g. 
* `candlestick_agg(ts, price, volume)`
* `candlestick_agg(ts, open, high, low, close, volume)`

?

If the two forms (the 3-arg version and the 5-arg version) aren't supposed to have the same name to me, I'll confess it's not obvious to me which one should be `ohlcv_agg` and which `candlestick_agg`.

I also think `ohlcv_agg(ts, price, volume)` would pair oddly with `ohlc(ts, price)`. Although maybe we can drop `ohlc` altogether soon rather than feel bound to be consistent with it?

extension/src/ohlc.rs
```diff
@@ -14,65 +15,164 @@ use tspoint::TSPoint;
...
+        #[derive(Serialize, Deserialize, Debug, Copy)]
+        enum VolKind {
+            storage_kind: u64,
```
epgts, line 21: I was going to ask what's up with this unused and strangely named value, but I figure I better test it out first :)

I see that it's clearly, surprisingly, quietly, and undocumentedly required, and required first!.

Looks like the name doesn't matter though.  How about `unused_but_required_by_flat_serialize` or something?

extension/src/ohlc.rs
thatzopoulos: I haven't used our flat_serialize crate before but this seems like it would be fine to me.

extension/src/ohlc.rs
```diff
@@ -476,37 +813,93 @@ mod tests {
...
+                let output = select_one!(
+                    client,
+                    format!("SELECT toolkit_experimental.candlestick({x}, {y}, {z})").as_str(),
+                    String
+                );
+                assert_eq!(output, None);
+            }
+
+            // TODO: This fails with invalid utf-8 sequence of 1 bytes from index 274
+            // client.select("SET timezone TO 'UTC'", None, None);
```
epgts, line 831: This seems bad?

extension/src/ohlc.rs
```diff
@@ -438,7 +739,25 @@ mod tests {
...
                             low:(ts:\"{}\",val:1),\
-                            close:(ts:\"{}\",val:1)\
+                            close:(ts:\"{}\",val:1),\
+                            volume:Missing()\
+                            )",
+                    extreme_time, extreme_time, extreme_time, extreme_time
+                );
+                assert_eq!(expected, output.unwrap());
+
+                let stmt = format!("SELECT toolkit_experimental.candlestick(ts, price, volume)::text FROM (VALUES ('{}'::timestamptz, 1.0, 1.0)) v(ts, price, volume)", extreme_time);
```
epgts, line 749: Can you wrap these lines?  I'm sure we could quibble over a good line length, but one of these is 202, and that's a bit much :)

rtwalker: 👍

extension/src/ohlc.rs
rtwalker: I copied `storage_kind` from [here](https://cs.github.com/timescale/timescaledb-toolkit/blob/6791f1ba377342f8e557c89f9ac955590f22457d/extension/src/hyperloglog.rs#L178):
```rust
flat_serialize_macro::flat_serialize! {
    #[derive(Debug, Serialize, Deserialize)]
    enum Storage<'a> {
        storage_kind: u64,
```
so it's possible the convention is `$(enum_name)_kind` but I like your suggestion :)

WireBaron (approved):

Smittyvb (approved):

extension/src/ohlc.rs
epgts: I uncommented and sprinkled some prints.  Looks like it's `output.unwrap()`.

What do you see when you do this in psql?

epgts: pushed a fix.

extension/src/ohlc.rs
```diff
@@ -14,63 +15,124 @@ use tspoint::TSPoint;
...
+        enum VolKind {
+            unused_but_required_by_flat_serialize: u64,
+            Missing: 1 {},
+            Transaction: 2 { vol: f64, vwap: f64 },
+        }
+    }
+
     pg_type! {
-        #[derive(Copy, Debug)]
-        #[repr(C)]
```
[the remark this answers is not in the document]
epgts: @WireBaron ?

extension/src/ohlc.rs
rtwalker: Thanks!!

extension/src/ohlc.rs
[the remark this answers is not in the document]
rtwalker: Hmmm the diff makes it appear more deliberate than it was. I started with fresh struct and slowly replaced the old one in functions until I could get rid of it. I guess I never needed the `repr(C)` on the new implementation and didn’t think to add it?

PR #598, opened 2022-10-31
GitHub Actions: bump actions/checkout and actions/cache to v3

rtwalker:
We're seeing  a lot of deprecation warnings on our CI runs lately e.g. https://github.com/timescale/timescaledb-toolkit/actions/runs/3359573483

which point to this [blog post](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/) about deprecations in some GitHub Actions commands.

Bumping the checkout and cache actions appears to make many of these warnings go away.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,7 +41,7 @@ jobs:
 
     steps:
     - name: Checkout Repository
-      uses: actions/checkout@v2
+      uses: actions/checkout@v3
       with:
         ref: ${{ github.event.pull_request.head.sha }}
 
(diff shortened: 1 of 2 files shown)


rtwalker:
We also use these actions in `.github/workflows/patch_build.yml`.

I made similar changes there initially, but I got back (among other deprecation warnings from our Docker related actions)[^1]:
```
Error response from daemon: Get "https://registry-1.docker.io/v2/": unauthorized: incorrect username or password
```

I removed changes to that file from this PR since it appears we haven't run that action in ~5 months, and I'm not eager to track down the username/password stuff at the moment. 

The PR still at least removes the deprecation warnings that appear on the workflows that we frequently use.

 [^1]: see https://github.com/timescale/timescaledb-toolkit/actions/runs/3363525906

Smittyvb (approved):

Issue #595, opened 2022-10-25
Error on stats_agg with moving aggregate mode can be large, performance is good. 


davidkohn88:
**Describe the bug**
Queries like `SELECT stddev(stats_agg(price) OVER (PARTITION BY symbol ORDER BY time RANGE '15 min' PRECEDING) FROM stocks_real_time WHERE symbol = 'AAPL'` while significantly more performant than the equivalent Postgres operation (`SELECT stddev(price) OVER (PARTITION BY symbol ORDER BY time RANGE '15 min' PRECEDING FROM stocks_real_time WHERE symbol = 'AAPL'`) can produce more error due to double precision arithmetic. 

You can find the "real" value by using: `SELECT stddev(price::numeric) OVER (PARTITION BY symbol ORDER BY time RANGE '15 min' PRECEDING FROM stocks_real_time WHERE symbol = 'AAPL'`
which is reasonably performant, while also producing a smaller amount of error. 

We'd like to reduce the error and have a performant moving aggregate mode, finding a better tradeoff by using higher precision floating point values or perhaps calling the numeric equivalents could be useful.

Smittyvb:
Notes from our discussion yesterday:
- Using a [`TwoFloat`](https://docs.rs/twofloat/latest/twofloat/index.html) (pair of `f64`s) internally when in rolling aggregate mode increases accuracy by around a few orders of magnitude, but is ~30% slower to create (benchmarked using `pgbench`). Here's [a query we were using](https://gist.github.com/Smittyvb/4f317f5fb96cdbf13b4eba2a553fc288) to test the accuracy of `TwoFloat` compared to `numeric` and the current implementation
- Using a Postgres `numeric` gives us the most accurate results possible (I use it as a source of truth for the correct values so I assume it is always exactly correct), but is a bit slower than `TwoFloat`.

When creating a stats aggregate in rolling aggregate mode we always convert it to a normal `f64`-based aggregate in the final function, since the extra precision is much more useful for preserving accuracy in the transition function than in the functions that ultimately get called on the aggregate.

My current work to make the internal aggregate data generic over the data type, and add `stats_agg_tf` which uses `TwoFloat`s internally is on the [sv/stats_agg-twofloat](https://github.com/timescale/timescaledb-toolkit/tree/sv/stats_agg-twofloat) branch.

Smittyvb:
[...]

PR #599, opened 2022-11-02
Use TwoFloats when using `stats_agg` in moving-aggregate

Smittyvb:
Makes `stats_agg` use `TwoFloat`s internally for keeping track of the state when in moving-aggregate mode to prevent floating-point error from accumulating. See #595 for some more details.

### DIFF
--- a/crates/stats-agg/src/stats1d.rs
+++ b/crates/stats-agg/src/stats1d.rs
@@ -1,34 +1,67 @@
-use crate::{StatsError, INV_FLOATING_ERROR_THRESHOLD, M3, M4};
+use crate::{FloatLike, StatsError, INV_FLOATING_ERROR_THRESHOLD, M3, M4};
 use serde::{Deserialize, Serialize};
+use twofloat::TwoFloat;
 
 #[derive(Debug, PartialEq, Copy, Clone, Serialize, Deserialize)]
 #[repr(C)]
-pub struct StatsSummary1D {
+pub struct StatsSummary1D<T: FloatLike> {
     pub n: u64,
-    pub sx: f64,
-    pub sx2: f64,
-    pub sx3: f64,
-    pub sx4: f64,
+    pub sx: T,
+    pub sx2: T,
+    pub sx3: T,
+    pub sx4: T,
 }
-
-impl Default for StatsSummary1D {
+impl<T> Default for StatsSummary1D<T>
+where
+    T: FloatLike,
+{
     fn default() -> Self {
         Self::new()
     }
 }
 
-impl StatsSummary1D {
-    fn n64(&self) -> f64 {
-        self.n as f64
+// can't make this impl generic without conflicting with the stdlib implementation of From<T> for T
+impl From<StatsSummary1D<f64>> for StatsSummary1D<TwoFloat> {
+    fn from(input_summary: StatsSummary1D<f64>) -> Self {
+        StatsSummary1D {
+            n: input_summary.n,
+            sx: input_summary.sx.into(),
+            sx2: input_summary.sx2.into(),
+            sx3: input_summary.sx3.into(),
+            sx4: input_summary.sx4.into(),
+        }
+    }
+}
+pub fn convert_tf_to_f64(tf: TwoFloat) -> f64 {
+    tf.hi() + tf.lo()
+}
+impl From<StatsSummary1D<TwoFloat>> for StatsSummary1D<f64> {
+    fn from(input_summary: StatsSummary1D<TwoFloat>) -> Self {
+        StatsSummary1D {
+            n: input_summary.n,
+            sx: input_summary.sx.into(),
+            sx2: input_summary.sx2.into(),
+            sx3: input_summary.sx3.into(),
+            sx4: input_summary.sx4.into(),
+        }
+    }
+}
+
+impl<T> StatsSummary1D<T>
+where
+    T: FloatLike,
+{
+    fn n64(&self) -> T {
+        T::from_u64(self.n)
     }
 
     pub fn new() -> Self {
         StatsSummary1D {
             n: 0,
-            sx: 0.0,
-            sx2: 0.0,
-            sx3: 0.0,
-            sx4: 0.0,
+            sx: T::zero(),
+            sx2: T::zero(),
+            sx3: T::zero(),
+            sx4: T::zero(),
         }
     }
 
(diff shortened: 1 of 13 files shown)


crates/stats-agg/src/stats2d/stats2d_flat_serialize.rs
```diff
@@ -0,0 +1,425 @@
+use super::*;
+
+// expanded from FlatSerializable derive macro and made to work right with generic arg
+#[allow(unused_imports)]
+unsafe impl<'a> flat_serialize::FlatSerializable<'a> for StatsSummary2D<f64> {
```
Smittyvb, line 5: `#[derive(FlatSerializable)]` doesn't seem to be able to handle types with generics, so I manually expanded it and fixed the output.

crates/stats-agg/src/stats1d.rs
```diff
@@ -316,15 +354,28 @@ mod tests {
...
         assert_relative_eq!(s1.sx, s2.sx);
         assert_relative_eq!(s1.sx2, s2.sx2);
         assert_relative_eq!(s1.sx3, s2.sx3);
         assert_relative_eq!(s1.sx4, s2.sx4);
     }
 
+    #[track_caller]
+    fn assert_close_enough_tf(s1: &StatsSummary1D<TwoFloat>, s2: &StatsSummary1D<TwoFloat>) {
+        assert_eq!(s1.n, s2.n);
+        assert!((s1.sx - s2.sx).abs() < 0.0001);
```
davidkohn88, line 373: Why are we reducing the precision o f this comparison so much?

Smittyvb: I made the assertion more precise.

crates/stats-agg/src/stats1d.rs
```diff
@@ -364,11 +415,82 @@ mod tests {
...
+        assert_relative_eq!(Into::<f64>::into(p.sx), 31.);
+        assert_relative_eq!(Into::<f64>::into(p.sx2), 218.8);
+        assert_relative_eq!(Into::<f64>::into(p.sx3), 1057.68);
+        assert_relative_eq!(Into::<f64>::into(p.sx4), 24016.336);
+
+        let p = p.remove(tf(18.0)).unwrap();
+
+        assert_eq!(p.n, 4);
+        assert_relative_eq!(Into::<f64>::into(p.sx), 13.);
+        assert_relative_eq!(Into::<f64>::into(p.sx2), 44.75, epsilon = 0.0000001);
```
davidkohn88, line 433: Why does this need to change the epsilon, it may be necessary, but I'd love to have a note as to why this needs it if it does.

Smittyvb: I added explanations for all of them.

extension/src/stabilization_info.rs
```diff
@@ -11,6 +11,17 @@
...
     STABLE_FUNCTIONS
+    "1.12.0" => {
+        stats1d_tf_combine(internal,internal),
+        stats1d_tf_inv_trans(internal,double precision),
+        stats1d_tf_final(internal),
+        stats1d_tf_trans_deserialize(bytea,internal),
+        stats1d_tf_trans(internal,double precision),
+        stats2d_tf_final(internal),
+        stats1d_tf_trans_serialize(internal),
+        stats2d_tf_trans(internal,double precision,double precision),
```
WireBaron, line 22: Looks like you're missing the combine, serialize, and deserialize for the stats2d.  I also see we have `stats1d/2d_summary_trans` and `stats1d/2d_summary_inv_trans` which we use for rollup, would it make sense to have these use TwoFloat as well?

crates/stats-agg/src/stats2d/stats2d_flat_serialize.rs
WireBaron: Would it be cleaner to use the TypeHack workaround you use below?

Smittyvb: Rust doesn't support applying `derive` macros on type aliases so that wouldn't work here. The ideal solution would be to fix the derive macro.

extension/src/stabilization_info.rs
Smittyvb: The combine/serialize/deserialize functions aren't used in moving aggregate mode (since moving aggregate mode doesn't support being parallelized) – I think it might be better to remove them for stats1d instead of adding them to stats2d?

TwoFloat summaries are never exposed to PostgreSQL (the final function converts TwoFloat summaries to normal single-float summaries) so they can't be rolled up.

extension/src/stats_agg.rs
```diff
@@ -1924,4 +2211,46 @@ mod tests {
             client.select("DROP TABLE test_table", None, None);
         });
     }
+
+    #[pg_test]
+    fn stats_agg_rolling() {
```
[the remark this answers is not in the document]
Smittyvb: I updated the fuzz tests to also test rolling aggregate mode where possible. It turns out there was an copy-paste error with my implementation of 2D stats summary conversion that the tests found, I fixed that in 12a974bf0a1ddc56c0cabd355e95fe94c589aaa6.

extension/src/stabilization_info.rs
davidkohn88: We can't remove them for stats1d because in normal mode they can still be parallelized? Or maybe I'm missing something?

davidkohn88: (by normal mode I mean when the aggregate is not used  in window function)

Smittyvb: An aggregate can be in partial aggregation mode, or moving aggregation mode, but not both at the same time. The serial/deserial/combine functions are never called in moving aggregate mode (and there is no way to specify such functions that would never be called). The only reason they exist currently is for the perma-experimental `stats_agg_tf` (which always uses twofloats for the aggregate state).

The serial/deserial/combine functions are already only used in the `stats_agg_tf` function for testing – removing them would only affect the performance of that experimental function, which I wouldn't expect anyone to be relying on.

davidkohn88: Aha! Okay. got it, thanks for the explanation! We might want to add a comment about that in the code somewhere so that when we inevitably forget this in 3 weeks, we don't have to reconstruct our logic quite as much :)

extension/src/stats_agg.rs
```diff
@@ -1521,25 +1749,34 @@ mod tests {
...
     }
 
     fn pg1d_aggy(agg: &str) -> String {
-        format!("SELECT {}(test_y) FROM test_table", agg)
+        format!("SELECT {agg}(test_y), (SELECT {agg}(test_y) OVER (ORDER BY test_x ROWS BETWEEN 1 PRECEDING AND 1 FOLLOWING) FROM test_table LIMIT 1 OFFSET 3) FROM test_table", agg = agg)
     }
 
     fn pg2d_agg(agg: &str) -> String {
-        format!("SELECT {}(test_y, test_x) FROM test_table", agg)
+        format!("SELECT {agg}(test_y, test_x), (SELECT {agg}(test_y, test_x) OVER (ORDER BY test_x ROWS BETWEEN 1 PRECEDING AND 1 FOLLOWING) FROM test_table LIMIT 1 OFFSET 3) FROM test_table", agg = agg)
```
davidkohn88, line 1771: I don't think this is going to be enough rows to actually see a difference and possibly will never even call the twofloat version because the condition that stops us from using the moving agg will be triggered here, I think we probably need to test with a bit more data...

extension/src/stats_agg.rs
```diff
@@ -1884,44 +2187,92 @@ mod tests {
...
+            );
+            assert!(vals.next().unwrap()[1].value::<f64>().unwrap().is_nan());
+            assert!(vals.next().unwrap()[1].value::<f64>().is_some());
+            assert!(vals.next().unwrap()[1].value::<f64>().is_some());
+
+            let mut vals = client.select(
+                "SELECT slope(data.stats_agg) FROM (SELECT stats_agg((EXTRACT(minutes FROM ts)), price) OVER (ORDER BY ts RANGE '50 minutes' PRECEDING) FROM prices) data;",
+                None, None
+            );
+            assert!(vals.next().unwrap()[1].value::<f64>().is_none()); // not sure why this is None?
```
davidkohn88, line 2273: There's only one point in that one so slope's undefined I believe.

Smittyvb: Right, documented.

extension/src/stats_agg.rs
Smittyvb: I made the floating point error threshold `f64::INFINITY` in test mode – hopefully that should do the trick here?

WireBaron (approved):
Looks good

PR #605, opened 2022-11-08
Parameterize update tests in CI.

epgts:
This way we run on all PG versions, not just 14.


Smittyvb:
I think we can remove the whole `testupdates` job now if we move the `sh tools/testbin check-control` part into the `testpostgres` job as well.

Smittyvb (approved):

PR #606, opened 2022-11-08
Prepare for next release cycle

rtwalker:

PR #608, opened 2022-11-09
release automation fixes

epgts:
Address some issues we discovered on first use for 1.12.0.

### DIFF
--- a/tools/release
+++ b/tools/release
@@ -249,7 +249,37 @@ commit Changelog.md
 finish_commit "start $DEV_VERSION"
 $nop git show
 
-# TODO Create a pull request instead?
 if $push; then
+    # TODO Create a pull request instead?
     $nop git push origin main
+
+    # File issue for release tasks that are not yet automated.
+    $nop gh issue create -R timescale/timescaledb-toolkit -F- -t "Release $VERSION" <<EOF
+- [ ] Docker HA image
+
+[Sample pull request](https://github.com/timescale/timescaledb-docker-ha/pull/298)
+
+Add new version to `TIMESCALEDB_TOOLKIT_EXTENSIONS` in:
+- `.github/workflows/build_image.yaml`
+- `.github/workflows/publish_image.yaml`
+- `Makefile`
+
+- [ ] hot-forge
+
+[Sample pull request](https://github.com/timescale/hot-forge/pull/67)
+
+Add two new lines to `bundles.yaml` containing the new version tag:
+```
+- repository: https://github.com/timescale/timescaledb-toolkit
+  tag: $VERSION
+```
+
+- [ ] Copy Changelog.md entries for this release into
+  [github release](https://github.com/timescale/timescaledb-toolkit/releases/tag/$VERSION)
+
+- [ ] Update Homebrew
+
+Build binaries on multiple Mac versions/architectures and submit a pull request like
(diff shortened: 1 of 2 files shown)


rtwalker:
I think I missed adding 1.12.0 to the `upgradeable_from` list in #606, but we probably want the release automation to take care of that as well going forward.

https://github.com/timescale/timescaledb-toolkit/blob/7c690d7068465de000d9b610daa64771ad0f1a23/extension/timescaledb_toolkit.control#LL8

tools/release
```diff
@@ -172,35 +185,101 @@ done
...
 $nop git show
 
-$nop git push
+# TODO Create a pull request instead?
+if $push; then
+    $nop git push origin main
+fi
+
+# File issue for release tasks that are not yet automated.
+gh issue create -R timescale/timescaledb-toolkit -F- -t "Release $VERSION" <<EOF
```
Smittyvb, line 258: I think we would only want to do this if `$push` is set? And also it should have a `$nop` at the start?

epgts: Right on both counts!  I'll address that later today.  Thanks!

epgts:
> I think I missed adding 1.12.0 to the `upgradeable_from` list in #606, but we probably want the release automation to take care of that as well going forward.
> 
> https://github.com/timescale/timescaledb-toolkit/blob/7c690d7068465de000d9b610daa64771ad0f1a23/extension/timescaledb_toolkit.control#LL8

Good point!  We need it to be on main AND on release branch in case of cherry picks.  We very nearly released 1.12.1 that couldn't upgrade from 1.12.0!

So I'm thinking:
- add $VERSION to `upgradeable_from` in the final "prep for next release" phase
- assert that the previous version is listed in `upgradeable_from` in the first "create release" phase

Except...  what's the previous version!  When we're cutting 1.13.0, we need to assert that 1.12.N is in upgradeable_from...  but what N!  Do we enumerate the tags?  Pretty gross...

We'll have to think about this one some more.

```
if [ $patch -eq 0 ]; then
	# E.g. releasing 1.13.0 - how do we know all the patch releases of 1.12 to require in upgradeable_from ?
	# Pick the last from `git tag | grep $major.$(( minor - 1)) | version-sort`?
else
	# E.g. releasing 1.13.2 - this one's easy, we can require every entry from 0..N-1 exists:
	prev=$major.$minor.$(( patch - 1 ))
fi
```

epgts:
I was thinking more about this earlier this morning and I realize now we only need to consider the cherry-pick case.  Full releases already have upgradeable_from fixed up.  It's only when cherry-picking that we might fail to cherry-pick the upgradeable_from bump.

I ran this by Smitty and he didn't say I was an idiot so maybe there's something to it.

Updating the example from above:

```
if [ $patch -eq 0 ]; then
	# E.g. releasing 1.13.0 - these are never cherrypicks and we automatically set upgradeable_from on main.
	# Pick the last from `git tag | grep $major.$(( minor - 1)) | version-sort`?
else
	# E.g. releasing 1.13.2 - this one might be a cherry-pick, so we better enforce it!
	prev=$major.$minor.$(( patch - 1 ))
fi
```

Smittyvb (approved):

thatzopoulos (approved):

WireBaron (approved):

PR #615, opened 2022-11-11
Adding new heartbeat_agg

### LINKED ISSUE #578: Signal Heartbeat Claculation
## Situation
We have about 71'000 signals coming from different power plants. Sometimes the connection to one of the power plant is lost. In order to realize that the connection was lost for each power plant we defined so called "heartbeat" signals, where we know we should get values at least every 10 seconds. 

## Dummy example 
timestamp, signal_id, value 
'2020-01-01T00:00:00', 9, 10
'2020-01-01T00:10:00', 9, 10;
'2020-01-01T00:20:00', 9, 10;
'2020-01-01T00:30:00', 9, 10;
'2020-01-01T00:40:00', 9, 10;
'2020-01-01T00:50:00', 9, 10;
'2020-01-01T01:00:00', 9, 10;
'2020-01-01T01:50:00', 9, 20;
'2020-01-01T02:00:00', 9, 20;
'2020-01-01T02:10:00', 9, 20;
'2020-01-01T02:20:00', 9, 20;
'2020-01-01T02:30:00', 9, 20;
'2020-01-01T02:40:00', 9, 20);
...

Now we say, signal 9 is our heartbeat signal and at least every 10 minute we should get a value otherwise the connection is broken. 

Our expected heartbeat query output should be something like that: 

ts_start_valid, ts_end_valid
'2020-01-01T00:00:00', '2020-01-01T01:00:00'
'2020-01-01T01:50:00', '2020-01-01T02:40:00'
....

## Workaround 
We created a workaround query with a materialized view

```
CREATE MATERIALIZED VIEW heartbeats_powerplant1 AS
with t1 as (
Select
ts,
ts - lag(ts,1) over (ORDER BY ts) as diff,
ts - lead(ts,1) over (ORDER BY ts) as diff_negative
From measurements
Where signal_id in (9)
),
t2 as (
select row_number() over (order by ts ) as id,
    ts as start_valid
    from t1 where (diff > '10 minutes' or diff is Null)
),
t3 as (
select row_number() over (order by ts ) as id,
ts as end_valid
    from t1 where (diff_negative < '- 10 minutes' or diff_negative is Null)
),
t4 as (
Select start_valid, end_valid from t2
FULL outer Join t3 on t2.id=t3.id
)
select * from t4;
``` 

## Problem we are facing with that materialized view
Since we have a lot of data the query takes about 7min and if we want to refresh the materialized view every 15min we ran into problem. Now we have seen the (CREATE MATERIALIZED VIEW (Continuous Aggregate)  from the timescale toolkid but the problem is, that we don't have a time bucketing function in our query.  

It would be nice to have a materialzed view if refresh is triggerd only the new data is taken into consideration.

WireBaron:
This adds a new postgres aggregate to our toolkit that can be use to measure the liveness of a system given regular heartbeats.

Fixes #578

### DIFF
--- a/extension/src/heartbeat_agg.rs
+++ b/extension/src/heartbeat_agg.rs
@@ -1,0 +1,886 @@
+use pgx::iter::TableIterator;
+use pgx::*;
+
+use crate::{
+    aggregate_utils::in_aggregate_context,
+    datum_utils::interval_to_ms,
+    flatten,
+    palloc::{Inner, InternalAsValue, ToInternal},
+    pg_type,
+    raw::{Interval, TimestampTz},
+    ron_inout_funcs,
+};
+
+use toolkit_experimental::HeartbeatAggData;
+
+const BUFFER_SIZE: usize = 1000; // How many values to absorb before consolidating
+
+// Given the lack of a good range map class, or efficient predecessor operation on btrees,
+// the trans state will simply collect points and then process them in batches
+pub struct HeartbeatTransState {
+    start: i64,
+    end: i64,
+    last: i64,
+    interval_len: i64,
+    buffer: Vec<i64>,
+    liveness: Vec<(i64, i64)>, // sorted array of non-overlapping (start_time, end_time)
+}
+
+impl HeartbeatTransState {
+    pub fn new(start: i64, end: i64, interval: i64) -> Self {
+        assert!(end - start > interval, "all points passed to heartbeat agg must occur in the 'agg_duration' interval after 'agg_start'");
+        HeartbeatTransState {
+            start,
+            end,
+            last: i64::MIN,
+            interval_len: interval,
+            buffer: vec![],
+            liveness: vec![],
+        }
+    }
+
+    pub fn insert(&mut self, time: i64) {
+        assert!(time >= self.start && time < self.end, "all points passed to heartbeat agg must occur in the 'agg_duration' interval after 'agg_start'");
+        if self.buffer.len() >= BUFFER_SIZE {
+            self.process_batch();
+        }
+        self.buffer.push(time);
+    }
+
+    pub fn process_batch(&mut self) {
+        if self.buffer.is_empty() {
+            return;
+        }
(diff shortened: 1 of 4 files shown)


epgts (approved):

davidkohn88 (commented):
Looking good!

extension/src/raw.rs
```diff
@@ -122,6 +122,21 @@ pub struct Interval(pub pg_sys::Datum);
...
+    fn from(interval: i64) -> Self {
+        let interval = pg_sys::Interval {
+            time: interval,
+            ..Default::default()
+        };
+        unsafe {
+            let ptr =
+                pg_sys::palloc(std::mem::size_of::<pg_sys::Interval>()) as *mut pg_sys::Interval;
+            *ptr = interval;
+            Interval(pgx::Datum::from(ptr))
```
Smittyvb, line 135: `pgx::Datum` won't work in pgx 0.6.0 (see #618)
```suggestion
            Interval(pg_sys::Datum::from(ptr))
```

WireBaron: fixed

extension/src/heartbeat_agg.rs
```diff
@@ -20,16 +20,19 @@ const BUFFER_SIZE: usize = 1000; // How many values to absorb before consolidati
...
     end: i64,
+    last: i64,
     interval_len: i64,
     buffer: Vec<i64>,
     liveness: Vec<(i64, i64)>, // sorted array of non-overlapping (start_time, end_time)
 }
 
 impl HeartbeatTransState {
     pub fn new(start: i64, end: i64, interval: i64) -> Self {
+        assert!(end - start > interval);
```
epgts, line 31: Can this be triggered by user input?  If so, is the error message comprehensible and can we test it?

Same for another assertion below.

WireBaron: I've added a more explicit error message here and to insert that should make it clear to users what went wrong in this case.

PR #617, opened 2022-11-15
Add tests for errors in map_series mapping function

Smittyvb:
pgx 0.6.0 will make some changes to error handling that will affect how the `map_series` pipeline function works. This PR adds some tests to verify that `map_series` behaves correctly when the mapping function throws an error or returns `NULL`.

### DIFF
--- a/extension/src/time_vector/pipeline/map.rs
+++ b/extension/src/time_vector/pipeline/map.rs
@@ -567,6 +567,105 @@ mod tests {
     }
 
     #[pg_test]
+    #[should_panic = "division by zero"]
+    fn test_pipeline_map_series_failure() {
+        Spi::execute(|client| {
+            client.select("SET timezone TO 'UTC'", None, None);
+            // using the search path trick for this test b/c the operator is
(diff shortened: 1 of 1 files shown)


extension/src/time_vector/pipeline/map.rs
```diff
@@ -566,6 +566,105 @@ mod tests {
         });
     }
 
+    #[pg_test]
+    #[should_panic = "division by zero"]
```
epgts, line 570: Interesting.  For some reason, we thought we had to use `#[pg_test(error = "...")]` and we could not get that to work under github actions even though it works locally on mac and linux.  I'll have a look at resurrecting all those disabled tests...

epgts (approved):

extension/src/time_vector/pipeline/map.rs
Smittyvb: Support for more test attributes is [new in 0.5.0](https://github.com/tcdi/pgx/pull/613). I'm surprised it ever worked prior to 0.5.0.

Issue #485, opened 2022-08-01
Expose `TDigest` type so it can be populated on the client and then inserted into a Timescale table


oliora:
**Is your feature request related to a problem? Please describe.**
I'm aggregating metrics on the client into a histogram and I want to store this aggregated data in TimescaleDB and be able to use continuous aggregates on this data to downsample and analyze i with using of TimescaleDB hyperfunctions. I have a control over how client aggregates the data the only limitation is that it has to be aggregated.

**Describe the solution you'd like**
I'd like to aggregate data on the client into some structure that I can later insert into TimescaleDB table as a `TDigest` value. Then I can use all the TimescaleDB functions that work over `TDigest` type (e.g. `rollup`, `approx_percentile` etc). This also unifies the approaches between storing aggregated and non-aggregated data in TimescaleDB.

**Describe alternatives you've considered**
I've considered two alternatives:
1. Collect metrics via Promscale. This is a viable alternative, but Promscale histogram has a limited use with continuous aggregates and has other limitations coming from the Prometheus metrics format.
2. Storing non-aggregated data in TimescaleDB. This works for some of my use cases but not for all. In some cases it is not possible to transfer all the non-aggregated data to TimescaleDB due to it's amount and data channel limits.

yalon:
Hi, we'd love to get t-digest importable too. We have a data pipeline where we pre-compute t-digests (and other metrics) in several places long before they reach timescale due to the sheer size of the data. Timescale holds the aggregated metrics and then does further re-aggregations (downsampling, etc.).
Since we can't ship all the raw data to timescale, we have to have a way to import t-digests into it, otherwise we can't use it.

kvc0:
I'm trying to get a tdigest into timescaledb:
```
DbError { severity: "ERROR", parsed_severity: Some(Error), code: SqlState(E42883), message: "no binary input function available for type tdigest", detail: None, hint: None, position: None, where_: None, schema: None, table: None, column: None, datatype: None, constraint: None, file: Some("lsyscache.c"), line: Some(2913), routine: Some("getTypeBinaryInputInfo") }
```

I use a binary ingest like this:
```
copy metrics (latency) from stdin with binary
```

and I was trying for a hello-world via:
```rust
impl ToSql for SqlTdigest {
    fn to_sql(
        &self,
        ty: &Type,
        out: &mut bytes::BytesMut,
    ) -> Result<postgres_types::IsNull, Box<dyn std::error::Error + Sync + Send>>
[...]

PR #620, opened 2022-11-16
Add experimental library for client-side use, with examples.

epgts:
For issue #485.

### DIFF
--- a/crates/t-digest-lib/Cargo.toml
+++ b/crates/t-digest-lib/Cargo.toml
@@ -1,0 +1,13 @@
+[package]
+name = "tdigest-lib"
+version = "0.0.0"
+edition = "2021"
+
+[lib]
+name = "timescaledb_toolkit_tdigest"
+crate-type = ["cdylib", "staticlib"]
+
+[dependencies]
+libc = "0.2.135"
+
+tdigest = { path="../t-digest" }
--- a/docs/examples/tdigest.c
+++ b/docs/examples/tdigest.c
@@ -1,0 +1,68 @@
+// cc -o tdigest tdigest.c $CARGO_TARGET_DIR/$PROFILE/libtimescaledb_toolkit_tdigest.a -lm -lpthread -ldl
+
+// Sample program which prints the expected output of the test_tdigest_io test.
+
+////////////////////////////////////////////////////////////////////////////////
+// TODO Generate a header from tdigest-lib crate.
+
+#include <sys/types.h>
+
+struct TDigestBuilder;
+struct TDigest;
+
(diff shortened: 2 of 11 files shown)


crates/t-digest-lib/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "tdigest-lib"
+version = "0.0.0"
```
Smittyvb, line 3: You should specify the Rust edition so it doesn't default to `2015`.

```suggestion
version = "0.0.0"
edition = "2021"
```

rtwalker:
This probably merits a CHANGELOG entry right?

docs/examples/tdigest.c
```diff
@@ -0,0 +1,65 @@
...
+main()
+{
+    struct TDigestBuilder *builder = timescaledb_toolkit_tdigest_builder_with_size(100);
+    double value;
+    for (value = 1.0; value <= 100.0; value++) {
+        timescaledb_toolkit_tdigest_push(builder, value);
+    }
+
+    struct TDigest *td = timescaledb_toolkit_tdigest_build(builder);
+    printf("%s\n", timescaledb_toolkit_tdigest_format_for_postgres(td));
```
Smittyvb, line 60: We should free the return value from `timescaledb_toolkit_tdigest_format_for_postgres` in our example here to avoid leaking memory. It's not actually a problem here since the OS will clean it up after `main` anyways, but we don't want users to copy this example and forget to `free` this buffer.

epgts:
> This probably merits a CHANGELOG entry right?

Actually, I'm not sure it does.  I think at this point we're just looking at seeing what one user thinks about it.

@davidkohn88 ?

extension/src/tdigest.rs
```diff
@@ -655,11 +646,10 @@ mod tests {
             assert_eq!(buffer, expected);
 
             let expected = pgx::varlena::rust_byte_slice_to_bytea(&expected);
-            let new_state =
-                tdigest_deserialize_inner(bytea(pg_sys::Datum::from(expected.as_ptr())));
+            let mut new_state =
+                tdigest_deserialize_inner(bytea(pgx::Datum::from(expected.as_ptr())));
```
WireBaron, line 650: I think we actually need to go the other way, and change all the `pgx::Datum` to `pg_sys::Datum` for pgx 0.6.

WireBaron (approved):
Looks good as a proof of concept.

extension/src/tdigest.rs
epgts: Hehe, oops.  Thanks!

PR #635, opened 2022-11-24
Adding asof joins for timevector

### LINKED ISSUE #633: asof_join for timevectors
This is just a way to combine two timevectors where we get the times and values from one of the timevectors, and the LOCF values from the other timevector at the corresponding timestamps from the first vector.

Here's a proposal for how this would work in practice.
```sql
    WITH s as (
        SELECT timevector(time, value) AS v1 FROM
        (VALUES 
            ('2022-10-1 1:00 UTC'::TIMESTAMPTZ, 20.0),
            ('2022-10-1 2:00 UTC'::TIMESTAMPTZ, 30.0),
            ('2022-10-1 3:00 UTC'::TIMESTAMPTZ, 40.0)
        ) as v(time, value)),
    t as 
        SELECT timevector(time, value) AS v2 FROM
        (VALUES 
            ('2022-10-1 0:30 UTC'::TIMESTAMPTZ, 15.0),
            ('2022-10-1 2:00 UTC'::TIMESTAMPTZ, 45.0),
            ('2022-10-1 3:30 UTC'::TIMESTAMPTZ, 60.0)
        ) as v(time, value))
    SELECT asof_join(v1, v2)  /* Alternately, v1 -> asof(v2) */
    FROM s, t;
```
```
         time          | val1 | val2
-----------------------+------+-----
2022-10-01 00:30:00+00 |      |  15
2022-10-01 02:00:00+00 |  30  |  45
2022-10-01 03:30:00+00 |  40  |  60
```

I'm a little uncertain as to which argument should provide the timestamps.  `v1 -> asof(v2)` seems to signify that we should use the timestamps from v2, but when the call is `asof_join(v1, v2)` it becomes much less clear.

WireBaron:
This change adds a new function `join_asof` which takes two timevectors and outputs the values of the first at the times given in the second (as well as the values in the second).

Fixes #633

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -7,8 +7,7 @@ This changelog should be updated as part of a PR if the work is worth noting (mo
 ## Next Release (Date TBD)
 
 #### New experimental features
-- [#615](https://github.com/timescale/timescaledb-toolkit/pull/615): Heartbeat aggregate
-
+- [#615](https://github.com/timescale/timescaledb-toolkit/pull/615): Heatbeat aggregate
   Users can use the new `heartbeat_agg(timestamp, start_time, agg_interval, heartbeat_interval)` to track the liveness of a system in the range (`start_time`, `start_time` + `agg_interval`). Each timestamp seen in that range is assumed to indicate system liveness for the following `heartbeat_interval`.
 
   Once constructed, users can query heartbeat aggregates for `uptime` and `downtime`, as well as query for `live_ranges` or `dead_ranges`. Users can also check for `live_at(timestamp)`.
@@ -22,6 +21,10 @@ This changelog should be updated as part of a PR if the work is worth noting (mo
   This is a lightly tested prototype; try it out at your own risk!
 
   [Examples](docs/examples/)
+
+- [#635](https://github.com/timescale/timescaledb-toolkit/pull/635): AsOf joins for timevectors
+
+  This allows users to join two timevectors with the following semantics `timevectorA -> asof(timevectorB)`. This will return records with the LOCF value from timevectorA at the timestamps from timevectorB. Specifically the returned records contain, for each value in timevectorB, {the LOCF value from timevectorA, the value from timevectorB, the timestamp from timevectorB}.
 
 #### Bug fixes
 
--- a/extension/src/time_vector.rs
+++ b/extension/src/time_vector.rs
@@ -4,7 +4,7 @@ use pgx::{iter::TableIterator, *};
 
 use crate::{
     aggregate_utils::in_aggregate_context,
-    build,
+    build, flatten,
     palloc::{Inner, Internal, InternalAsValue, ToInternal},
     pg_type, ron_inout_funcs,
 };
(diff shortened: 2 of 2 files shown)


extension/src/time_vector.rs
```diff
@@ -368,6 +368,91 @@ CREATE AGGREGATE rollup(\n\
...
+            name!(value1, Option<f64>),
+            name!(value2, f64),
+        ),
+    > {
+        let mut from = from
+            .into_iter()
+            .map(|points| (points.ts.into(), points.val))
+            .peekable();
+        let into = into.into_iter().map(|points| (points.ts, points.val));
+        let (mut from_time, mut from_val) = from.next().unwrap();
```
Smittyvb, line 392: The error from `.unwrap()` isn't particularly user friendly:
```suggestion
        let (mut from_time, mut from_val) = from.next().expect("Must have at least one point in joining column");
```

WireBaron: Good point, however it's also going to have potentially ugly behavior if the into vector is empty, so I added an assert that neither is empty to the start of this function.

extension/src/time_vector.rs
```diff
@@ -368,6 +368,95 @@ CREATE AGGREGATE rollup(\n\
...
+        'a,
+        (
+            name!(value1, Option<f64>),
+            name!(value2, f64),
+            name!(time, crate::raw::TimestampTz),
+        ),
+    > {
+        assert!(
+            from.num_points > 0 && into.num_points > 0,
+            "both timevectors must be populated for an asof join"
```
epgts, line 389: Can we test these two errors?

WireBaron: Added...it's actually surprisingly difficult to create an empty timevector.

Changelog.md
```diff
@@ -7,11 +7,14 @@ This changelog should be updated as part of a PR if the work is worth noting (mo
 ## Next Release (Date TBD)
 
 #### New experimental features
-- [#615](https://github.com/timescale/timescaledb-toolkit/pull/614): Heatbeat aggregate
+- [#615](https://github.com/timescale/timescaledb-toolkit/pull/615): Heatbeat aggregate
```
epgts, line 10: Hehe we caught the same thing!  Also "heatbeat" though, and also there are some rendering problems without blank lines.  I sent a pull request for these.

Changelog.md
```diff
@@ -7,11 +7,14 @@ This changelog should be updated as part of a PR if the work is worth noting (mo
...
   Users can use the new `heartbeat_agg(timestamp, start_time, agg_interval, heartbeat_interval)` to track the liveness of a system in the range (`start_time`, `start_time` + `agg_interval`). Each timestamp seen in that range is assumed to indicate system liveness for the following `heartbeat_interval`.
   Once constructed, users can query heartbeat aggregates for `uptime` and `downtime`, as well as query for `live_ranges` or `dead_ranges`. Users can also check for `live_at(timestamp)`.
   Heartbeat aggregates can also interpolated to better see behavior around the boundaries of the individual aggregates.
 
+- [#635](https://github.com/timescale/timescaledb-toolkit/pull/635): AsOf joins for timevectors
+  This allows users to join two timevectors with the following semantics `timevectorA -> asof(timevectorB)`. This will return records with the LOCF value from timevectorA at the timestamps from timevectorB. Specifically the returned records contain, for each value in timevectorB, {the LOCF value from timevectorA, the value from timevectorB, the timestamp from timevectorB}.
```
epgts, line 16: Without a blank line between lines 15 and 16 here, Github squishes it all into one paragraph.  The blank does not disrupt the list.

epgts (approved):

Changelog.md
WireBaron: Fixed this and merged your fix for the above.  Will merge once I verify everything looks good.

PR #637, opened 2022-11-28
Support rolling up state/timeline aggs

Smittyvb:
Adds `rollup` for `state_agg` and `timeline_agg`. Still to do:
- Determining how out of order aggregates should be handled (currently they don't work)
- Possibly supporting parallel aggregation for the new `rollup` functions, if out-of-order aggregation is possible

(this should be merged after #636)


Smittyvb:
GitHub seems to close open PRs against a branch when it's renamed, so this PR has been moved to #640.

PR #638, opened 2022-11-28
Time Vector Templates

thatzopoulos:


extension/src/time_vector.rs
```diff
@@ -115,6 +119,48 @@ pub fn unnest<'a>(
...
+    let mut times: Vec<String> = Vec::new();
+    let mut values: Vec<String> = Vec::new();
+    if series.has_nulls() {
+        for (i, point) in series.iter().enumerate() {
+            times.push(timestamptz_to_string(point.ts).unwrap());
+            if series.is_null_val(i) {
+                values.push("null".to_string())
+            } else {
+                match point.val.to_string().as_ref() {
+                    "NaN" | "Inf" => panic!(),
```
Smittyvb, line 142: Could use a better error message here:
```suggestion
                    "NaN" | "Inf" => panic!("All values in the series must be finite"),
```

extension/src/time_vector.rs
```diff
@@ -115,6 +119,48 @@ pub fn unnest<'a>(
...
+                    x => values.push(x.to_string()),
+                }
+            }
+        }
+    } else {
+        // optimized path if series does not have any nulls
+        for point in series {
+            times.push(timestamptz_to_string(point.ts).unwrap());
+            values.push(point.val.to_string());
+        }
```
Smittyvb, line 152: This should handle the case where the series has no nulls but some NaNs/infinities.

thatzopoulos: Updated with a check for NaNs/infinities

extension/src/time_vector.rs
```diff
@@ -115,6 +119,51 @@ pub fn unnest<'a>(
...
+    let mut times: Vec<String> = Vec::new();
+    let mut values: Vec<String> = Vec::new();
+    if series.has_nulls() {
+        for (i, point) in series.iter().enumerate() {
+            times.push(timestamptz_to_string(point.ts).unwrap());
+            if series.is_null_val(i) {
+                values.push("null".to_string())
+            } else {
+                match point.val.to_string().as_ref() {
+                    "NaN" | "Inf" => panic!("All values in the series must be finite"),
```
Smittyvb, line 142: This should also handle negative infinity. Also, I think that `Inf` should be lowercase – `f64::INFINITY.to_string()` gives `inf` (ideally there would be a test for this).
```suggestion
                    "NaN" | "inf" | "-inf" => panic!("All values in the series must be finite"),
```

thatzopoulos: Good catch, it does seem like it should be lowercase when shortened:

> When writing these values as constants in an SQL command, you must put quotes around them, for example UPDATE table SET x = '-Infinity'. On input, these strings are recognized in a case-insensitive manner. The infinity values can alternatively be spelled inf and -inf.

Smittyvb (approved):

extension/src/time_vector.rs
```diff
@@ -424,6 +477,99 @@ mod tests {
...
+                .get_one::<String>()
+                .unwrap();
+
+            assert_eq!(
+                test_paired_template,
+        	"[{\"time\": \"2020-01-01 00:00:00+00\", \"val\": 30}, {\"time\": \"2020-01-02 00:00:00+00\", \"val\": 45}, {\"time\": \"2020-01-03 00:00:00+00\", \"val\": null}, {\"time\": \"2020-01-04 00:00:00+00\", \"val\": 55.5}, {\"time\": \"2020-01-05 00:00:00+00\", \"val\": 10} ]"
+            );
+
+            let test_user_supplied_template = client.select(
+                "SELECT toolkit_experimental.format_timevector(timevector(time, value),'{\"times\": {{ TIMES | json_encode(pretty=true) | safe  }}, \"vals\": {{ VALUES | json_encode(pretty=true) | safe }}}') FROM data",
```
WireBaron, line 525: Is the `json_encode(pretty=true) | safe` required to make this look like valid JSON, can we do that by default (probably not `pretty=true` by default though)?

extension/src/time_vector.rs
```diff
@@ -115,6 +119,55 @@ pub fn unnest<'a>(
...
+/// Util function to convert from *const ::std::os::raw::c_char to String
+/// TimestampTz -> *const c_char -> &CStr -> &str -> String
+pub fn timestamptz_to_string(time: pg_sys::TimestampTz) -> Result<String, Utf8Error> {
+    let char_ptr = unsafe { timestamptz_to_str(time) };
+    let c_str = unsafe { CStr::from_ptr(char_ptr) };
+    c_str.to_str().map(|s| s.to_owned())
+}
+
+#[pg_extern(immutable, schema = "toolkit_experimental", parallel_safe)]
+pub fn format_timevector<'a>(series: Timevector_TSTZ_F64<'a>, format_string: String) -> String {
```
WireBaron, line 131: I don't love the name here.  I would prefer something like `to_json`, though I realize that might be a stretch if we're not validating the result is proper JSON.  `to_string`, `to_text`, `output_as`?

extension/src/time_vector.rs
```diff
@@ -115,6 +119,55 @@ pub fn unnest<'a>(
...
+                    panic!("All values in the series must be finite")
+                }
+                x => values.push(x.to_string()),
+            }
+        }
+    }
+    context.insert("TIMES", &times);
+    context.insert("VALUES", &values);
+    match format_string.as_ref() {
+	"plotly" =>     Tera::one_off("{\"times\": {{ TIMES | json_encode(pretty=true) | safe  }}, \"vals\": {{ VALUES | json_encode(pretty=true) | safe }}}", &context, true).expect("Failed to create plotly template"),
```
WireBaron, line 164: Not sure how I feel about overloading the postgres facing function be either a user supplied format or some known magic string.  I'd prefer this to be an undocumented usage, and for us to proved a user facing `to_plotly` which is essentially just `format_output('plotly')`

extension/src/time_vector.rs
```diff
@@ -115,6 +119,55 @@ pub fn unnest<'a>(
...
+	"paired" =>    {
```
WireBaron, line 165: We do need a better way to allow users to drop paired values into their syntax.  Can we provide some sort of `context.insert("TIMEVALS", ...)` where the ... matches this output?

extension/src/time_vector.rs
thatzopoulos: Thoughts on using the following for the functions we have / will have after I address other comments?:
- to_text() for the base function
- to_plotly() for plotly specific output
- possibly to_json() for any output that we want to actually parse to make sure its valid json in the future (we do not currently do this)

extension/src/time_vector.rs
thatzopoulos: Swapping so that to_plotly() is its own separate function

extension/src/time_vector.rs
thatzopoulos: json_encode() is required but `pretty=true` isn't so I can turn that off if we want to.
If you were to write this instead of using json_encode:
```{\"times\": {{ TIMES }}, \"vals\": {{ VALUES }}}``` it would render as:
```
`"{\"times\": [2020-01-01 00:00:00+00, 2020-01-02 00:00:00+00, 2020-01-03 00:00:00+00, 2020-01-04 00:00:00+00, 2020-01-05 00:00:00+00], \"vals\": [30, 45, null, 55.5, 10]}"
```
json_encode without pretty printing:
```
{\"times\": {{ TIMES | json_encode() | safe  }}, \"vals\": {{ VALUES | json_encode() | safe }}}'
``` looks like this:
```
"{\"times\": [\"2020-01-01 00:00:00+00\",\"2020-01-02 00:00:00+00\",\"2020-01-03 00:00:00+00\",\"2020-01-04 00:00:00+00\",\"2020-01-05 00:00:00+00\"], \"vals\": [\"30\",\"45\",\"null\",\"55.5\",\"10\"]}"`
```
There doesn't seem to be a good builtin way to run json_encode() on all user template code by default though. This might be something we can investigate in a future improvement?

File: extension/src/time_vector.rs (hunk 7 above)
Comment (thatzopoulos) [reply extension/src/time_vector.rs:165]: If we provide a context for TIMEVALS it will look like the following if used directly:
```		"[[2020-01-01 00:00:00+00, 30], [2020-01-02 00:00:00+00, 45], [2020-01-03 00:00:00+00, null], [2020-01-04 00:00:00+00, 55.5], [2020-01-05 00:00:00+00, 10]]" 
```
Does that seem fine/useful?
There doesn't seem to be a great builtin way to get it to output something like this without the user having to write some template code to loop through TIMEVALS: 
```
"[{\"time\": \"2020-01-01 00:00:00+00\", \"val\": 30}, {\"time\": \"2020-01-02 00:00:00+00\", \"val\": 45}, {\"time\": \"2020-01-03 00:00:00+00\", \"val\": null}, {\"time\": \"2020-01-04 00:00:00+00\", \"val\": 55.5}, {\"time\": \"2020-01-05 00:00:00+00\", \"val\": 10} ]" 
```

extension/src/time_vector.rs
WireBaron: This seems like a good set of functionality for initial release.

extension/src/time_vector.rs
```diff
@@ -115,6 +119,76 @@ pub fn unnest<'a>(
...
+    let timevals: Vec<(String, String)> = times
+        .clone()
+        .into_iter()
+        .zip(values.clone())
+        .collect::<Vec<_>>();
+
+    context.insert("TIMES", &times);
+    context.insert("VALUES", &values);
+
+    context.insert("TIMEVALS", &timevals); // paired timevals: ex. (time,value).
```
WireBaron, line 182: I'm not sure either of the representations here (line 171 and 172) are that useful, we'd really like to see this output a valid JSON vector of time, value structs.

I'm guessing we can't just specify a format string like `"{\"times\": {{ TIMES | json_encode(pretty=true) | safe  }}, \"vals\": {{ VALUES | json_encode(pretty=true) | safe }}}"` here?  Could we just parse the string ourselves and replace occurrences of `TIMEVAL` with that string?

WireBaron (approved):
This is in a good state for experimental release, but I would like us to have a good solution for users who want a TIMEVALs vector without having to learn Tera formatting.

extension/src/time_vector.rs
thatzopoulos: That makes sense, I am currently trying to get that first idea to work and if it does not I will try out the second approach.

PR #640, opened 2022-11-29
Support rollup and integer states state/timeline aggs

Smittyvb:
- Adds `rollup` for `state_agg` and `timeline_agg`.
- Supports integer states for `state_agg` and `timeline_agg`. An aggregate can have integer states or string states, but not both. For a few functions this would mean having two functions with the same argument types but differing return types, so in that case I made a separate function for integer-valued aggregates (e.g. `state_int_timeline`).

Ideally this should have been two separate PRs, but my changes to `rollup` and the rest of the `state_agg` code have become intertwined.

(this should be merged after #636)

### DIFF
--- a/extension/src/state_aggregate/rollup.rs
+++ b/extension/src/state_aggregate/rollup.rs
@@ -353,6 +353,7 @@ pub fn state_agg_rollup_combine(
     }
 }
 
+#[allow(clippy::redundant_clone)] // clone is needed so we don't mutate shared memory
 pub fn state_agg_rollup_combine_inner(
     state1: Option<Inner<RollupTransState>>,
     state2: Option<Inner<RollupTransState>>,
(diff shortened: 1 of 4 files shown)


extension/src/state_aggregate/rollup.rs
```diff
@@ -0,0 +1,407 @@
...
+) -> Option<Inner<RollupTransState>> {
+    unsafe {
+        in_aggregate_context(fcinfo, || match (state1, state2) {
+            (None, None) => None,
+            (Some(x), None) => Some(x),
+            (None, Some(x)) => Some(x),
+            (Some(mut x), Some(mut y)) => {
+                x.values.append(&mut y.values);
+                Some(x.into())
+            }
```
WireBaron, line 369: The inputs to combine are pointers to shared memory and shouldn't be modified or returned, as another thread is free to reuse the memory as soon as this call completes.

extension/src/state_aggregate/rollup.rs
```diff
@@ -0,0 +1,407 @@
...
+        })
+    }
+}
+
+#[cfg(any(test, feature = "pg_test"))]
+#[pg_schema]
+mod tests {
```
WireBaron, line 376: Would be good to add a test for a successful rollup.  Preferably one which covers the handling of gaps between the aggregates in both cases (matching states on both sides and not).

WireBaron (approved):
I believe the inputs to the trans function all below to the aggregate context so shouldn't need to be cloned.

extension/src/state_aggregate/rollup.rs
Smittyvb: I've added some doc-tests that use `rollup` that now test matching sides and non-matching sides.

epgts (approved):

PR #643, opened 2022-11-30
Document rollup internals

Smittyvb:
Adds documentation for some implementation details of `rollup`.


docs/rollup_internals.md
```diff
@@ -0,0 +1,52 @@
+# Technical notes on `rollup`
+## About
+Most aggregates in Toolkit have an associated `rollup` aggregate that aggregates multiple aggregates covering various ranges of time and combines them into one merged aggregate.
+
+Most `rollup` aggregates require that each aggregate to be rolled up cover a non-overlapping period of time.
```
WireBaron, line 5: I think this is still a minority of our aggregates that have this limitation on aggregation.  It would be nice to make it more explicit that this approach is specifically for aggregates where we require time-ordered input, and that other cases should continue to implement fully parallel rollups.

WireBaron (approved):

rtwalker:
I believe that the scope of this pattern extends beyond `rollup`.  We have aggregates, e.g. `counter_agg`, that employ this technique even when called without `rollup`.

Also it's probably outside the scope of this PR, but I think it would be helpful for us to eventually include a treatment of why this is preferable to ordered set aggregates.

rtwalker (approved):

docs/ordered-aggregates.md
```diff
@@ -0,0 +1,58 @@
...
+
+PostgreSQL doesn't have a way to guarantee that only non-overlapping time periods can be sent to each parallel worker, rows are distributed essentially as they are seen in a round robin. This means that the aggregates cannot be marked as parallel safe. So then, why do they need combine functions at all? Well, there is another time when combine functions are called and that is in the case of partiionwise aggregation. Partitionwise aggregation is used to perform part of the aggregation on a particular partition and then take the state and combine with aggregates from other partitions. Partitions are disjoint in time for us (this assumes some things and we should still have checks to make sure that we are not getting out of order / overlapping data). We believe the test for this is whether they have a combine function, not whether they are marked parallel safe. Therefore, we always mark these aggregates as parallel restricted rather than parallel safe, which hopefully will allow them to be used for partitionwise but not parallel aggregates. Partitionwise aggregation is a potential large optimization area for multinode so we wanted to make sure we could support that case.
```
WireBaron, line 9:
```suggestion
PostgreSQL doesn't have a way to guarantee that only non-overlapping time periods can be sent to each parallel worker, rows are distributed essentially as they are seen in a round robin. This means that the aggregates cannot be marked as parallel safe. So then, why do they need combine functions at all? Well, there is another time when combine functions are called and that is in the case of partitionwise aggregation. Partitionwise aggregation is used to perform part of the aggregation on a particular partition and then take the state and combine with aggregates from other partitions. Partitions are disjoint in time for us (this assumes some things and we should still have checks to make sure that we are not getting out of order / overlapping data). We believe the test for this is whether they have a combine function, not whether they are marked parallel safe. Therefore, we always mark these aggregates as parallel restricted rather than parallel safe, which hopefully will allow them to be used for partitionwise but not parallel aggregates. Partitionwise aggregation is a potential large optimization area for multinode so we wanted to make sure we could support that case. 
```

WireBaron (approved):

PR #650, opened 2022-12-07
Moving some interpolated functions out of experimental

thatzopoulos:
Moved interpolation functions for average, delta, and rate out of the `toolkit_experimental` schema. Tests now check that the output of these functions match the output of the new `toolkit_experimental.interpolated_*` versions from this PR.

### DIFF
--- a/extension/src/counter_agg.rs
+++ b/extension/src/counter_agg.rs
@@ -619,20 +619,29 @@ fn counter_agg_extrapolated_delta<'a>(summary: CounterSummary<'a>, method: &str)
     }
 }
 
-#[pg_extern(
-    name = "interpolated_delta",
-    immutable,
-    parallel_safe,
-    schema = "toolkit_experimental"
-)]
+// Public facing interpolated_delta
+extension_sql!(
+    "\n\
+     CREATE FUNCTION toolkit_experimental.interpolated_delta(summary countersummary,\n\
+          start timestamptz,\n\
+          duration interval,\n\
+          prev countersummary,\n\
+          next countersummary) RETURNS DOUBLE PRECISION\n\
+     AS $$\n\
+          SELECT interpolated_delta(summary,start,duration,prev,next) $$ LANGUAGE SQL IMMUTABLE PARALLEL SAFE;\n\
+",
+    name = "experimental_interpolated_delta", requires=[counter_agg_interpolated_delta]
+);
+
+#[pg_extern(name = "interpolated_delta", immutable, parallel_safe)]
 fn counter_agg_interpolated_delta<'a>(
     summary: CounterSummary<'a>,
     start: crate::raw::TimestampTz,
-    interval: crate::raw::Interval,
+    duration: crate::raw::Interval,
     prev: Option<CounterSummary<'a>>,
     next: Option<CounterSummary<'a>>,
 ) -> f64 {
-    let interval = crate::datum_utils::interval_to_ms(&start, &interval);
+    let interval = crate::datum_utils::interval_to_ms(&start, &duration);
     summary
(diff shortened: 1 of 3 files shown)


thatzopoulos:
Open to suggestions on what to call `intervalparam` since we can't call it `interval` anymore

extension/src/stabilization_info.rs
```diff
@@ -11,6 +11,11 @@
 
 crate::functions_stabilized_at! {
     STABLE_FUNCTIONS
+    "1.13.0" => {
```
thatzopoulos, line 14: Marked this as stable in 1.13.0 since it is not going in the upcoming release

davidkohn88:
> Open to suggestions on what to call `intervalparam` since we can't call it `interval` anymore

Why can't we call it interval anymore? I think you just need to add double quotes around the name in your SQL and it'll work just fine.

davidkohn88:
One thing I'd like to do is a) write a short readme that codifies our decision about when we are doing the keep the experimental version around for one release thing and b) add a comment that these will be removed in the next release and mentions the readme. 


The readme could be something like this: 
```
We decided that when we are stabilizing accessors that have no changes from their current function signatures we are going  to move the function to the public schema and stabilize but also re-create the version in the experimental schema and keep it around for one version. 

We are doing this because some people may have started using experimental accessors in their code and this gives them a version to update their code without it breaking. We are not going to do this when we are stabilizing a full function family but only for accessors that work on already stabilized function families, these type of accessors are safer to use in production settings as their is little risk that something will be stored in, say, a continuous aggregate that would break, and we'd like to encourage people to use them and not cause unnecessary breakage pain when possible. 

We will only keep them around for one version however so they will have a limited time to update their code and not continue using things in the experimental schema indefinitely. 
```
@epgts does that capture what we discussed reasonably well?

thatzopoulos:
>

extension/src/stabilization_info.rs
```diff
@@ -11,6 +11,11 @@
...
+    interpolated_average(timeweightsummary,timestamp with time zone,interval,timeweightsummary,timeweightsummary),
+    interpolated_delta(countersummary,timestamp with time zone,interval,countersummary,countersummary),
+    interpolated_rate(countersummary,timestamp with time zone,interval,countersummary,countersummary),
```
WireBaron, line 17: Indent these to be consistent with the following blocks.

extension/src/stabilization_info.rs
WireBaron: If it's not going into the upcoming release (we can actually discuss this since we pushed the release out to next week), shouldn't this be marked as stable in 1.14.0?

WireBaron (commented):
Not sure if you were looking for a review yet since this is still marked as draft, but added a few comments for you.

extension/src/counter_agg.rs
```diff
@@ -619,20 +619,25 @@ fn counter_agg_extrapolated_delta<'a>(summary: CounterSummary<'a>, method: &str)
...
+",
+    name = "experimental_interpolated_delta", requires=[counter_agg_interpolated_delta]
+);
+
+#[pg_extern(name = "interpolated_delta", immutable, parallel_safe)]
 fn counter_agg_interpolated_delta<'a>(
     summary: CounterSummary<'a>,
     start: crate::raw::TimestampTz,
-    interval: crate::raw::Interval,
+    intervalparam: crate::raw::Interval,
```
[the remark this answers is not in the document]
thatzopoulos: Renamed to duration

rtwalker (approved):

WireBaron (approved):

PR #656, opened 2022-12-13
Fix locale issue on Enterprise Linux systems 

Smittyvb:
(currently untested)

### DIFF
--- a/tools/update-tester/src/testrunner.rs
+++ b/tools/update-tester/src/testrunner.rs
@@ -256,10 +256,26 @@ impl TestClient {
                 db_creation_client
                     .simple_query(&drop)
                     .unwrap_or_else(|e| panic!("could not drop db {} due to {}", test_db_name, e));
-                let create = format!(
-                    r#"CREATE DATABASE "{}" LC_COLLATE 'C.UTF-8' LC_CTYPE 'C.UTF-8'"#,
-                    test_db_name
-                );
+                let locale_flags = {
+                    #[cfg(target_os = "macos")]
+                    {
+                        "LC_COLLATE 'C' LC_CTYPE 'UTF-8'"
+                    }
+                    #[cfg(not(target_os = "macos"))]
+                    {
+                        match std::process::Command::new("locale").arg("-a").output() {
+                            Ok(cmd)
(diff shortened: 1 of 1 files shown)


tools/update-tester/src/testrunner.rs
```diff
@@ -256,10 +256,26 @@ impl TestClient {
...
+                    {
+                        match std::process::Command::new("locale").arg("-a").output() {
+                            Ok(cmd)
+                                if String::from_utf8_lossy(&cmd.stdout)
+                                    .lines()
+                                    .any(|l| l == "C.UTF-8") =>
+                            {
+                                "LC_COLLATE 'C.UTF-8' LC_CTYPE 'C.UTF-8'"
+                            }
+                            _ => "LC_COLLATE 'C' LC_CTYPE 'C'",
```
epgts, line 274: This does fix on rocky 8 (about to test all systems), but now that I see that C works, why don't we use that everywhere?  I thought we required utf-8 for some reason.

But C is the one locale guaranteed to work.

I'm testing everything with this now:
```
                let locale_flags = "LC_COLLATE 'C' LC_CTYPE 'C'";
```

Thanks!

epgts: sigh:

```
thread 'main' panicked at 'could not create db
original_update_tests_1.12.1 due to db error: ERROR: new
collation (C) is incompatible with the collation of the template
database (C.UTF-8)
```

epgts: (which is to say, 'just use C everywhere' is no good:  trying your branch everywhere now)

epgts (approved):
Looks good on centos7, rocky8, debian1[01], ubuntu{18,20,22}.04

Thanks!

tools/update-tester/src/testrunner.rs
Smittyvb: Just to note: the code is like this because it needs to use the same method as [this code from pgx](https://github.com/tcdi/pgx/blob/5b3b3fa3e702ff837f5490f2b33713cdcf6c7a4c/pgx-pg-config/src/lib.rs#L25-L42).

PR #661, opened 2022-12-15
Bit of post release housekeeping

rtwalker:
Addressed clippy warnings that appear on current stable and ran a spell checker to correct some typos. Did this in a lint-per-commit manner so that any controversial lints could be easily reverted. 

I can also happily add any of these commits to our `.git-blame-ignore-revs` if requested.


epgts (approved):
Looks good to me!  I don't care if we blame-ignore these; this is a lot less churn than the Great Formatting of '22.

rtwalker:
Dropped a typo fix in `docker/patch/Dockerfile` because the CI run failed with a permissions error and I don't know how to resolve.

crates/t-digest/src/lib.rs
```diff
@@ -843,9 +843,7 @@ mod tests {
     fn test_merge_sorted_against_skewed_distro() {
         let t = TDigest::new_with_size(100);
         let mut values: Vec<f64> = (1..=600_000).map(f64::from).collect();
-        for _ in 0..400_000 {
-            values.push(1_000_000.0);
-        }
+        values.resize(1_000_000, 1_000_000.0);
```
Smittyvb, line 846: This line (and the other identical one below) should be
```rust
        values.resize(400_000, 1_000_000.0);
```
since the first argument is the new length.

rtwalker: I got the `1_000_000` (for the first argument) from the `400_000` loop range + the `600_000` from the current length of `values` created in the line above.
 
maybe something like this would be a better fix?
```rust
         let mut values: Vec<f64> = (1..=600_000).map(f64::from).collect();
         values.resize(values.len() + 400_000, 1_000_000.0);
```

epgts:
> Dropped a typo fix in `docker/patch/Dockerfile` because the CI run failed with a permissions error and I don't know how to resolve. Should we consider dropping this workflow? We haven't run it successfully in a looooong time.

I couldn't find the pull request where I tried and failed to update it, but that was six months ago or more.

I think it's past time to nuke it.

crates/t-digest/src/lib.rs
Smittyvb: yeah, that looks a lot cleaner

PR #667, opened 2023-01-03
Add option to build TSDB in CI

Smittyvb:
Adds an option to manually run the CI using an arbitrary TimescaleDB commit. You can also optionally specify a TimescaleDB repo to test the CI with a fork of TimescaleDB.


.github/workflows/ci.yml
```diff
@@ -67,6 +76,19 @@ jobs:
     - name: Run pgx tests
       run: su postgres -c 'sh tools/build -pg${{ matrix.pgversion }} test-extension 2>&1'
 
+    - name: Build and install TimescaleDB
+      if: ${{ inputs.tsdb-commit != '' }}
+      run: |
+        apt-get install -y wget cmake gcc build-essential libkrb5-dev
```
epgts, line 82: How about moving this into a standalone shell script?  `tools/install-timescaledb`?
We should already have build-essential and gcc.

I don't see where you're using wget.  If somehow curl is insufficient, we should switch our curl uses to wget so we don't need both.

We should just add cmake into the image.  It's useful and not too big (and we used to ship it anyway).

Does TimescaleDB really require Kerberos?  That's surprising.

.github/workflows/ci.yml
```diff
@@ -67,6 +76,19 @@ jobs:
...
+        git clone '${{ inputs.tsdb-repo }}'
+        cd timescaledb
+        git pull --all
```
epgts, line 85: pull is fetch + merge and we discard the merge work on the next line.  But git fetch immediately after git clone is unnecessary as clone fetched everything already.  (Also --all is for all remotes but there's only one here).

.github/workflows/ci.yml
```diff
@@ -67,6 +76,19 @@ jobs:
...
+        git switch --detach ${{ inputs.tsdb-commit }}
+        mkdir build
+        cd build
+        cmake .. -DUSE_OPENSSL=0 -DLINTER=0 -DCMAKE_PROGRAM_PATH=/usr/lib/postgresql/${{ matrix.pgversion }}/bin
```
epgts, line 89: Does this CMAKE_PROGRAM_PATH cause our cmake scripts to find the right pg_config and thereby cause `make install` to place the extension into the right place under /usr/lib and thereby clobbering the files from the deb?  That merits a comment.

.github/workflows/ci.yml
Smittyvb: I added cmake and libkrb5-dev directly into the image. TimescaleDB really does require Kerberos, since it depends on libpq, which depends on Kerberos. I don't think TimescaleDB actually uses libpq's Kerberos support though.

.github/workflows/ci.yml
Smittyvb: added a comment

.github/workflows/ci.yml
```diff
@@ -22,6 +30,7 @@ jobs:
     container:
       image: ${{ inputs.container-image || 'timescaledev/toolkit-builder:debian-11-amd64' }}
     strategy:
+      fail-fast: false
```
epgts, line 33: Is this temporary for debugging?

docker/ci/setup.sh
```diff
@@ -166,7 +167,9 @@ EOF
                     make \
                     pkg-config \
                     postgresql-common \
-                    sudo
+                    sudo \
+                    cmake \
+                    libkrb5-dev
```
epgts, line 172: Can we keep these alphabetical?

epgts (commented):
Looks good!  Thanks.

.github/workflows/ci.yml
Smittyvb: discussed at standup, decided we want to disable fast failure

epgts (approved):

PR #669, opened 2023-01-03
Stop installing curl on RHEL-derived systems.

epgts:
It's already installed on the images we build on (centos:7,
rockylinux:8, rockylinux:9) and attempting to install 'curl' no longer
works on rockylinux:9 due to conflicting 'curl-minimal' package.

### DIFF
--- a/docker/ci/setup.sh
+++ b/docker/ci/setup.sh
@@ -101,7 +101,6 @@ if $privileged; then
             # - make
             # - pkg-config
             yum -q -y install \
-                curl \
                 gcc \
                 git \
                 make \


Smittyvb:
This line should be removed since that's where it fails currently:

https://github.com/timescale/timescaledb-toolkit/blob/004a14a3436d234e002101562fe193eae347be83/docker/ci/setup.sh#L104

epgts:
> This line should be removed since that's where it fails currently:
> 
> https://github.com/timescale/timescaledb-toolkit/blob/004a14a3436d234e002101562fe193eae347be83/docker/ci/setup.sh#L104

Yeah, already aborted that build, then found the flags don't work on older RHEL releases, then finally did the SMART thing LAST:  found that curl is already preinstalled anyway!

Will un-draft after the image builds finish.

epgts:
all clear:  https://github.com/timescale/release-build-scripts/actions/runs/3831747002

Smittyvb (approved):

PR #683, opened 2023-01-13
Fixing bugs in heartbeat combine code

### LINKED ISSUE #660: Heartbeat aggregate rollup should interpolate aggregates
Currently rollup just combines the intervals from each of its component aggregates, meaning a 5 minute heartbeat received three minutes before the end of an interval is only valid for three minutes.  If the first heartbeat of the next interval arrives 30 seconds in, the aggregate will mark the 30 seconds prior as dead.  We've solved this problem in the interpolate code, but we should also be correcting this automatically in `rollup`.

As part of fixing this, consider whether any of our other aggregates which support both interpolation and rollup also requires a similar fix.

### LINKED ISSUE #679: Heartbeat agg rollup producing invalid aggregates.
A rollup of hourly data into a daily object gave us the following heartbeat agg object: 

(version:1,start_time:722682000000000,end_time:722685600000000,last_seen:722684189000000,interval_len:60000000,num_intervals:3,interval_starts:[722683956000000,722688060000000,722705763000000],interval_ends:[722684249000000,722688237000000,722685600000000])

There are several problems with this aggregate object.

1. This aggregate only covers one hour, where we expect to see a 24 hour aggregate coming from the rollup.
2. The last_seen value is the last value from the first interval.
3. There are only three intervals present, presumably there should be more. (need to verify this)
4. The last interval has an end time prior to its start (this is caused by the aggregate finalizer trimming the end value of the last range to end_time).

We need to make sure that we're updating start_time, end_time, and last_seen when combining ranges.  Some more investigation is needed to understand if and why we're dropping intervals.

WireBaron:
Prior to this the combine code was only combining the intervals, not updating the interval start and end times or the last point.  As part of this change we're now extending any interval that was truncated by the end of an aggregate that now falls in the midst of the combined area.

Fixes #679 
Fixes #660

### DIFF
--- a/extension/src/heartbeat_agg.rs
+++ b/extension/src/heartbeat_agg.rs
@@ -11,6 +11,8 @@ use crate::{
     ron_inout_funcs,
 };
 
+use std::cmp::{max, min};
+
 use toolkit_experimental::HeartbeatAggData;
 
 const BUFFER_SIZE: usize = 1000; // How many values to absorb before consolidating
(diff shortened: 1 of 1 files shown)


extension/src/heartbeat_agg.rs
```diff
@@ -80,6 +82,22 @@ impl HeartbeatTransState {
         }
     }
 
+    // In general we shouldn't need to change these creation time parameters, but if
+    // we're combining with another interval this may be necessary.
+    fn extend_covered_interval(&mut self, new_start: i64, new_end: i64) {
+        assert!(new_start <= self.start && new_end >= self.end);
```
epgts, line 88: Can a user trigger this, and if so, don't we need an error message that references the user interface?  Should we have two assertions so it's immediately clear which error the user hit?

If it should be impossible, a comment like the one on line 93 would help.

epgts (approved):

Issue #691, opened 2023-01-20
Stabilize timeline aggregate as the new state_aggregate


WireBaron:
## Timeline aggregate

**What evidence do we have the feature is being used**

We do see some instances using this in our function telemetry, though this may be test instances.

**Why do we feel this feature is ready to be stable**

This aggregate is just a specialized form of a state aggregate, which is being stabilized now as well (#690). 

**Is there any known further work needed on this feature after stabilization**

No

**Are there any compatibility concerns that may arise during future work on this feature**

No

### Feature History
- Experimental release version: 1.13.0
- Last version modifying on-disk format: 1.13.0
- Target stabilization version: 1.14.0

### Stabilization checklist:
- [x] Ensure tests exist for all public API
- [ ] Ensure API documentation exists and is accurate
- [x] Remove `toolkit_experimental` tags and update test usages
- [x] Add arrow operators for accessors if applicable
- [x] Ensure arrow operators have test coverage
- [x] If present, ensure `combine` and `rollup` are tested
- [x] Add serialization tests for text format
- [x] Add serialization tests for on disk format
- [x] Add upgrade tests
- [x] Add continuous aggregate test
- [ ] Add feature level documentation

WireBaron:
We've decided that timeline aggregate ended up as a far more useful form of the state_aggregate, encasing a superset of that aggregates functionality.  For that reason we'll be renaming `timeline_agg` as the more understandable `state_agg`.  The current `state_agg` will be renamed and left as experimental to determine if it has enough value to stabilize later.

PR #692, opened 2023-01-23
Pre-stabilization changes for state_agg - part 1

### LINKED ISSUE #687: Add duration_in(state, start, end) accessor for Timeline Agg
Want to have one data structure that makes it easy to get the state over a range of times and then query the duration for just the part of the timeline I want to describe. 

duration_in(state, start, end) accessor for `timeline_agg` which allows you to get the duration in a particular state between two times rather than always describing the full period.

Smittyvb:
- Adds optional `start` and `end` parameters to `duration_in` to only get the duration in the specified range. If `end` is omitted all states after `start` are included. `interpolated_duration_in` now uses these ones as well.
- Renames `state_agg` -> `compressed_state_agg` (final name TBD), `timeline_agg` -> `state_agg`
- Removes mostly unused `next` parameters from interpolated state aggregate functions

Closes #687. Related to #691.

### DIFF
--- a/docs/state_agg.md
+++ b/docs/state_agg.md
@@ -46,7 +46,7 @@ INSERT INTO states_test_5 VALUES
 Compute the amount of time spent in a state as INTERVAL.
 
 ```SQL
-SELECT toolkit_experimental.duration_in('ERROR', toolkit_experimental.compressed_state_agg(ts, state)) FROM states_test;
+SELECT toolkit_experimental.duration_in('ERROR', toolkit_experimental.compact_state_agg(ts, state)) FROM states_test;
 ```
 ```output
  interval
@@ -54,7 +54,7 @@ SELECT toolkit_experimental.duration_in('ERROR', toolkit_experimental.compressed
  00:00:03
 ```
 ```SQL
-SELECT toolkit_experimental.duration_in(2, toolkit_experimental.compressed_state_agg(ts, state)) FROM states_test_4;
+SELECT toolkit_experimental.duration_in(2, toolkit_experimental.compact_state_agg(ts, state)) FROM states_test_4;
 ```
 ```output
  interval
@@ -67,7 +67,7 @@ Extract as number of seconds:
 ```SQL
 SELECT
   EXTRACT(epoch FROM
-    toolkit_experimental.duration_in('ERROR', toolkit_experimental.compressed_state_agg(ts, state))
+    toolkit_experimental.duration_in('ERROR', toolkit_experimental.compact_state_agg(ts, state))
   )::INTEGER
 FROM states_test;
 ```
@@ -133,7 +133,7 @@ SELECT toolkit_experimental.duration_in('OK', toolkit_experimental.state_agg(ts,
 
 ```SQL
 SELECT state, duration FROM toolkit_experimental.into_values(
-    (SELECT toolkit_experimental.compressed_state_agg(ts, state) FROM states_test))
+    (SELECT toolkit_experimental.compact_state_agg(ts, state) FROM states_test))
     ORDER BY state, duration;
 ```
 ```output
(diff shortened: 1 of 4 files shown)


extension/src/state_aggregate.rs
```diff
@@ -337,7 +337,7 @@ pub mod toolkit_experimental {
                     state: self.durations.as_slice()[self.first_state as usize]
                         .state
                         .materialize(&states),
-                    time: self.first_time,
+                    time: interval_start,
```
Smittyvb, line 340: This change is needed to prevent the newly added start and range assertions from failing with `interpolated_duration_in`, but I'm not entirely sure that this is actually correct.

extension/src/state_aggregate.rs
```diff
@@ -738,19 +738,28 @@ impl CompressedStateAggTransState {
...
+            assert!(
+                !agg.0.compressed,
+                "unreachable: interval specified for compressed aggregate"
+            );
+            assert!(
+                start >= agg.0.first_time,
+                "Start time ({}) cannot be before first state ({})",
+                start,
+                agg.0.first_time,
+            );
```
WireBaron, line 761: I'm not sure if it's more reasonable to assert here, or to simply exclude the interval before `first_time`.

docs/state_agg.md
```diff
@@ -79,38 +79,56 @@ FROM states_test;
...
+SELECT toolkit_experimental.duration_in('OK', toolkit_experimental.state_agg(ts, state), '2020-01-01 00:01:00+00', '2 days') FROM states_test;
 ```
 ```output
  duration_in
 -------------
  00:00:57
 ```
 ```SQL
-SELECT toolkit_experimental.duration_in('OK', toolkit_experimental.state_agg(ts, state), '2020-01-01 00:01:00+00') FROM states_test;
+SELECT toolkit_experimental.duration_in('OK', toolkit_experimental.state_agg(ts, state), '2020-01-01 00:01:00+00', NULL) FROM states_test;
```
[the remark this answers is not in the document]
Smittyvb: Nope – just forgot to re-add the default argument. I added it now.

extension/src/state_aggregate.rs
Smittyvb: The failure was from the other assert you commented on:
https://github.com/timescale/timescaledb-toolkit/blob/2db5a02b5d6cc4023ca69858f309d9fb037b3455/extension/src/state_aggregate.rs#L756-L760

I removed the assert and undid this change here.

extension/src/state_aggregate.rs
Smittyvb: I removed this assert; I think excluding the prior interval is a better choice here.

extension/src/state_aggregate.rs
```diff
@@ -737,8 +737,32 @@ impl StateAggTransState {
...
+    range: Option<(TimestampTz, TimestampTz)>,
 ) -> crate::raw::Interval {
-    let time: i64 = state.and_then(|state| aggregate?.get(state)).unwrap_or(0);
+    let time: i64 = if let Some((start, end)) = range {
+        let (start, end) = (start.into(), end.into());
+        assert!(end >= start, "End time must be after start time");
+        if let (Some(state), Some(agg)) = (state, aggregate) {
+            let state = state.materialize(agg.states_as_str());
+            let mut total = 0;
+            for tis in agg.combined_durations.iter() {
```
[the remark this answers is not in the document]
Smittyvb: I optimized the loop in d0cd0c32cde85b825e59b7b908c921f465504f01.

WireBaron (approved):

epgts (approved):

Smittyvb:
@WireBaron looks like [this frequency test](https://github.com/timescale/timescaledb-toolkit/blob/cac9a9ef9bf9dcbb749b453e768f69dcef30740b/extension/src/frequency.rs#L2173) is flaky.

bors retry

PR #693, opened 2023-01-23
Backport testbin fix to 1.13.1 release branch

Smittyvb:
This will allow us to make RPM binaries for 1.13.1 by backporting 75216243 to the 1.13.1 release branch.

### DIFF
--- a/tools/testbin
+++ b/tools/testbin
@@ -98,6 +98,14 @@ skip_from_version() {
 
 # Requires:
 # - FROM_VERSION
+# - PG_VERSION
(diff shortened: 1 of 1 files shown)


epgts:
Won't this necessitate 1.13.1 ?  I think we'd rather just build the same tagged source code we built for other platforms for 1.13.0, and I think adding the `|| :` (ignore testbin failure) to the RH release script will allow that.

Smittyvb:
I'll go with that – just started releasing 1.13.1 with that change in relase-build-scripts.

PR #696, opened 2023-01-31
Add rollup functionality to frequency aggregates

### LINKED ISSUE #685: rollup for freq_agg and topn_agg
I have several daily`topn_agg()` values in materialized view and I want to aggregate them to get `topn_agg()` values for a week, for example

Add rollup function for `topn_agg()` and `freq_agg()` the same way as it works for other aggregates

```sql
CREATE MATERIALIZED VIEW hourly AS
    SELECT timebucket('1hour', dt) as hour, TOPN_AGG(100, field) as field_topn FROM raw_table GROUP BY 1;
SELECT topn(10, rollup(field_topn)) FROM hourly WHERE hour BETWEEN 'A' and 'B'
```

**Describe alternatives you've considered**
It it possible to get all the `topn_agg()` results to the app and aggregate in business logic, but it requires downloading of 6MB hourly data to get a monthly value

**Additional context**
There is already `space_saving_aggregate_combine` rust function, which does exactly what I need, but it is not exposed

WireBaron:
This change implements adds in rollup functions for frequency aggregates.

Note that while the rollup of a set of frequency aggregates will not necessarily be identical to computing a single aggregate over the underlying data (may not have the exact same upper and lower bounds on frequency), the rollup maintains the same invariants and will be able to identify most common elements as long as the frequency is different enough.

Fixes #685

### DIFF
--- a/extension/src/frequency.rs
+++ b/extension/src/frequency.rs
@@ -16,7 +16,7 @@ use serde::{
 };
 
 use crate::{
-    aggregate_utils::{get_collation, in_aggregate_context},
+    aggregate_utils::{get_collation_or_default, in_aggregate_context},
     build,
     datum_utils::{
         deep_copy_datum, DatumFromSerializedTextReader, DatumHashBuilder, DatumStore,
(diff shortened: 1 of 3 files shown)


extension/src/frequency.rs
```diff
@@ -233,6 +233,47 @@ impl SpaceSavingTransState {
...
     }
 
+    fn ingest_aggregate_data(
+        &mut self,
+        val_count: u64,
+        values: &DatumStore,
+        counts: &[u64],
+        overcounts: &[u64],
+    ) {
+        self.total_vals = val_count;
```
epgts, line 243: Should this one assert TEXTOID?

extension/src/frequency.rs
```diff
@@ -233,6 +233,47 @@ impl SpaceSavingTransState {
...
+
+        for (idx, datum) in values.iter().enumerate() {
+            self.entries.push(SpaceSavingEntry {
```
epgts, line 246: Should these assert entries is empty?

It seems these are just constructor part two; have you considered folding the calls in?

extension/src/frequency.rs
thatzopoulos: Also wondering this, unless there's a reason for only checking it in ingest_aggregate_ints()?

thatzopoulos (commented):
I only have the one question about assert!, otherwise looks good to me

extension/src/frequency.rs
WireBaron: This same function is called by both text aggregates and raw aggregates (the AnyElement version of this).  So the OID can really be anything here.

extension/src/frequency.rs
WireBaron: Asserting entries is empty is probably a good idea.  As for constructors, the problem is that there are a couple constructors and which one is used is completely independent of which call we use to fill them.  So if I wanted to fold them in I'd have to have a function for every combination of calls.

epgts (approved):
I think that pg15 failure might be a flake; I've had that test running for a few minutes now with no failures.

WireBaron:
> I think that pg15 failure might be a flake; I've had that test running for a few minutes now with no failures.

That was my thought as well, that test shouldn't be affected by this change at all.  Still, I'll take a closer look and see if we can make the test less flakey.

PR #698, opened 2023-02-08
Adding some missing changelog entries for the upcoming release

WireBaron:


Changelog.md
```diff
@@ -9,8 +9,11 @@ This changelog should be updated as part of a PR if the work is worth noting (mo
 #### New experimental features
 
 #### Bug fixes
+- [#660](https://github.com/timescale/timescaledb-toolkit/issues/660): Heartbeat aggregate rollup should interpolate aggregates
+- [#679](https://github.com/timescale/timescaledb-toolkit/issues/679): Heartbeat agg rollup producing invalid aggregates.
```
Smittyvb, line 13: We should be consistent about whether we end changelog lines lines with a period or not.

thatzopoulos (approved):

Changelog.md
WireBaron: Good point.  Right now these are just copy/paste from the issue, but we should probably try to have some rough style guidelines.

PR #699, opened 2023-02-08
state_agg API tweaks

Smittyvb:
- `interpolated_duration_in`/`duration_in`/`interpolated_state_periods`/`state_periods` have the first two arguments swapped: now the aggregate is first and the state is second
- `into_values`/`into_int_values` now returns a table with intervals instead of microseconds

### DIFF
--- a/extension/src/state_aggregate.rs
+++ b/extension/src/state_aggregate.rs
@@ -1346,8 +1346,8 @@ fn state_periods_inner<'a>(
 
 #[pg_extern(immutable, parallel_safe, schema = "toolkit_experimental")]
 pub fn state_periods<'a>(
-    state: String,
     agg: StateAgg<'a>,
+    state: String,
 ) -> TableIterator<
     'a,
     (
@@ -1366,8 +1366,8 @@ pub fn state_periods<'a>(
     name = "state_periods"
 )]
 pub fn state_int_periods<'a>(
-    state: i64,
     agg: StateAgg<'a>,
+    state: i64,
 ) -> TableIterator<
     'a,
     (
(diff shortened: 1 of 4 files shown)


thatzopoulos (approved):

extension/src/state_aggregate.rs
```diff
@@ -671,6 +671,30 @@ fn state_agg_int_trans(
     }
 }
 
+/// Makes an interval from an `i64` representing the number of microseconds.
+fn make_interval(time: i64) -> crate::raw::Interval {
```
WireBaron, line 675: We actually have a `impl From<i64> for Interval` already in `raw.rs` that does most of this already.  It seems cleaner to just add a `justify_interval` there as well and use something like `justify_interval(time.into())` instead of this function.

WireBaron: That works.  I was a little reluctant to suggest this as I'm guessing the justification isn't sticky...if we add two justified intervals I don't think the result is guaranteed to be justified.  However, I'm pretty sure we never actually do math on these after we convert them to intervals, so that shouldn't be a problem.

WireBaron (approved):

Smittyvb:
I forced the CLA bot to recheck this PR by logging into https://cla-assistant.io/ and clicking "Recheck PRs".

bors retry

Smittyvb:
Looks like the flaky `test_rollups` test failed again.

bors retry

epgts:
Smittyvb wrote:
> Looks like the flaky |test_rollups| test failed again.

Hmmm, did we know that one to be flaky?  I only remember 
`stats_agg::tests::pg_stats_agg_fuzz`...

(not that it needs to block merging this, but we probably ought 
to file an issue with a list of known flaky tests...)

Smittyvb:
> > Looks like the flaky |test_rollups| test failed again.
>
> Hmmm, did we know that one to be flaky?
> I only remember `stats_agg::tests::pg_stats_agg_fuzz`...

It's [this one](https://github.com/timescale/timescaledb-toolkit/blob/cf27d8d51f1515428ae04e7c6a1d18a0f774eb1b/extension/src/frequency.rs#L2173) – we added a few days ago and it's been quite flakey. We should probably make it less flakey or ignore it before the release.

PR #715, opened 2023-02-21
Fix errors in state_agg rollup

Smittyvb:
This PR should fix out-of-bounds indexing in `rollup`:
- before, `last_state` in the merged aggregate was computed incorrectly; it is now determined correctly
- merging aggregates now sorts them first

It also improves the various error messages that could arise when using rollup.

### DIFF
--- a/extension/src/state_aggregate/rollup.rs
+++ b/extension/src/state_aggregate/rollup.rs
@@ -509,29 +509,65 @@ mod tests {
             compact: true,
             integer_states: true,
         };
-        let expected = OwnedCompactStateAgg {
+        let s3 = OwnedCompactStateAgg {
             durations: vec![
                 DurationInState {
-                    duration: 500 * 2 + 12345,
+                    duration: 500,
                     state: StateEntry::from_integer(555_2),
                 },
                 DurationInState {
-                    duration: 400 * 2,
+                    duration: 400,
                     state: StateEntry::from_integer(555_1),
                 },
             ],
             combined_durations: vec![],
-            first_time: 100,
-            last_time: 1900 + 12345,
+            first_time: 1900 + 12345,
+            last_time: 1900 + 12345 + 900,
             first_state: 1,
             last_state: 0,
             states: vec![],
             compact: true,
             integer_states: true,
         };
-        let merged = s1.clone().merge(s2.clone());
+        let expected = OwnedCompactStateAgg {
+            durations: vec![
+                DurationInState {
+                    duration: 500 * 3 + 12345,
+                    state: StateEntry::from_integer(555_2),
+                },
+                DurationInState {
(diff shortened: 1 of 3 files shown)


extension/src/state_aggregate/rollup.rs
```diff
@@ -80,14 +80,18 @@ impl OwnedCompactStateAgg {
...
+            Ordering::Equal => panic!(
+                "can't merge overlapping aggregates (same start time: {})",
+                self.first_time
+            ),
         };
+
         assert!(
             earlier.last_time <= later.first_time,
-            "can't merge overlapping aggregates"
+            "can't merge overlapping aggregates (earlier={earlier:?}, later={later:?})"
```
WireBaron, line 94: Might be useful to include both start and end times for the overlapping aggregates

extension/src/state_aggregate/rollup.rs
```diff
@@ -142,22 +146,36 @@ impl OwnedCompactStateAgg {
 
         let gap = later.first_time - earlier.last_time;
         assert!(gap >= 0);
-        merged_durations[earlier.last_state as usize].duration += gap;
+        merged_durations
+            .get_mut(earlier.last_state as usize)
+            .expect("earlier.last_state doesn't point to a state")
```
WireBaron, line 163: Are these error messages meant to be for us or for end users?  As written, these will be very helpful in letting us see exactly went wrong, but any cloud user who gets an error like this is likely going to have a hard time making any sense of it.

extension/src/state_aggregate/rollup.rs
```diff
@@ -187,8 +199,8 @@ impl OwnedCompactStateAgg {
 
             first_time: earlier.first_time,
             last_time: later.last_time,
-            first_state: earlier.first_state,
-            last_state: added_entries + later.last_state,
+            first_state: earlier.first_state, // indexes into earlier durations are same for merged_durations
+            last_state: merged_last_state,
```
WireBaron, line 203: It would be good to add a test case that fails with the old code, but passes with the error fixed.

extension/src/state_aggregate/rollup.rs
Smittyvb: That error message should never be displayed to an end user — it would only happen if the aggregate is already in an inconsistent state. If a user hits this error it's because of a bug somewhere.

extension/src/state_aggregate/rollup.rs
Smittyvb: I expanded a test I added in this PR so that:
- removing the aggregate sorting results in `merges_compact_aggs_correctly` failing
- removing the fix to `last_state` results in the state_agg doctest failing

WireBaron (approved):

PR #723, opened 2023-03-01
Stabilize state aggregate

### LINKED ISSUE #691: Stabilize timeline aggregate as the new state_aggregate

Smittyvb:
Closes #691.


extension/src/state_aggregate/accessors.rs
```diff
@@ -0,0 +1,495 @@
...
+        AccessorIntoIntValues, AccessorIntoValues, AccessorStateIntTimeline, AccessorStateTimeline,
+    },
+    datum_utils::interval_to_ms,
+    raw::{Interval, TimestampTz},
+    state_aggregate::*,
+};
+
+#[pg_operator(immutable, parallel_safe)]
+#[opname(->)]
+pub fn arrow_state_agg_into_values<'a>(
```
WireBaron, line 12: I did find it more useful to group these functions with the non-arrow versions.  I'm not strictly opposed to having them in a separate file like this, but I don think it's important that we're consistent with how we organize our code.

extension/src/state_aggregate/accessors.rs
```diff
@@ -0,0 +1,495 @@
...
+    (
+        pgx::name!(state, i64),
+        pgx::name!(start_time, TimestampTz),
+        pgx::name!(end_time, TimestampTz),
+    ),
+> {
+    state_int_timeline(agg)
+}
+
+macro_rules! interpolated_state_timeline_accessor {
```
WireBaron, line 70: I don't feel that these macros ended up being very effective here.  You define several macros that you only end up using twice each.  This ends up being much harder to read through and follow, but not that much more efficient at removing replicated code.  The rule of thumb I've always heard is that it's not worth refactoring code until you're using it three times.

extension/src/state_aggregate.rs
```diff
@@ -2272,4 +2273,95 @@ SELECT toolkit_experimental.duration_in('one', toolkit_experimental.compact_stat
...
+                        time: 10005000
+                    }
+                ],
+                integer_states: true,
+            }
+        );
+    }
+
+    #[pg_test]
+    fn binary_serialization() {
```
WireBaron, line 2384: I've made this mistake in a few places as well, but we should be testing the binary state of the finalized aggregate, not the trans state.  The trans state's binary state will never persist into a version update, so we shouldn't need to test it.

WireBaron (approved):

PR #716, opened 2023-02-24
Adding arrow operators for stable interpolating functions

WireBaron:
Also removing experimental copies of stable interpolation functions.

PR #724, opened 2023-03-01
Stabilizing integral and interpolated integral

### LINKED ISSUE #719: Stabilize Integral and interpolated_integral
## Integral / Interpolated Integral

**What evidence do we have the feature is being used**

**Why do we feel this feature is ready to be stable**

**Is there any known further work needed on this feature after stabilization**

No

**Are there any compatibility concerns that may arise during future work on this feature**

No

### Feature History
- Experimental release version: 1.11.0
- Last version modifying on-disk format: 1.12.0
- Target stabilization version: 1.15.0

### Stabilization checklist:
- [x] Ensure tests exist for all public API
- [ ] Ensure API documentation exists and is accurate
- [x] Remove `toolkit_experimental` tags and update test usages
- [x] Add arrow operators for accessors if applicable
- [x] Ensure arrow operators have test coverage
- [x] Add serialization tests for on disk format
- [x] Add serialization tests for text format
- [x] Add upgrade tests

thatzopoulos:
Fixes #719


Smittyvb (approved):

docs/test_caggs.md
```diff
@@ -114,3 +115,14 @@ ORDER BY week;
...
+
+```SQL
+SELECT average(rollup(twa)) as time_weighted_average 
+FROM weekly_aggs;
+```
+
+```output
+ time_weighted_average
+-----------------------
+     67.0032679739
```
WireBaron: Did you validate this result?

tests/update/time-weighted-average.md
```diff
@@ -1,53 +1,56 @@
...
 ```
 
 ```output
- time_weighted_average
+ interpolated_integral
 -----------------------
-     6.636111111111111
+                   360
+                   690
+                   310
```
WireBaron, line 43: Did you validate these numbers?  The first value at least seems wrong to me...when I do it by hand I get 364.  Can you check to see that we're interpolating the end time for the day properly and not just taking the first value for the next day?

WireBaron (changes requested):
It looks like we also need an arrow accessor for the `interpolated_integral`.  You can look at how I did this in https://github.com/timescale/timescaledb-toolkit/pull/716 if you have any questions.

docs/test_caggs.md
thatzopoulos: I did but I also ended up deleting this test from caggs since it did not actually test the newly stabilized features.

tests/update/time-weighted-average.md
thatzopoulos: Updated the test to use interpolated_integral instead which does have a value of 364.

extension/src/time_weighted_average.rs
```diff
@@ -548,6 +538,44 @@ pub fn time_weighted_average_interpolated_integral<'a>(
...
+    } else {
+        None
+    };
+    let next = if accessor.flags & 2 == 2 {
+        Some(accessor.next.clone().into())
+    } else {
+        None
+    };
+
+    // Convert from num of milliseconds to DurationUnit and then to string
```
thatzopoulos, line 558: This section of code is a workaround for the DurationUnit enum not working with the flat_serialize macro.

extension/src/time_weighted_average/accessors.rs
```diff
@@ -54,3 +55,61 @@ fn time_weight_interpolated_average_accessor<'a>(
...
+
+ron_inout_funcs!(TimeWeightInterpolatedIntegralAccessor);
+
+#[pg_extern(immutable, parallel_safe, name = "interpolated_integral")]
+fn time_weight_interpolated_integral_accessor<'a>(
+    start: crate::raw::TimestampTz,
+    interval: crate::raw::Interval,
+    prev: Option<TimeWeightSummary<'a>>,
+    next: Option<TimeWeightSummary<'a>>,
+    unit: String,
```
WireBaron, line 80: This should have the same default to second behavior that the non-arrow version has.

extension/src/time_weighted_average/accessors.rs
```diff
@@ -54,3 +55,61 @@ fn time_weight_interpolated_average_accessor<'a>(
...
     }
 }
+
+pg_type! {
+    #[derive(Debug)]
+    struct TimeWeightInterpolatedIntegralAccessor {
+        start : i64,
+        interval : i64,
+        prev : TimeWeightSummaryData,
+    pad : [u8;3],
```
WireBaron, line 65: Uff, bit late to do anything about this, but seems like we should have made our pg_type objects aligned so they could be composed without needing padding.

PR #736, opened 2023-03-13
Require all CI jobs to pass, not one

Smittyvb:
It turns out that bors interprets "Test Postgres (%)" as requiring at least one job to pass matching that, instead of all of them.


thatzopoulos (approved):
LGTM

PR #739, opened 2023-03-14
Use alter database to set timezone in Update Tester

thatzopoulos:


tests/update/time-vector.md
```diff
@@ -3,11 +3,11 @@
 ```sql,creation
 CREATE TABLE time_vector_data(time TIMESTAMPTZ, value DOUBLE PRECISION);
 INSERT INTO time_vector_data VALUES
-    ('2020-1-1', 30.0),
-    ('2020-1-2', 45.0),
-    ('2020-1-3', NULL),
-    ('2020-1-4', 55.5),
-    ('2020-1-5', 10.0);
+    ('2020-1-1 UTC', 30.0),
```
WireBaron, line 6: Were you seeing this break with the new change?  I would have thought the timezone on the insert would just default to the database setting.

tools/update-tester/src/testrunner.rs
```diff
@@ -360,7 +363,10 @@ impl TestClient {
                 test_client.validate_stable_objects_exist();
 
                 test_client
-                    .simple_query("SET TIME ZONE 'UTC';")
+                    .simple_query(&format!(
+                        r#"ALTER DATABASE "{}" SET timezone TO 'UTC';"#,
+                        &test_db_name
+                    ))
```
WireBaron, line 369: This shouldn't be necessary here.  This persistently sets the default for the database, so just doing it when the database is created should be sufficient.

WireBaron (changes requested):
In the update test `time-weighted-average.md` you have a several rows of output with timestamps like `2020-01-01 00:00:00-06`.  If you're able to run this locally without that test failing then this change isn't working.  We need to be able to specify the timezone for that output, as that's what's different between rocky and debian/centos.

tests/update/time-vector.md
thatzopoulos: Tests pass locally for me still with this change for both this file and the time weighted average tests.

WireBaron (approved):

PR #740, opened 2023-03-17
Fix right boundary when interpolating 'locf' TWA

### LINKED ISSUE #732: Last datapoint is ignored in the calculation of average(time_weight('locf', TIME, value))
 - OS: [timescale/timescaledb-ha:pg14.6-ts2.9.0-latest]
 - PostgreSQL version : 14.6
 - TimescaleDB Toolkit version : [1.12.1]
 - Installation method: ["docker"]

**Describe the bug**]
reported in https://stackoverflow.com/questions/75680213/time-weighted-average-in-timescaledb-using-last-observation-carried-forward

incorrect average is computed from `time_weight`; seems like the last datapoint is not counted in

**To Reproduce**
```
DROP TABLE IF EXISTS t;


CREATE TABLE t(TIME TIMESTAMP WITH TIME ZONE NOT NULL,
                                             value float, k integer);


INSERT INTO t
VALUES ('2020-01-01 00:00:00', 1, 0),
       ('2020-01-01 00:00:01', 1, 0),
       ('2020-01-01 23:00:01', 1000, 1),
       ('2020-01-01 23:59:59', 1000, 2);


SELECT 0 AS k,
       time_bucket('1 days', TIME) AS timebucket,
       average(time_weight('locf', TIME, value)), (time_weight('locf', TIME, value))
FROM t
WHERE (TIME BETWEEN TIMESTAMP '2020-01-01 00:00:00+00:00' AND TIMESTAMP '2020-01-02 00:00:00+00:00')
  AND k <= 0
GROUP BY timebucket
UNION
ALL
SELECT 1,
       time_bucket('1 days', TIME) AS timebucket,
       average(time_weight('locf', TIME, value)), (time_weight('locf', TIME, value))
FROM t
WHERE (TIME BETWEEN TIMESTAMP '2020-01-01 00:00:00+00:00' AND TIMESTAMP '2020-01-02 00:00:00+00:00')
  AND k <= 1
GROUP BY timebucket
UNION
ALL
SELECT 2,
       time_bucket('1 days', TIME) AS timebucket,
       average(time_weight('locf', TIME, value)), (time_weight('locf', TIME, value))
FROM t
WHERE (TIME BETWEEN TIMESTAMP '2020-01-01 00:00:00+00:00' AND TIMESTAMP '2020-01-02 00:00:00+00:00')
  AND k <= 2
GROUP BY timebucket;
```
**Expected behavior**
I guess for `k=1` and `k=2` the same resultset would be expected; but instead the `average` is equal to the `k=0` case

**Actual behavior**

```
 k |       timebucket       |      average      |                                                               time_weight                                                                
---+------------------------+-------------------+------------------------------------------------------------------------------------------------------------------------------------------
 0 | 2020-01-01 00:00:00+00 |                 1 | (version:1,first:(ts:"2020-01-01 00:00:00+00",val:1),last:(ts:"2020-01-01 00:00:01+00",val:1),weighted_sum:1000000,method:LOCF)
 1 | 2020-01-01 00:00:00+00 |                 1 | (version:1,first:(ts:"2020-01-01 00:00:00+00",val:1),last:(ts:"2020-01-01 23:00:01+00",val:1000),weighted_sum:82801000000,method:LOCF)
 2 | 2020-01-01 00:00:00+00 | 42.60235650875589 | (version:1,first:(ts:"2020-01-01 00:00:00+00",val:1),last:(ts:"2020-01-01 23:59:59+00",val:1000),weighted_sum:3680801000000,method:LOCF)
(3 rows)
```

WireBaron:
This change will make it so a time weighted average with the locf method will include the range between the last data point and the interpolation boundary, even when there is no next aggregate to interpolate.

Note that the behavior here might not be entirely intuitive, so we need to make sure we update our docs to describe our exact behavior.

Fixes #732

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -10,6 +10,7 @@ This changelog should be updated as part of a PR if the work is worth noting (mo
 
 #### Bug fixes
 - [#733](https://github.com/timescale/timescaledb-toolkit/pull/733): Fix a bug when rolling up overlapping heartbeat_aggs
+- [#740](https://github.com/timescale/timescaledb-toolkit/pull/740): When interpolating an 'locf' time weighted average, extend last point to interpolation boundary
 
 #### Other notable changes
 
--- a/extension/src/time_weighted_average.rs
+++ b/extension/src/time_weighted_average.rs
@@ -80,11 +80,19 @@ impl<'input> TimeWeightSummary<'input> {
             }
             _ => self.first,
         };
-        let new_end = match next {
-            Some(next) => {
+        let new_end = match (self.method, next) {
+            (_, Some(next)) => {
                 let new_end = self
                     .method
                     .interpolate(self.last, Some(next.first), end)
+                    .expect("unable to interpolate end of interval");
+                new_sum += self.method.weighted_sum(self.last, new_end);
+                new_end
+            }
+            (TimeWeightMethod::LOCF, None) => {
+                let new_end = self
+                    .method
+                    .interpolate(self.last, None, end)
                     .expect("unable to interpolate end of interval");
                 new_sum += self.method.weighted_sum(self.last, new_end);
                 new_end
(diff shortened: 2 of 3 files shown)


thatzopoulos (approved):
LGTM

PR #742, opened 2023-03-28
Handle NULL values in hyperloglog rollup

### LINKED ISSUE #649: Add NULL support to `rollup(hyperloglog)`
Add NULL support to `rollup(hyperloglog)` to improve user experience.

`rollup(hyperloglog)` should simply ignore NULL hyperloglog objects instead of returning an error. For Toolkit 1.12.0, the returned error is:

> ERROR:  other_ is null
> CONTEXT:  extension/src/hyperloglog.rs:277:5

Consider this query:
```SQL
SELECT distinct_count(rollup(logs))
FROM (
	(SELECT hyperloglog(16, v::text) logs FROM generate_series(1, 5) v)
	UNION ALL
	(SELECT hyperloglog(16, v::text) FROM generate_series(6, 10) v WHERE v <=5)
) hll;
```
It should return the same value as this query as the second `hyperloglog` object will be NULL:
```SQL
SELECT distinct_count(rollup(logs))
FROM (
	(SELECT hyperloglog(16, v::text) logs FROM generate_series(1, 5) v)
) hll;
```

**Describe alternatives you've considered**
An alternative is to do nothing and ask the user to always include `FILTER (WHERE hll IS NOT NULL)` clause if `hll` can be null. For the example above, the query would look like this,
```SQL
SELECT distinct_count(rollup(logs) FILTER (WHERE logs IS NOT NULL))
FROM (
	(SELECT hyperloglog(16, v::text) logs FROM generate_series(1, 5) v)
	UNION ALL
	(SELECT hyperloglog(16, v::text) FROM generate_series(6, 10) v WHERE v <=5)
) hll;
```
The alternative works, it's just more boilerplate SQL code to write.

WireBaron:
Fixes #649

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -11,6 +11,7 @@ This changelog should be updated as part of a PR if the work is worth noting (mo
 #### Bug fixes
 - [#733](https://github.com/timescale/timescaledb-toolkit/pull/733): Fix a bug when rolling up overlapping heartbeat_aggs
 - [#740](https://github.com/timescale/timescaledb-toolkit/pull/740): When interpolating an 'locf' time weighted average, extend last point to interpolation boundary
+- [#742](https://github.com/timescale/timescaledb-toolkit/pull/742): Ignore incoming NULL values in hyperloglog rollup
 
 #### Other notable changes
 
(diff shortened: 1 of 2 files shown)


Smittyvb (approved):

WireBaron:
Looks like github is having issues, let's try this again...

bors r+

Issue #747, opened 2023-03-30
functions_stabilized_at() macro panics at types that contain '[]'


thatzopoulos:
The following soon to be stabilized functions cause the `functions_stabilized_at()` macro to fail because they contain the characters '[]':
- "approx_percentile_array(double precision[],uddsketch)"
- "approx_percentiles(double precision[])"

```
error: no rules expected the token `[`
   --> extension/src/stabilization_info.rs:19:49
    |
19  |         approx_percentile_array(double precision[],uddsketch),
    |                                                 ^ no rules expected this token in macro call
    |
   ::: extension/src/stabilization_tests.rs:156:1
    |
156 | macro_rules! functions_stabilized_at {
    | ------------------------------------ when calling this macro
    |
note: while trying to match `)`
   --> extension/src/stabilization_tests.rs:161:64
    |
161 |                 $($fn_name: ident ( $( $($fn_type: ident)+ ),* ) ),* $(,)?
    |                                                                ^

error[E0425]: cannot find function, tuple struct or tuple variant `STABLE_FUNCTIONS` in module `crate::stabilization_info`
   --> extension/src/stabilization_tests.rs:143:36
    |
143 |         crate::stabilization_info::STABLE_FUNCTIONS()
    |                                    ^^^^^^^^^^^^^^^^ not found in `crate::stabilization_info`
```
For now this will be handled with a quick fix that skips these particular functions when running stabilization tests.

PR #748, opened 2023-03-31
Stabilizing approx percentile array

thatzopoulos:
Workaround for #747, we will need a better workaround for this in the future when we make changes to stability checks

### DIFF
--- a/extension/src/stabilization_info.rs
+++ b/extension/src/stabilization_info.rs
@@ -29,6 +29,9 @@ crate::functions_stabilized_at! {
         num_live_ranges(heartbeatagg),
         trim_to(heartbeatagg,timestamp with time zone,interval),
         trim_to(timestamp with time zone,interval),
+        accessorpercentilearray_in(cstring),
+        accessorpercentilearray_out(accessorpercentilearray),
+        arrow_uddsketch_approx_percentile_array(uddsketch,accessorpercentilearray),
     }
     "1.15.0" => {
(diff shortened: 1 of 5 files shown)


extension/src/stabilization_info.rs
```diff
@@ -14,6 +14,9 @@ crate::functions_stabilized_at! {
     "1.16.0" => {
         approx_count_distinct(anyelement),
         approx_count_distinct_trans(internal,anyelement),
+        accessorpercentilearray_in(cstring),
+        accessorpercentilearray_out(accessorpercentilearray),
+        arrow_uddsketch_approx_percentile_array(uddsketch,accessorpercentilearray),
```
WireBaron, line 19: This last function shouldn't go here, the entry you added for `STABLE_OPERATORS` should cover it.  Do you get an error without this entry?

WireBaron (approved):
Address the one comment and this should be good to go.

extension/src/stabilization_info.rs
thatzopoulos: It passes without that entry, I will delete it.

thatzopoulos: If I delete that line, the test fails with:
```
unexpectedly released features: [
    "function arrow_uddsketch_approx_percentile_array(uddsketch,accessorpercentilearray)",
]
```

Smittyvb:
I just trigged a CLA re-check, this should be able to be merged now.

bors retry

PR #751, opened 2023-04-03
Stabilize nmost functions

Smittyvb:
Note that I didn't add arrow accessors for `into_values` for `min_n_by`/`max_n_by` since those functions take dummy `NULL`s.


Smittyvb:
Merging this now so we can get #752 in without more merge conflicts.

bors r+

thatzopoulos (approved):

PR #770, opened 2023-09-06
feat(heartbeat): add average and stddev functions to heartbeat agg

ollz272:
This would be useful for us to have for our own usage (We want to use the stats on average uptime/downtime to do some more analysis).

Let me know if anymore is needed.

### DIFF
--- a/extension/src/heartbeat_agg.rs
+++ b/extension/src/heartbeat_agg.rs
@@ -395,6 +395,78 @@ pub fn uptime(agg: HeartbeatAgg<'static>) -> Interval {
     agg.sum_live_intervals().into()
 }
 
+
+#[pg_extern]
+pub fn average_downtime(agg: HeartbeatAgg<'static>) -> Option<f64> {
+    (agg.end_time - agg.start_time - agg.sum_live_intervals()).into() / agg.num_intervals
+}
+
+#[pg_extern]
+pub fn average_uptime(agg: HeartbeatAgg<'static>) -> Option<f64> {
+    (agg.sum_live_intervals()).into() / agg.num_intervals
+}
+
+#[pg_extern]
+pub fn stddev_downtime(agg: HeartbeatAgg<'static>) -> Option<f64>  {
+// Dead ranges are the opposite of the intervals stored in the aggregate
+    let mut starts = agg.interval_ends.clone().into_vec();
+    let mut ends = agg.interval_starts.clone().into_vec();
+
+    // Fix the first point depending on whether the aggregate starts in a live or dead range
+    if ends[0] == agg.start_time {
+        ends.remove(0);
+    } else {
+        starts.insert(0, agg.start_time);
+    }
+
+    // Fix the last point depending on whether the aggregate starts in a live or dead range
+    if *starts.last().unwrap() == agg.end_time {
+        starts.pop();
+    } else {
+        ends.push(agg.end_time);
+    }
+
+    let mut data = vec![];
+    for i in 0..agg.num_intervals as usize {
+        vec![].push(ends[i] - starts[i]);
+    }
+    match (mean(data), data.len()) {
+        (Some(data_mean), count) if count > 0 => {
+            let variance = data.iter().map(|value| {
+                let diff = data_mean - (*value as f32);
+
+                diff * diff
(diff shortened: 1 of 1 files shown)


syvb:
Please do `cargo fmt` so that we have consistent formatting - thanks!

extension/src/heartbeat_agg.rs
```diff
@@ -395,6 +395,78 @@ pub fn uptime(agg: HeartbeatAgg<'static>) -> Interval {
...
+    }
+}
+
+#[pg_extern]
+pub fn stddev_uptime(agg: HeartbeatAgg<'static>) -> Option<f64> {
+    let starts = agg.interval_starts.as_slice();
+    let ends = agg.interval_ends.as_slice();
+    let mut data = vec![];
+    for i in 0..agg.num_intervals as usize {
+        vec![].push(ends[i] - starts[i]);
```
syvb, line 456: I think you meant this?
```suggestion
        data.push(ends[i] - starts[i]);
```

extension/src/heartbeat_agg.rs
```diff
@@ -395,6 +395,78 @@ pub fn uptime(agg: HeartbeatAgg<'static>) -> Interval {
...
+    }
+    match (mean(data), data.len()) {
```
syvb, line 458: This doesn't compile for me (or CI):
```
error[E0425]: cannot find function `mean` in this scope
   --> extension/src/heartbeat_agg.rs:432:12
    |
432 |     match (mean(data), data.len()) {
    |            ^^^^ not found in this scope
    |
help: use the `.` operator to call the method `Statistics::mean` on `Vec<_>`
    |
432 -     match (mean(data), data.len()) {
432 +     match (data.mean(), data.len()) {
    |
```

extension/src/heartbeat_agg.rs
```diff
@@ -395,6 +395,78 @@ pub fn uptime(agg: HeartbeatAgg<'static>) -> Interval {
     agg.sum_live_intervals().into()
 }
 
+
+#[pg_extern]
+pub fn average_downtime(agg: HeartbeatAgg<'static>) -> Option<f64> {
+    (agg.end_time - agg.start_time - agg.sum_live_intervals()).into() / agg.num_intervals
```
syvb, line 400: This doesn't compile for me (or the CI), same with the one below:
```
error[E0283]: type annotations needed
   --> extension/src/heartbeat_agg.rs:400:64
    |
400 |     (agg.end_time - agg.start_time - agg.sum_live_intervals()).into() / agg.num_intervals
    |                                                                ^^^^
    |
    = note: cannot satisfy `_: From<i64>`
    = note: required for `i64` to implement `Into<_>`
help: try using a fully qualified path to specify the expected types
    |
400 |     <i64 as Into<T>>::into((agg.end_time - agg.start_time - agg.sum_live_intervals())) / agg.num_intervals
    |     +++++++++++++++++++++++                                                          ~
```

extension/src/heartbeat_agg.rs
```diff
@@ -395,6 +395,84 @@ pub fn uptime(agg: HeartbeatAgg<'static>) -> Interval {
...
+    // Fix the last point depending on whether the aggregate starts in a live or dead range
+    if *starts.last().unwrap() == agg.end_time {
+        starts.pop();
+    } else {
+        ends.push(agg.end_time);
+    }
+
+    let mut data = vec![];
+    for i in 0..agg.num_intervals as usize {
+        vec![].push(ends[i] - starts[i]);
```
syvb, line 430: I think you meant this?
```suggestion
        data.push(ends[i] - starts[i]);
```

jonatas:
Hi @ollz272, very good, I'm not a Rust dev here, but I'm an avid user of toolkit too 🤗 

1. can we add a few more tests just calling the functions and asserting the values?
2. or at least add it on [docs/test_caggs](https://github.com/timescale/timescaledb-toolkit/blob/4d3dbde580f683cbb76c7ce1e430050620ef5441/docs/test_caggs.md?plain=1#L120) that could cover and exemplify.

We still don't have the docs/heartbeat_agg which would be the right place to document it. Feel free to introduce the file if you think it makes sense.

PR #776, opened 2023-10-03
Ensure PostgreSQL 16 compatibility

feikesteenbergen:
DateTimeParseError: PostgreSQL source code documents that we can pass null pointers here:

 > extra can be NULL if not needed for the particular dterr value.
 > [...]
 > If escontext points to an ErrorSaveContext node, that is filled instead
 > of throwing an error

DecodeDateTime however does not allow us to pass a null pointer, reading the source code, it is expected that extra is not null. Therefore, in order for us to get this thing to compile, we create an empty DateTimeErrorExtra, which gets passed along.

### DIFF
--- a/tools/dependencies.sh
+++ b/tools/dependencies.sh
@@ -7,9 +7,9 @@
 # All our automation scripts read this, so at least we're not duplicating this
 # information across all those.
 
-PG_VERSIONS='12 13 14 15'
-# TODO: remove this once TimescaleDB supports PostgreSQL 15: issue #648
-TSDB_PG_VERSIONS='12 13 14'
+PG_VERSIONS='12 13 14 15 16'
+# TODO: extend this with 16 this once TimescaleDB supports PostgreSQL 16: issue #5752
+TSDB_PG_VERSIONS='12 13 14 15'
 
 CARGO_EDIT=0.11.2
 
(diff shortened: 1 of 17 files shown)


tools/dependencies.sh
```diff
@@ -7,9 +7,9 @@
 # All our automation scripts read this, so at least we're not duplicating this
 # information across all those.
 
-PG_VERSIONS='12 13 14 15'
-# TODO: remove this once TimescaleDB supports PostgreSQL 15: issue #648
-TSDB_PG_VERSIONS='12 13 14'
+PG_VERSIONS='12 13 14 15 16'
+# TODO: remove this once TimescaleDB supports PostgreSQL 16: issue #5752
```
feikesteenbergen, line 11:
```suggestion
# TODO: extend this with 16 this once TimescaleDB supports PostgreSQL 16: issue #5752
```

feikesteenbergen:
The CI image was [built here](https://github.com/timescale/release-build-scripts/actions/runs/6504988196). We still need to merge that PR in the release build scripts repo.

syvb:
I pushed commits to get the CI closer to passing - https://github.com/timescale/timescaledb/pull/6077 broke some of the tests when it introduced a breaking change to TimescaleDB.

syvb:
Merging this manually since the main bors instance is gone.

PR #796, opened 2024-04-22
Fix approx_percentile arguments in two-step_aggregation example

vanderhoop:
Tiny little baby PR that fixes some example code


syvb (approved):
Thanks!

PR #804, opened 2024-07-01
Bump pgrx -> 0.11.4

matthewmcnew:


syvb:
I've run the tests with various pgrx versions, and weirdly I can replicate the serialization test failure on both pgrx 0.10.0 and 0.11.4. I'm going to try testing with more versions to see if I can find the pgrx version that introduced the change that's causing test failures.

syvb:
Bizarrely, I can replicate the test failure locally even when running on the main branch, but not on the Actions CI on main. I'm not sure why there would be a difference between my local machine and CI.

PR #820, opened 2024-10-23
PG17 Support

philkra:
Imported through https://github.com/timescale/timescaledb-toolkit/pull/819 /cc @Jo0

PR #824, opened 2024-11-04
Upgrade to pgrx 0.12.8 and add PostgreSQL 17 support

zilder:
Based on #820. Fixes compilation issues that came with pgrx 0.12.*.

Major changes:
- Since 0.12.* pgrx does not ['staticize functions' return types'](https://github.com/pgcentralfoundation/pgrx/pull/1476) in `pg_extern` macro expansion anymore. So all `pg_extern` functions' return types now have to explicitly use `'static' lifetime.
- Due to the above, auto-generated `_out` functions in `PostgresType` macro do not work (do not compile) because they get their return type annotated with rust type's lifetime (`'input` in our case). (This may actually be a bug and should probably be addressed `pgrx`.) For this reason I added default implementation for `_in`/`_out` functions, as well as ArgAbi and BoxRet (it seems to be a [package deal](https://docs.rs/pgrx-macros/0.12.8/src/pgrx_macros/lib.rs.html#861) when you implement custom in/out functions)

Minor things:
- `PostgresType` now requires a type to implement `Serializable`/`Deserializable` traits. As far as I understand they are used for default implementation of `IntoDatum`/`FromDatum`. Toolkit has its own implementation for these so ser/de aren't really needed. Also the default implementation of `Deserializable` has some conflicts with type's lifetime, so I added a trivial implementation and put `unimplemented!()` there as it isn't used anyway. If anyone has a better solution, pls let me know.
- Added `BoxRet` implementation that is now also required by `PostgresType` (I just copied the one that pgrx uses for a bunch of predefined types).
- `NodeTag` is now an `enum` hence many `const => enum` replacements.
- The dependency discovery algorithm in pgrx has changed which uncovered some dependencies (between `pg_extern` objects that is) in toolkit haven't been set up properly causing incorrect order of DDL commands in the installation script.


Jo0:
Thank you so much for handling this!

bors.toml
```diff
@@ -1,17 +1,19 @@
 status = [
```
syvb, line 1: We don't use bors anymore; this whole file can be removed.

extension/src/raw.rs
```diff
@@ -49,9 +49,9 @@ macro_rules! raw_type {
             fn type_oid() -> pg_sys::Oid {
                 $tyid
             }
-            fn array_type_oid() -> pg_sys::Oid {
-                $arrayid
-            }
+            // fn array_type_oid() -> pg_sys::Oid {
+            //     $arrayid
+            // }
```
JamesGuthrie, line 54: Remove?

JamesGuthrie (approved):

bors.toml
zilder: Thanks, I removed the file

extension/src/accessors.rs
```diff
@@ -16,6 +16,7 @@ macro_rules! accessor {
             $crate::pg_type!{
                 // TODO Move into pg_type as we don't care to vary it.
                 #[derive(Debug)]
+                // #[derive(serde::Serialize, serde::Deserialize)]
```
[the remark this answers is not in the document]
zilder: Thanks, I missed it

extension/src/raw.rs
zilder: Done

syvb (approved):

PR #827, opened 2024-11-13
fix: correct unaligned slice access

JamesGuthrie:
Tests started failing for an unknown reason. It turns out that we were attempting to build a slice with an unaligned pointer, and `slice::from_raw_parts` was panicking.

The reason the memory address is unaligned is that the Rust type has an 8-byte alignment requirement, but the Postgres type does not have a configured alignment, so it has the default 4-byte alignment. It is unclear why the type alignment was not configured in Postgres, presumably an oversight.

It appears as though we only end up with unaligned addresses when the types are stored on disk.

The affected code is doing zero-copy deserialization of Rust structs with Postgres-allocated memory. The deserialization code takes care of all internal alignment requirements, and assumes that the memory to be deserialized is aligned. Not all types are affected, only those which contain a slice of 8-byte aligned elements, because the slice is reconstructed with a (potentially 4-byte aligned) address.

This fix detects when we would attempt to deserialize data stored at an un-aligned address, and copies the data to a new (8-byte aligned) memory location, allowing for safe processing. It does not constrain this detection to only types with 8-byte aligned slices.

### DIFF
--- a/extension/src/type_builder.rs
+++ b/extension/src/type_builder.rs
@@ -225,7 +225,35 @@ macro_rules! pg_type_impl {
                         ptr = pg_sys::pg_detoast_datum_copy(ptr);
                     }
                     let data_len = pgrx::varsize_any(ptr);
-                    let bytes = std::slice::from_raw_parts(ptr as *mut u8, data_len);
+
+                    // NOTE: varlena types are aligned according to the `ALIGNMENT` with which they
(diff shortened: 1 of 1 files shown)


JamesGuthrie:
Note: We should also follow up by setting the postgres type alignment to 'double' for all varlena types which require 8-byte alignment.

zilder (approved):
Great work, thank you!

PR #834, opened 2024-11-29
Doc fix: broken continuous aggregates ref in T-Digest

xyztony:
replaced https://docs.timescale.com/latest/using-timescaledb/continuous-aggregates with https://docs.timescale.com/use-timescale/latest/continuous-aggregates/


syvb:
Thanks!

PR #854, opened 2025-04-10
Default to PostgreSQL 17, minor bump for pgrx

feikesteenbergen:


syvb:
You need to update the PGRX version in a few more places - do a global search for "0.12.8" to find all the places where it's relevant.

PR #882, opened 2025-08-27
Mark time_weight (and its rollup) parallel unsafe

syvb:
Parallel `time_weight` seems to cause issues when used with TimescaleDB compression.

(see https://github.com/timescale/Support-Dev-Collab/issues/3283)

### DIFF
--- a/extension/src/time_weighted_average.rs
+++ b/extension/src/time_weighted_average.rs
@@ -393,7 +393,7 @@ extension_sql!(
         combinefunc = time_weight_combine,\n\
         serialfunc = time_weight_trans_serialize,\n\
         deserialfunc = time_weight_trans_deserialize,\n\
-        parallel = restricted\n\
+        parallel = unsafe\n\
     );\n\
 \n\
     CREATE AGGREGATE rollup(tws TimeWeightSummary)\n\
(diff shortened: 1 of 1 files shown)


extension/src/time_weighted_average.rs
```diff
@@ -399,7 +399,7 @@ extension_sql!(
         combinefunc = time_weight_combine,\n\
         serialfunc = time_weight_trans_serialize,\n\
         deserialfunc = time_weight_trans_deserialize,\n\
-        parallel = restricted\n\
+        parallel = unsafe\n\
```
fabriziomello, line 402: I think you should also set `PARALLEL UNSAFE` for all internal functions:
* time_weight_trans
* time_weight_final
* time_weight_combine
* time_weight_trans_serialize
* time_weight_trans_deserialize

syvb: I think the functions that make up the aggregates should still be parallel safe:

https://github.com/timescale/timescaledb-toolkit/blob/d895f628334769f852979df4a5ca6b2102487934/extension/src/time_weighted_average.rs#L173

svenklemm (approved):

PR #905, opened 2026-05-10
Add guard for large BIGINT values in 1D aggregations.

### LINKED ISSUE #871: Hyperfunction returns wrong value for large bigint values
 - OS: macOS 15.4.1
 - PostgreSQL version (output of `SELECT version();`): 17.4
 - TimescaleDB Toolkit version (output of `\dx timescaledb_toolkit` in `psql`): 1.21.0
 - Installation method: docker

**Describe the bug**
When stats_agg is run on large bigint values like 5223372036854775805 and 5223372036854775806  that approach the limit of bigint, hyperfunctions like average return the wrong value. 

**To Reproduce**
Run this SQL:
`SELECT average(stats_agg(data))
FROM generate_series(5223372036854775805::bigint, 5223372036854775806) data;`

**Expected behavior**
Either an overflow error if not supported, or a workaround using decimal or numeric value types. 

**Actual behavior**
For the stated reproduction the result is '5223372036854775800'. The actual result should be '5223372036854775805.5'

**Screenshots**
Image

Image

surister:
Fixes https://github.com/timescale/timescaledb-toolkit/issues/871.

Previously, values were automatically cast to double precision, and any value outside the exact 53-bit f64 integer range could lose precision. 

For example:

```sql
SELECT 5223372036854775805::bigint::double precision
-- 5223372036854775800
```

Where `5223372036854775805` is outside of the 53-bit range.

```rust
const MAX_SAFE_INTEGER_F64: u64 = (1u64 << 53) - 1;
// 9007199254740991
```

This PR introduces BIGINT overload that checks whether the value can be represented without precision loss and raises an error if it cannot. This is a **breaking change** as it introduces an actual error instead of returning wrong data.

This only fixes 1D aggregations, 2D aggregations will most likely have the same behaviour.

### DIFF
--- a/extension/src/stabilization_info.rs
+++ b/extension/src/stabilization_info.rs
@@ -11,6 +11,12 @@
 
 crate::functions_stabilized_at! {
     STABLE_FUNCTIONS
+    "1.23.0" => {
+        stats_agg(bigint),
+        stats1d_i64_tf_inv_trans(internal,bigint),
+        stats1d_i64_tf_trans(internal,bigint),
+        stats1d_i64_trans(internal,bigint),
+    }
     "1.20.0" => {
         total(tdigest),
         total(uddsketch),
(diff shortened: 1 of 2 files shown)


surister:
This will need to be documented btw

dbeck (approved):
🚢

PR #915, opened 2026-05-15
Fix `max_n` and `max_n_by` when used on empty tables

### LINKED ISSUE #860: min_n_by/max_n_by fail when used on empty dataset
 - OS: AlmaLinux 9.5 x64
 - PostgreSQL version: 16.8
 - TimescaleDB Toolkit version: 1.19.0
 - Installation method: RPM

**Describe the bug**
I want to get the minimum and maximum value together with their timestamp (when did the min and max occur). The following query works fine in general, but it fails if there is no data in the specified time span (WHERE-clause).
```sql
SELECT "data" AS "timestamp", "value"
      FROM into_values((
          SELECT min_n_by("value", "timestamp", 1)
          FROM "measurement"
          WHERE "timestamp">='2024-01-01' AND "timestamp"<'2025-01-01'
      ), NULL::timestamptz);
```

The error I get is 
```
ERROR: called `Option::unwrap()` on a `None` value
```
Smells like Rust, doesn't it?

**To Reproduce**
1. Create an empty hypertable.
2. Execute a query like the one above.
3. See error

**Expected behavior**
I expect an empty result set.

**Actual behavior**
The error
```
ERROR: called `Option::unwrap()` on a `None` value
```
is returned.

Dvdandrades:
Fix #860

This only fixes max_n and max_n_by aggregates. I intend to fix min_n and min_n_by aggregates as soon as possible, since they leak the same Option::unwrap() on a `None` value error.

### DIFF
--- a/extension/src/nmost/max_by_float.rs
+++ b/extension/src/nmost/max_by_float.rs
@@ -107,7 +107,7 @@ pub fn max_n_by_float_to_values(
                 .into_iter()
                 .zip(agg.data.clone().into_anyelement_iter()),
         ),
-        None => TableIterator::new(std::iter::empty()),
+        None => TableIterator::empty(),
     }
 }
 
--- a/extension/src/nmost/max_by_float.rs
+++ b/extension/src/nmost/max_by_float.rs
@@ -225,7 +225,7 @@ mod tests {
     }
 
     #[pg_test]
-    fn max_n_by_empty_input_returns_null() {
+    fn max_by_float_empty_input_returns_null() {
         Spi::connect_mut(|client| {
             client.update(
                 "CREATE TABLE data(ts TIMESTAMPTZ, value DOUBLE PRECISION);", None, &[],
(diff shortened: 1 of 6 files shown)


extension/src/nmost/max_by_float.rs
```diff
@@ -86,22 +86,29 @@ pub fn max_n_by_float_rollup_trans(
...
-    )
+    match agg {
+        Some(agg) => TableIterator::new(
+            agg.values
+                .values
+                .clone()
+                .into_iter()
+                .zip(agg.data.clone().into_anyelement_iter()),
+        ),
+        None => TableIterator::new(std::iter::empty()),
```
surister: `TableIterator`has an `empty()` method already that you can reuse

surister (changes requested):

surister:
CI failures are unrelated and will be gone once we merge the update of al/10 + ubuntu 26.04

surister (approved):
lgtm

surister:
Please rebase with master.

PR #918, opened 2026-05-23
Apply Rustc and Clippy fixes

Dvdandrades:
This PR applies automated fixes recommended by rustc compiler suggestions and clippy lints across the codebase.

It updates calls to `extract_bits` to use fully qualified syntax via the `Extractable` trait. The change removes `unstable_name_collisions` warnings caused by potential conflicts with methods that may be added to the Rust standard library.

It moves the workspace resolver configuration to the root `Cargo.toml`, since resolver settings defined in non-root packages are ignored by Cargo workspaces. The resolver remains set to version `2` to preserve the current workspace behavior, although it could be migrated to resolver `3` in the future.

### DIFF
--- a/crates/encodings/src/lib.rs
+++ b/crates/encodings/src/lib.rs
@@ -91,7 +91,7 @@ pub mod zigzag {
 
     #[inline(always)]
     pub fn decode(n: u64) -> i64 {
-        if n % 2 == 0 {
+        if n.is_multiple_of(2) {
             // positive number
             (n / 2) as i64
         } else {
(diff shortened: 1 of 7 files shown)


Dvdandrades:
`cargo clippy` currently reports a future incompatibility warning coming from `nalgebra v0.27.1`. It does not break the build right now, but it may become an issue, so it could be worth tracking or updating the dependency chain at some point.

surister:
> `cargo clippy` currently reports a future incompatibility warning coming from `nalgebra v0.27.1`. It does not break the build right now, but it may become an issue, so it could be worth tracking or updating the dependency chain at some point.

Yes, I agree. We should fix that sometime soon.

surister (approved):
Thank you!

PR #921, opened 2026-05-31
Fix inverted is_sorted check in ASAP timevector smoothing

### LINKED ISSUE #844: Possible bug in `asap_smooth()`
Shouldn't this condition be reversed? I mean why sort an already sorted `timevector`?

https://github.com/timescale/timescaledb-toolkit/blob/84828fe8fb494a6a61412a3da96517fc80f7bb20/extension/src/asap.rs#L143-L145

Dvdandrades:
Fix #844 

Fix an inverted sorting condition in `asap_on_timevector`.

`Timevector_TSTZ_F64::is_sorted()` returns `true` when the `FLAG_IS_SORTED` flag is set, but the current implementation sorts the series only when `is_sorted()` is `true` ans skips sorting otherwise.

This change updates the condition so that sorting is performed only for unsorted timevectors, matching the behavior used in `asap_final_inner`.

### DIFF
--- a/extension/src/asap.rs
+++ b/extension/src/asap.rs
@@ -138,9 +138,7 @@ pub fn asap_on_timevector(
     resolution: i32,
 ) -> Option<Timevector_TSTZ_F64<'static>> {
     // TODO: implement this using zero copy (requires sort, find_downsample_interval, and downsample_and_gapfill on Timevector)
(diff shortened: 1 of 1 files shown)


surister (approved):
Good catch, thanks 🤩

Issue #875, opened 2025-06-23
Continous aggregate with a column of type heartbeatagg cannot be refreshed


juantxorena:
**Relevant system information:**
 - OS: dockerized debian 12
 - PostgreSQL version 16.9
 - TimescaleDB Toolkit version 1.21.0
 - Installation method: docker

**Describe the bug**
In a continuous aggregate where one of the columns is of type heartbeatagg, I cannot call the refresh_continuous_aggregate function, because it seems to parse the column type as it if where a function.

**To Reproduce**
Minimal example:
1. Create a hypertable like this:
```
CREATE TABLE measurement (
	"timestamp" timestamptz NOT NULL,
	vehicle_id uuid NOT NULL
)
SELECT public.create_hypertable('measurement', public.by_range('timestamp', INTERVAL '7 days'));
```
2. Create a continuous aggregate:
```
CREATE MATERIALIZED VIEW IF NOT EXISTS heartbeat_half_hour
WITH (timescaledb.continuous, timescaledb.materialized_only = false) AS
SELECT
    m.vehicle_id,
    time_bucket('30 minutes', m.timestamp, 'UTC') AS bucket_half_hour,
    heartbeat_agg(
        m.timestamp,
        time_bucket('30 minute', m.timestamp, 'UTC'),
        INTERVAL '50 minutes',
        INTERVAL '20 minutes'
    ) AS heartbeat
FROM measurement AS m
GROUP BY vehicle_id, bucket_half_hour
;
```
3. Insert some test data
4. Refresh the aggregate so it includes the inserted data:
```
CALL refresh_continuous_aggregate('heartbeat_half_hour', now() - INTERVAL '1 month', now())
```
5. It fails

**Expected behavior**
The refresh should work. It shouldn't try to read the type "public.heartbeatagg" as a function for some reason.

**Actual behavior**
This is the error:
```
2025-06-23 14:08:04.143 UTC [4198] ERROR: operator does not exist: public.heartbeatagg = public.heartbeatagg at character 376


2025-06-23 14:08:04.143 UTC [4198] HINT: No operator matches the given name and argument types. You might need to add explicit type casts.


[...]

PR #922, opened 2026-06-02
Add `heartbeat_agg` equality operator

Dvdandrades:
Add equality operators for `heartbeat_agg` to prevent errors when comparing values of this aggregate. This addresses the issue reported in #875 and lays the groundwork for supporting equality comparisons on other custom aggregate that may be affected by the same limitation.

This change implements `Eq` and `PartialEq` for `HeartbeatAgg`, allowing `pg_operator` to successfully define equality operators for the type.

As a result, the following query:

```sql
WITH aggs AS (
  SELECT
    heartbeat_agg(ts, batch, '2h', '20m') AS a,
    heartbeat_agg(ts, batch, '2h', '20m') FILTER (WHERE ts < '2020-01-01 01:00 UTC') AS b
  FROM poc
)
SELECT a = a, a <> a, a = b, a <> b
FROM aggs;
```

now returns:

```sql
 ?column? | ?column? | ?column? | ?column?
----------+----------+----------+----------
 t        | f        | f        | t
(1 row)
```

### DIFF
--- a/extension/src/heartbeat_agg.rs
+++ b/extension/src/heartbeat_agg.rs
@@ -327,6 +327,20 @@ impl HeartbeatAgg<'_> {
     }
 }
 
+impl PartialEq for HeartbeatAgg<'_> {
+    fn eq(&self, other: &Self) -> bool {
+        self.start_time == other.start_time
+            && self.end_time == other.end_time
+            && self.last_seen == other.last_seen
+            && self.interval_len == other.interval_len
+            && self.num_intervals == other.num_intervals
+            && self.interval_starts.as_slice() == other.interval_starts.as_slice()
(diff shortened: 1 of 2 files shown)


Dvdandrades:
Initially, I implemented a `pg_extern` function for the equality operator, but later reconsidered it as it may not be necessary. It can be added back if needed in the future.

For this PR in general, this was the approach that allowed implementing equality operators without breaking the build with multiple errors. I'm aware that `pgrx` provides `#[derive(PostgresEq)]` for automatically implementing the standart Postgres equality, comparison, and hash operators/functions, along with their OPERATOR CLASS and OPERATOR FAMILY definitions. I attempted to use it, but ran into multiple lifetime-related errors.

surister (approved):
LGTM tried it locally and looks nice, thank you!

Issue #913, opened 2026-05-15
Remove linking fix after v0.18.1 lands


surister:
Workaround fixes were introduced to circumvent a pgrx v1.18.0 bug.

There are two places:
1. https://github.com/timescale/timescaledb-toolkit/commit/a57ba5c89e5261adc33bb091aa7acec165221f82
2. https://github.com/timescale/timescaledb-toolkit/commit/b888ca2e9aebc91e45e6bf22f840e0ebb6dd0ca7#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03f

PR #935, opened 2026-06-22
Update Rust dependencies and fix resulting errors and lints

mfelsche:
Only some pest deprecation-warnings are left.

I know this is a big one. Feel free to pull what you need from it.
I can also split it into different commits for each crate.

No AI was used doing this PR or the changes therein.


surister:
Hey there, I'll have a look shortly, btw your name was somehow familiar to me and it looks like you worked at CrateDB as well, funny how small the world is.

surister:
Looks like we will need to upgrade rustc to 1.96

mfelsche:
I can bump the rust version in the docker image built stuff in the repo, but I feel like this is a chicken and egg problem. I guess the best way forward is to have another PR updating the rust version?

extension/Cargo.toml
```diff
@@ -17,9 +17,9 @@ pg_test = ["approx"]
 [dependencies]
 # Keep synchronized with `cargo install --version N.N.N cargo-pgrx` in Readme.md and docker/ci/Dockerfile
 # Also `pgrx-tests` down below in `dev-dependencies`.
-pgrx = "=0.18.0"
-pgrx-macros = "=0.18.0"
-pgrx-sql-entity-graph = "=0.18.0"
+pgrx = "=0.18.1"
```
surister, line 20: Upgrading to 0.18.1 also is an opportunity to go for https://github.com/timescale/timescaledb-toolkit/issues/913

surister (approved):
It looks good overall, went through it a couple of times. Run tests locally and found a couple of things to fix, i will just commit it to the PR's branch.

Now that we upgrade to 18.1 we should cleanup the 18.0 fixes and upgrade the CI rust versions, this needs some internal docker image building that I'll have to take care of, could take a bit since the build matrix is big and it wouldn't be weird to find an issue in one of the build targets.

Thank you a lot!

mfelsche:
Ok, sounds good to me!
If you need me to do anything with this PR, feel free to reach out here. :)

surister:
`Build CI` fails because the token is not passed when the source trigger is not from the repo itself, in this PR it's the fork `mfelsche`. I triggered builds manually on my end.

PR #946, opened 2026-07-24
Relax `stats_agg::tests::pg_stats_agg_fuzz` flakiness

### LINKED ISSUE #912: `stats_agg::tests::pg_stats_agg_fuzz` is flakky
Happens in CI and locally:

```

Client Error:
Output didn't match between postgres command: SELECT regr_intercept(test_y, test_x)::float, (SELECT regr_intercept(test_y, test_x) OVER (ORDER BY test_x ROWS BETWEEN 1 PRECEDING AND 1 FOLLOWING) FROM test_table LIMIT 1 OFFSET 3)::float FROM test_table                                                                                                                                                                                                                 
and stats_agg command: SELECT intercept(stats_agg(test_y, test_x))::float, (stats_agg(test_y, test_x)->intercept())::float, intercept((SELECT stats_agg(test_y, test_x) OVER (ORDER BY test_x ROWS BETWEEN 1 PRECEDING AND 1 FOLLOWING) FROM test_table LIMIT 1 OFFSET 3))::float FROM test_table                                                                                                                                                                           
        postgres result: 648237.0965988464                                                                                                                                                                                            
        statsagg result: 648237.096598847                                                                                                                                                                                             
        relative difference:         0.0000000000000008979378258182045                                                                                                                                                                
        allowed relative difference: 0.0000000000000002220446049250313                                                                                                                                                                
Failed after 3 successful iterations, run using 10000 values generated from seed 14705134910594273637                                                                                                                                 
postgres location: stats_agg.rs:1719
rust location: <unknown>
```

Dvdandrades:
Fixes #912

This PR fixes an intermittent test failure caused by an overly strict floating point tolerance during fuzz testing of the regression functions.

The test suite previously enforced an `EPS1` (`f64::EPSILON`) tolerance of approximately `2.22e-16`. In some cases, the relative difference exceeded this value, causing the test to panic.

As a workaround, this PR introduces a `5. * f64::EPSILON` tolerance of approximately `1.11e-15` for `regr_intercept`. This should be sufficient to accommodate the relative difference observed by @surister in #912. 

I suspect there may be a better solution that avoids introducing a new constant or relaxing the tolerance, but this change should address the intermittent failure for now.

### DIFF
--- a/extension/src/stats_agg.rs
+++ b/extension/src/stats_agg.rs
@@ -2002,6 +2002,7 @@ mod tests {
             const EPS1: f64 = f64::EPSILON; // Generally enough to handle float rounding
             const EPS2: f64 = 2. * f64::EPSILON; // stddev is sqrt(variance), so a bit looser bound
             const EPS3: f64 = 3. * f64::EPSILON; // Sum of squares in variance agg accumulates a bit more error
+            const EPS4: f64 = 4. * f64::EPSILON;
             const BILLIONTH: f64 = 1e-9; // Higher order moments exponentially compound the error
 
             check_agg_equivalence(
@@ -2235,7 +2236,7 @@ mod tests {
                 client,
                 &pg2d_agg("regr_intercept"),
                 &tk2d_agg("intercept"),
-                EPS1,
+                EPS4,
                 true,
             );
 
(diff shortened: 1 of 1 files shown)


Dvdandrades:
Looking for another possible maximum value for the relative difference in the `regr_intercept` test, I found that `variance_y` fails for the same reason in one of the failed [CI](https://github.com/timescale/timescaledb-toolkit/actions/runs/29992591932/job/89158561154#step:8:909) runs.

The test suite previously enforced an `EPS3` (`3. * f64::EPSILON`) tolerance of approximately `6.66e-16`, but the observed relative difference was `7.43e-16`, causing the test to fail. 

This PR also increases the `variance_y` tolerance to `4. * f64::EPSILON`,  approximately `8.88e-16`.

surister (approved):

PR #962, opened 2026-08-11
Assign points by index instead of i / n elapsed time in asap_smooth

### LINKED ISSUE #845: ASAP smoothing doesn't respect time dimension
 - PostgreSQL version (output of `SELECT version();`): PostgreSQL 16.8 (Ubuntu 16.8-1.pgdg22.04+1) on x86_64-pc-linux-gnu, compiled by gcc (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0, 64-bit
 - TimescaleDB Toolkit version: 1.18.0
 - Installation method: apt
 - TimescaleDB version: 2.18.2

**Describe the bug**
The `asap` smoothing algorithm messes up the time dimension if points are not equally spaced. This seems to be a bug in the Toolkit implementation, as a Python reference implementation does not have that issue.

**To Reproduce**

Run this query: https://gist.github.com/pedrokost/19bea6e07a98a8260a96ba104971c703#file-query-sql

Compare with the results of the [Python implementation](https://gist.github.com/pedrokost/19bea6e07a98a8260a96ba104971c703):

```
# make sure asap.py and data.csv are in the same folder.
uv pip install numpy pandas matplotlib scipy
uv run asap.py
```

**Expected behavior**
The asap smoothing line should closely follow the raw data line.

**Actual behavior**
There is a notable deviation in the asap smoothing line from the raw data line.

**Screenshots**

See how the green line deviates from the blue line in this TimescaleDB SQL query result:
Image

Comparing with the Python reference implementation: 
Image

**Additional context**
The `lttb` function does not have this issue.

Please note that the python code for plotting the results of the asap function is based [on this](https://github.com/stanford-futuredata/ASAP/blob/master/ASAP.ipynb), and has been slightly modified to plot the timeseries data appropriately.

surister:
After smoothing an array of values, while re-assigning a `value` to its timestamp it assumes that all points were evenly spread over time, meaning that `i/n` (where n in the length) was correct, but on data that is not evenly spread this assumption is incorrect, resulting in assigning a value to a unrealistic timestamp and thus stretching and giving very error-looking/skewed plots.

This means that our asap_smooth ultimately loses information.

A more correct way of implementing this is assigning by original index. To implement this we introduce `asap_smooth_with_metadata` which is a wrapper of the old asap_smooth, but it returns `AsapSmoothResult`, which contains enough metadata to reconstruct the indices, later in `smooth_points` we reconstruct and assign by index

Note:
The original paper assumes same distance points https://arxiv.org/pdf/1703.00983

### DIFF
--- a/crates/asap/src/lib.rs
+++ b/crates/asap/src/lib.rs
@@ -25,23 +25,47 @@
 
 mod fft;
 
-// Smooth out the data to promote human readability, resolution is an upper bound on the number of points returned
+/// Smooth out the data to promote human readability, resolution is an upper bound on the number of points returned
 pub fn asap_smooth(data: &[f64], resolution: u32) -> Vec<f64> {
+    asap_smooth_with_metadata(data, resolution).values
+}
+
+pub struct AsapSmoothResult {
+    pub values: Vec<f64>,
+    pub preaggregation_window: u32,
+    pub smoothing_window: u32,
+}
+
+pub fn asap_smooth_with_metadata(data: &[f64], resolution: u32) -> AsapSmoothResult {
     use std::borrow::Cow;
 
     if resolution == 0 || data.is_empty() {
-        return Vec::new();
+        return AsapSmoothResult {
+            values: Vec::new(),
+            preaggregation_window: 0,
+            smoothing_window: 0,
+        };
     }
 
-    let data = if data.len() > 2 * resolution as usize {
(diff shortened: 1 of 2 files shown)


surister:
User created:

asap_smooth vs python implementation (which uses index assignment)
image

new asap_smooth vs python implementation
Actual Toolkit vs Python reference

dbeck (approved):

PR #966, opened 2026-08-16
Fix type handling in timevector pipelines and space-saving aggregates

svenklemm:
Both the timevector `map` pipeline elements and the space-saving
(`freq_agg`/`mcv_agg`) aggregates can be built from their text
representation, which let untrusted input diverge from the invariants the
readers rely on, reinterpreting a stored datum as the wrong type.


surister (approved):

surister:
had a look at other parts of the code and didn't find anything similar

PR #971, opened 2026-08-20
Change pgrx::* imports to pgrx::prelude::*

Dvdandrades:
Following #586, this PR changes `pgrx::*` to `pgrx::prelude::*`, avoiding scope pollution for the importer and making it clearer where items come from.

### DIFF
--- a/extension/src/asap.rs
+++ b/extension/src/asap.rs
@@ -1,5 +1,5 @@
 use asap::*;
-use pgrx::*;
+use pgrx::prelude::*;
 use serde::{Deserialize, Serialize};
 
 use crate::{
--- a/extension/src/asap.rs
+++ b/extension/src/asap.rs
@@ -1,5 +1,5 @@
-use asap::*;
-use pgrx::prelude::*;
+use asap::asap_smooth_with_metadata;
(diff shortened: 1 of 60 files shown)


extension/src/asap.rs
```diff
@@ -1,5 +1,5 @@
 use asap::*;
-use pgrx::*;
+use pgrx::prelude::*;
```
surister, line 2: this is kind of the same no?

Dvdandrades: Mostly yes. [prelude](https://github.com/tcdi/pgx/pull/699) is similar, but unlike `pgrx::*`, it only imports commonly used identifiers, reducing namespace pollution and potential name collisions.

surister: I get that, but if we are in the business of making it less polluted, we might as well just avoid '*' altogether and specifically import what we need no?

Dvdandrades: You are technically correct. The best kind of correct. I can finish the job and replace every `*` with the specific import. if you agree.

surister: yeah

PR #974, opened 2026-08-24
Remove combine/serialize/deserialize from time_weight and rollup

Poroma-Banerjee:
These aggregates are marked PARALLEL UNSAFE because their combine
function requires strictly disjoint input time ranges. However, while
PostgreSQL's partitionwise aggregation and TimescaleDB's chunk-wise
support partial aggregation neither will guarantee the time_weight
combine function's requirements.

https://github.com/timescale/Support-Dev-Collab/issues/4601


akuzm:
> However, neither PostgreSQL's partitionwise aggregation nor TimescaleDB's chunk-wise aggregation checks proparallel before splitting an aggregate into partial + combine.

The point is that they are not supposed to. There's no way in Postgres to express the guarantee that the time_weight combine function requires.

akuzm (approved):

akuzm:
I think you also need some kind of upgrade handling to drop the existing functions on extension upgrade. Not sure how it's done in toolkit.

akuzm:
Turns out the update tests were dead, I fixed them on main, you'll have to update this PR branch to run them.

PR #975, opened 2026-08-25
Fix upgrade tests

akuzm:
Looks like it was broken more than 3 years ago, because we removed the support for the old versions we're using as a cutoff for something we can upgrade from.

Broken after this change: https://github.com/timescale/timescaledb-toolkit/pull/743

Additionally, the update tester code needs a fix to support this dependencies update: https://github.com/timescale/timescaledb-toolkit/pull/935

### DIFF
--- a/tools/update-tester/src/testrunner.rs
+++ b/tools/update-tester/src/testrunner.rs
@@ -458,6 +458,7 @@ fn get_values(query_results: Vec<SimpleQueryMessage>) -> QueryValues {
                 }
                 Some(values)
             }
+            SimpleQueryMessage::RowDescription(_) => None,
             _ => unreachable!(),
         })
         .collect()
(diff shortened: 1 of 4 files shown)


dbeck (approved):

tools/update-tester/src/testrunner.rs
```diff
@@ -245,19 +245,7 @@ impl TestClient {
                 db_creation_client
                     .simple_query(&drop)
                     .unwrap_or_else(|e| panic!("could not drop db {test_db_name} due to {e}"));
-                let locale_flags = {
```
akuzm, line 248: I can't find why it is required, and it creates problems if it doesn't match the locale used during initdb. If it turns out we do need this, probably have to set during initdb to avoid mismatches.

